PR #126, opened 2013-10-18
Avoid spawning console windows when running from .pyw

boppreh:
I'm developing a small background application (Windows + PyQt) to track Git repositories.

The GitPython module works great, but when called from a windowless script (.pyw) each command spawns a console window. Those windows are empty and last only a second, but I think they should not be there at all.

I saw it was using `Popen` to call the `git` executable and added a `shell=True` parameter to avoid spawning those windows.

Be aware I haven't tested this on other machines or operating systems.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -336,6 +336,7 @@ class Git(LazyMixin):
 						stderr=PIPE,
 						stdout=PIPE,
 						close_fds=(os.name=='posix'),# unsupported on linux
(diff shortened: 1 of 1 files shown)


Byron:
Please note that with the release of *v0.3.7*, you will need to to set `git.Git.USE_SHELL=True` to get back to the same behaviour.
See related issue #248 for more information.

PR #221, opened 2015-01-02
[FIX] remote.fetch: branch names may contain '#' characters

tremlin:
The regular expression for parsing `git fetch -v` does not handle '#' characters in branch names correctly.


Byron:
Thank you !

PR #228, opened 2015-01-07
Make a commit with another author

firm1:
We could not do a commit by filling the author directly. This PR provides this functionality.

I have include test and documentation.


Byron:
Thank you.
I have altered the test case to include unicode characters and verify this works as well.

firm1:
Thanks

Issue #234, opened 2015-01-18
Allow setting per-repo SSH key


teeberg:
I've locally patched this by temporarily setting the `GIT_SSH` environment variable to a script that would include a deployment SSH key in the SSH command. I think this would be awesome to have built-in. I find this much more convenient and flexible than having to parse and update the `~/.ssh/config`.

Would there be interest in a pull request for this feature?

Byron:
Of course, I would be looking forward to receiving one.
I will assign this ticket to the 'far in the future' release until I receive a PR or get around to implementing this myself.
Version 0.3.6 could be released this week, and your PR could be part of it if it makes it here in time.
Thank you

teeberg:
Glad to hear that. I'm looking into it! :)

Right now, I'm trying to figure out what would be the best syntax to support it. Should it be a keyword argument to `fetch`/`pull`/`clone`/`push`?
```py
repo.pull(sshkey=GIT_SSH_KEY)
```
Or a property of `Remote` that gets passed in upon instantiation? If you need it for either of fetch/pull/clone/push, then you will need it for all other commands too, so the latter seems more appropriate. But remote objects may not have been initialized by the user, so what do we do if it wasn't?

In my own class, I've made it a context manager that temporarily injects `GIT_SSH` into the environment:
```py
with remote.sshkey('...'):
    remote.pull()
```
Being much more familiar with the codebase than I am, what's your favorite/opinion/recommendation? I'm also open to other proposals and implementation suggestions!

Byron:
From what I see, you have pinpointed the options precisely already !

If your envisioned implementation indeed sets environment variables that are to be inherited by called git processes, it might be most comfortable to set it up where it happens, in the [`Git`](https://github.com/gitpython-developers/GitPython/blob/master/git/cmd.py#L1) type.

In that case, the API could look like this:
```python
r = git.Repo('repo_dir')
previous_env = r.git.set_environment(dict(GIT_SSH_KEY="I really don't know what to pass here ;)"))
# Now all upcoming git calls will have the GIT_SSH_KEY environment variable set, along with all the other default ones.
# Returning the previous version of the dict (which may be empty) helps to change environments on the fly, for example with a custom context manager.
```
As you can see, my suggestion is very generic, but should be just what you need to get GIT_SSH_KEY to work.

[...]

PR #242, opened 2015-01-21
Add 'sshkey' context manager

teeberg:
Proposed implementation for #234

I have done some manual tests, but not set up automated tests yet. I will see how it fits into your suite tomorrow if you haven't implemented it yet. :)

Let me know if something's not quite in the right place.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -618,7 +618,7 @@ class Git(LazyMixin):
     def environment(self):
         return self._environment
 
-    def set_environment(self, **kwargs):
+    def update_environment(self, **kwargs):
         """
         Set environment variables for future git invocations. Return all changed
         values in a format that can be passed back into this function to revert
@@ -626,8 +626,8 @@ class Git(LazyMixin):
 
         ``Examples``::
 
-            old_env = self.set_environment(PWD='/tmp')
-            self.set_environment(**old_env)
+            old_env = self.update_environment(PWD='/tmp')
+            self.update_environment(**old_env)
 
         :param kwargs: environment variables to use for git processes
         :return: dict that maps environment variables to their old values
@@ -648,23 +648,23 @@ class Git(LazyMixin):
         return old_env
 
     @contextmanager
-    def environment(self, **kwargs):
+    def with_environment(self, **kwargs):
         """
-        A context manager around the above set_environment to restore the
+        A context manager around the above update_environment to restore the
         environment back to its previous state after operation.
 
         ``Examples``::
 
-            with self.environment(GIT_SSH='/bin/ssh_wrapper'):
+            with self.with_environment(GIT_SSH='/bin/ssh_wrapper'):
                 repo.remotes.origin.fetch()
 
-        :param kwargs: see set_environment
+        :param kwargs: see update_environment
         """
-        old_env = self.set_environment(**kwargs)
+        old_env = self.update_environment(**kwargs)
         try:
             yield
         finally:
-            self.set_environment(**old_env)
+            self.update_environment(**old_env)
 
     @contextmanager
     def sshkey(self, sshkey_file):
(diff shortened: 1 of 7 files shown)


git/cmd.py
```diff
@@ -603,6 +610,73 @@ def execute(self, command,
...
+        :param kwargs: see set_environment
+        """
+        old_env = self.set_environment(**kwargs)
+        try:
+            yield
+        finally:
+            self.set_environment(**old_env)
+
+    @contextmanager
+    def sshkey(self, sshkey_file):
```
Byron: For a moment I thought this method is to specific, but on the other hand, it makes setting ssh-keys (by using key-files) really easy.
Therefore the only thing I will do is see how this works on windows, and possibly put in an assertion to run on non-windows systems only.

scripts/ssh_wrapper.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
```
Byron: Possibly the MANIFEST.in file needs adjustment to assure this file also gets distributed when making new releases through setuptools.

git/cmd.py
```diff
@@ -603,6 +610,73 @@ def execute(self, command,
         else:
             return stdout_value
 
+    def set_environment(self, **kwargs):
```
Byron: From looking at the implementation, I'd be tempted to call this method `update_environment`. `set_environment` seems to be more like this:
```python
res = self._environment
self._environment = kwargs
return res
```

git/cmd.py
```diff
@@ -603,6 +610,73 @@ def execute(self, command,
...
+                    old_env[key] = None
+                self._environment[key] = value
+            # remove key from environment if its value is None
+            elif key in self._environment:
+                old_env[key] = self._environment[key]
+                del self._environment[key]
+        return old_env
+
+    @contextmanager
+    def environment(self, **kwargs):
```
Byron: This is an interesting one ! When used as `with repo.git.environment(FOO='bar'):`, the name makes total sense. When I imagine to write a test-case, I'd probably want to query the current state of the custom environment like this too: `repo.git.environment()`, following the `set_foo()`, `foo()` convention.

Can you imagine a better name for `environment()` to get a context manger ? I know, it's sweet and short already and anything else will make it longer ... `with altered_environment()`, `with changed_environment()`, `with custom_environment()` ?

Byron:
Thank you !
It looks very good already, and is more feature-rich than I would have done it, thanks to your personal interest.

git/test/test_git.py
```diff
@@ -153,3 +153,23 @@ def test_env_vars_passed_to_git(self):
         editor = 'non_existant_editor'
         with mock.patch.dict('os.environ', {'GIT_EDITOR': editor}):
             assert self.git.var("GIT_EDITOR") == editor
+
+    def test_environment(self):
```
teeberg: I don't see how I could test the `sshkey` context manager without accessing an actual repo. Also, how could I check whether the environment is actually passed into the git process? The former would cover the latter. Do you have ideas?

Byron: Something that could work is to adjust the script for the test to actually fail with a distinct error code, and possibly emit something distinctive to STDERR.
That way, you could call the real git process, on a bogus git@server:foo/bar repo URL, and verify that it failed under your conditions. That can proof that the script was actually called.
The cool thing is that this would require you to put in a way to adjust the path the script handling the GIT_SSH_KEY, which could be done so that advanced users can either set it, or derive from the Git command implementation.

Byron:
You did it !
The only minor issue I am still choking on is the `with repo.git.with_environment(...)` notation of the context manager and the seemingly redundant use of `with_` .

**Note to self: Test this on windows and at least throw saying sshkey doesn't work, if it doesn't work**

Also I believe today will be the day, as I should get through my docs review, and prepare a new release.

Byron:
Before I forget: I had to remove `sshkey` from the release, as I was unable to tell `setup.py` to keep that file where it is after installation, AND keep it executable.
I think I even recorded my futile attempt to beat setuptools in all it's glory and uploaded it on youtube.

teeberg:
Thanks for the note, I already saw it! :) I didn't realise it would be such a pain! I'll stick with my local SSH script then in the meantime. :-) Thanks for merging everything else in!

PR #256, opened 2015-02-09
Replace GIT_SSH with GIT_SSH_COMMAND for SSH key management.

teeberg:
Also move untestable documentation out of test.

Related: #234, #242

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -331,12 +331,11 @@ You can easily access configuration information for a remote by accessing option
     :start-after: # [26-test_references_and_objects]
     :end-before: # ![26-test_references_and_objects]
 
-You can also specify per-call custom environments using a new context manager on the Git command
+You can also specify per-call custom environments using a new context manager on the Git command, e.g. for using a specific SSH key.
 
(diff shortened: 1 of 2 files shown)


teeberg:
I actually think it's a really cool idea to test the documentation like this, but if you have to say that we don't actually make the call because the test setup won't allow us to, I'd rather remove it from the test... :-)

Byron:
Thank you ! I wouldn't mind putting in the example in clear-text in this case. 
However, as GitPython aims to support multiple different versions of the git commandline program, it would be great to keep both versions, stating which `git --version` you require to make use of either one.
Do you think you can make this change for me so I can make the merge ?
Thanks in advance

Byron:
I'd like to make a new release this week, and would be glad if I could merge this PR. Can you have a look at my previous request ? It shouldn't take long. If you have no time for this just let me know and I will do it for you.
Thanks

Byron:
You can watch the development stream [on youtube](http://youtu.be/Q-sl62-MWAE).

teeberg:
Thanks for taking over. :-) I was willing to do it, but I'm indeed rather busy and it probably takes you way less time.

PR #261, opened 2015-02-20
BF: do not checkout master -- that ruins testing of PRs

yarikoptic:
For PRs or testing of other branches, we should remain in the tree as travis prepairs it.  Switching back to the state of master is somewhat ruins the point ;)


yarikoptic:
Long path short -- should work now ;)

PR #266, opened 2015-03-02
Store path attribute on Diff object

jeblair:
If a file in a commit contains no changes (for example, if only the
file mode is changed) there will be no blob attached.  This is
usually where the filename is stored, so without it, the calling
context can not tell what file was changed.  Instead, always
store a_path and b_path on the Diff object so that information
is available.


Byron:
Thank you :) ! If travis is green, I am green too !

PR #282, opened 2015-04-15
fix(util): Handle 'Finding sources' messages in RemoteProgress

matt-jordan:
When running a long running operation (such as a clone on a large repo),
Git may return a message indicating that it is 'Finding sources'. Since
there is no bit field value for this message, this causes a large amount
of error messages to be emitted to stderr.

This patch fixes this by adding another bit field value for this
message, FINDING_SOURCES. Derived classes can look for this op_code and
handle it appropriately.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -165,7 +165,7 @@ class RemoteProgress(object):
     and git-fetch and to dispatch callbacks allowing subclasses to react to the progress.
     """
     _num_op_codes = 7
-    BEGIN, END, COUNTING, COMPRESSING, WRITING, RECEIVING, RESOLVING = [1 << x for x in range(_num_op_codes)]
+    BEGIN, END, COUNTING, COMPRESSING, WRITING, RECEIVING, RESOLVING, FINDING_SOURCES = [1 << x for x in range(_num_op_codes)]
     STAGE_MASK = BEGIN | END
     OP_MASK = ~STAGE_MASK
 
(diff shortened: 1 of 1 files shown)


Byron:
Thank you ! It seems it wasn't actually tested locally though. Setting `_num_op_codes` to 8 should make it work.

matt-jordan:
Eek. Sorry about that - I had originally made the change in another branch, and did a poor copy over to the new branch. Whoops.

matt-jordan:
The flake8 error is complaining about the length of the line. I don't normally like using '\', but would the following be acceptable:

```
168     BEGIN, END, COUNTING, COMPRESSING, WRITING, RECEIVING, RESOLVING, FINDING_SOURCES = \
169         [1 << x for x in range(_num_op_codes)]

```

I'd be happy with any other suggestion for wrapping that line.

matt-jordan:
I'm not sure I know what is occurring with this test:

```

FAIL: test_init_repo_object (git.test.test_docs.Tutorials)

File "/home/travis/build/gitpython-developers/GitPython/git/ext/gitdb/gitdb/test/lib.py", line 87, in wrapper

return func(self, path)

File "/home/travis/build/gitpython-developers/GitPython/git/test/test_docs.py", line 97, in test_init_repo_object

assert (now.commit.tree / 'VERSION').data_stream.read().decode('ascii').startswith('0')

AssertionError
```

If you can give me a hint, I'd be happy to go in and fix it.

Byron:
Thanks, I will take it from here :).

You can watch the development stream [on youtube](http://youtu.be/ZnzuwiLjiG4).

PR #284, opened 2015-04-21
git-daemon test fixes

moben:
We are building packages in a sandbox that restricts network access somewhat.

Here are a couple of changes I made that should be generally useful.

a) make the git-daemon bind to localhost instead of 0.0.0.0 (aka all interfaces), as there's no reason it should be accessible from the outside
b) make the listen port configurable via an environment variable, in case it is already taken.

### DIFF
--- a/git/test/lib/helper.py
+++ b/git/test/lib/helper.py
@@ -200,7 +200,7 @@ def with_rw_and_rw_remote_repo(working_tree_ref):
             temp_dir = osp(_mktemp())
             # On windows, this will fail ... we deal with failures anyway and default to telling the user to do it
             try:
-                gd = Git().daemon(temp_dir, enable='receive-pack', as_process=True)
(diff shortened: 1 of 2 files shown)


Byron:
Thank you very much, I will gladly merge. Would you mind having a look at this line, as [indicated by flake8](https://travis-ci.org/gitpython-developers/GitPython/jobs/59387903#L659) beforehand ?
```bash
The command "nosetests -v --with-coverage" exited with 0.
1.47s$ flake8
./git/test/lib/helper.py:204:121: E501 line too long (125 > 120 characters)
The command "flake8" exited with 1.
```

You can try it yourself by running:
```
tox -e flake8
```

If you are out of time, I can also fix that one later today - just let me know.
Thanks

Byron:
Thank you :) ! I didn't even notice that the fix was already pushed.

PR #288, opened 2015-05-13
Fix type error (startswith expects bytes)

johnwalker:


Byron:
Thank you !

johnwalker:
You're welcome, thank you for gitpython!

PR #291, opened 2015-06-01
support passing odbt for using with Repo

hvnsweeting:


Byron:
Thank you ! I'd be glad to merge, but would be happy if you could add the docstring for the newly added parameter beforehand ?

Byron:
Thank you. Sorry for the late merge, but I don't get notifications if a new commit is pushed to a PR.

hvnsweeting:
ah yes, I should have commented after that :grin:

PR #294, opened 2015-06-05
Allow submodules to be ignored in is_dirty

Javex:
I had a use case with [PyGitUp](https://github.com/msiemens/PyGitUp) where I did not want submodules to be considered when asking whether a repo is dirty. This change allows that. I am a bit unsure with the "untracked_files" part: I would think that it's not necessary, as I can't imagine an untracked submodule (how would that work? Git would not see it as a submodule), but I included this change anyway. I can of course revert it if you are sure that there can't be a relevant case.

One problem though: I could not run (and thus not write) tests as even the most current tag had test failures for me on Python 2.7.

Please let me know what you think.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -563,7 +563,8 @@ class Repo(object):
     alternates = property(_get_alternates, _set_alternates,
                           doc="Retrieve a list of alternates paths or set a list paths to be used as alternates")
 
-    def is_dirty(self, index=True, working_tree=True, untracked_files=False):
+    def is_dirty(self, index=True, working_tree=True, untracked_files=False,
+                 consider_submodules=True):
         """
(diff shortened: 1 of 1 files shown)


Byron:
Thank you very much !

Byron:
You can watch the development stream [on youtube](https://youtu.be/q0x0jLeyaHY).

*`GitPython #5 [merge PR 294 - improve submodule handling of is_dirty method]`*

[\thumb](https://youtu.be/q0x0jLeyaHY)

Javex:
The youtube thing was great :) Nice to hear some comments on the PR. You could of course have asked, even if 5 days have passed, I would have gladly made the changes. Thanks for being so quick in merging the request.

Issue #10, opened 2010-12-01
Review Exception Types


Byron:
Currently, whenever something git-specific happens, git-python will throw a more or less generic Python exception, such as ValueError or TypeError, including a very git specific message though.

Client Code is less readable that way, as it will handle these general exceptions very specifically although this is not indicated by the exception type.

Review the all raise statements and introduce custom Exception types which are derived from the type you currently see. This way client code will be more readable without breaking existing code.

Byron:
When looking through half of all the exception thrown by git-python, I couldn't find a single one that would be too generic or not matching the situation in which they have been thrown.
Therefore I believe it's safe to assume this issue is old enough to not be relevant anymore in the current code base.

PR #300, opened 2015-06-25
Added NullHandlers to all loggers to prevent "No handler" messages

jnowell129:
When the code is run without setting up loggers, the loggers have no
handlers for the emitted messages. The logging module displays:
`No handlers could be found for logger "git.cmd"` on the
console. By adding a NullHandler (a no-op) the message disappears,
and doesn't affect logging when other handlers are configured.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -44,6 +44,7 @@ execute_kwargs = ('istream', 'with_keep_cwd', 'with_extended_output',
                   'output_stream')
 
 log = logging.getLogger('git.cmd')
+log.addHandler(logging.NullHandler())
 
 __all__ = ('Git', )
 
--- a/git/config.py
+++ b/git/config.py
@@ -32,6 +32,7 @@ __all__ = ('GitConfigParser', 'SectionConstraint')
 
 
 log = logging.getLogger('git.config')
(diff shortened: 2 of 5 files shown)


jnowell129:
So I see the CI build ran with Python 2.6. Is there a list of supported Python versions? I can add a NullHandler as described here, but I'm not sure whre you would want that to live: https://docs.python.org/release/2.6/library/logging.html#configuring-logging-for-a-library

jnowell129:
I'm not able to reproduce the error on my machine in Python 2.7. Also on the CI build, 'FOO' is totally in the exception. I'm not really sure what's supposed to be going on:

`GitCommandError: 'Error when fetching: fatal: protocol error: bad line length character: FOO' returned with exit code 2`

Byron:
Thanks a for your contribution!

I took the liberty to add the required NullHandler implementation myself. Besides, you are right, the `Foo` test is entirely unrelated to this PR, and needs additional attention.

Byron:
You can watch the development stream [on youtube](https://youtu.be/3tK_v1X5qxI).

*`GitPython #10 [PR 300 - add logging Null handlers]`*

[\thumb](https://youtu.be/3tK_v1X5qxI)

PR #305, opened 2015-06-29
Fix docstring of Repo.clone_from()

jonls:
Adds an additional newline in the comment of `Repo.clone_from()` in order for the docstring to be correctly parsed by Sphinx.


Byron:
Thank you !

PR #307, opened 2015-06-30
Add env parameter to Repo.clone_from() for setting environment variables

jonls:
Adds the optional keyword parameter `env` to `Repo.clone_from()`. The parameter is a dictionary containing the desired environment variables for the git clone invocation. The environment is applied to the temporary `Git` instance before calling `Repo._clone()`.


Byron:
Looks good, thank you !

PR #309, opened 2015-07-06
care tilde in include.path config

mihyaeru21:
According to the [git documentation](http://git-scm.com/docs/git-config), include.path is subject to tilde expansion.

> You can include one config file from another by setting the special include.path variable to the name of the file to be included. The included file is expanded immediately, as if its contents had been found at the location of the include directive. If the value of the include.path variable is a relative path, the path is considered to be relative to the configuration file in which the include directive was found. ***The value of include.path is subject to tilde expansion***: ~/ is expanded to the value of $HOME, and ~user/ to the specified user’s home directory. See below for examples.

So, I added to care code for tilde expansion.

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -386,7 +386,7 @@ class GitConfigParser(with_metaclass(MetaParserBuilder, cp.RawConfigParser, obje
             # We expect all paths to be normalized and absolute (and will assure that is the case)
             if self._has_includes():
                 for _, include_path in self.items('include'):
-                    if '~' in include_path:
+                    if include_path.startswith('~'):
                         include_path = os.path.expanduser(include_path)
                     if not os.path.isabs(include_path):
                         if not close_fp:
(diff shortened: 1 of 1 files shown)


Byron:
Thank you.

Do you think it would be safer to just expand `~` if it is the first character ?

mihyaeru21:
Yes.
I think using `include_path.startswith('~')` is safer than using `'~' in include_path`.
(e.g. zsh and bash expand `~/`, but they doesn't expand `/~`)

I will fix it.
Thank you!

Byron:
Cool, please leave a quick note here once it's done as I don't get notifications if a new commit comes in.
Thanks again.

mihyaeru21:
fixed!

Byron:
Thanks a lot !

mihyaeru21:
Thanks too!!

Issue #321, opened 2015-07-20
Command `git merge-base --is-ancestor` not working as expected


bwrsandman:
### Description

Using:
- git version 2.4.6
- gitdb==0.6.4
- GitPython==1.0.1

The use case for `git merge-base --is-ancestor` from git manpage:

> Check if the first <commit> is an ancestor of the second <commit>, and exit with status 0 if true, or with status 1 if not. Errors are signaled by a non-zero status that is not 1.

Related to https://github.com/gitpython-developers/GitPython/issues/169#issuecomment-72946021

The problem is that the output of `git merge-base --is-ancestor` comes from the return code and unlike without the `--is-ancestor` flag, this command doesn't return commit SHAs but is a simple yes-no check. The current implementation doesn't take into account the return code and the command cannot be usefully run.
#### revA is ancestor to revB

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
0
```

``` python
>>> repo.merge_base(revA, revB, is_ancestor=True)
[ ]
```
#### revA is not ancestor to revB

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
1
```

``` python
>>> repo.merge_base(revA, revB, is_ancestor=True)
[ ]
```
#### revA is not ancestor to revB and revB doesn't exist

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
fatal: Not a valid commit name $revB
128
```

``` python
>>> repo.merge_base(revA, revB, is_ancestor=True)
GitCommandError: 'git merge-base --is-ancestor $revA $revB' returned with exit code 128
stderr: 'fatal: Not a valid commit name $revB'
```
### Desired Behaviour

When the `kwarg` `is_ancestor` is used, don't return a list, but return a bool that is true if the return code is 0. Return False if return code is 1. Raise Error as before for all other cases

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
0
```

``` python
>>> repo.merge_base(revA, revB, is_ancestor=True)
True
```
#### revA is not ancestor to revB

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
1
```

``` python
>>> repo.merge_base(revA, revB, is_ancestor=True)
False
```
#### revA is not ancestor to revB and revB doesn't exist

``` bash
$ git merge-base --is-ancestor $revA $revB; echo $?
fatal: Not a valid commit name $revB
128
```

``` python
[...]

PR #322, opened 2015-07-21
Implement is_ancestor

bwrsandman:
Wrap `git merge-base --is-ancestor` into its own function because it acts
as a boolean check unlike base `git merge-base call`

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -504,6 +504,21 @@ class Repo(object):
 
         return res
 
+    def is_ancestor(self, ancestor_rev, rev):
+        """Check if a commit  is an ancestor of another
+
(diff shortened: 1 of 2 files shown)


Byron:
Thank you  !

Have you managed to get nosetest to work for you ? It should be easy to add a test to verify that 
* it returns True the rev is an ancestor
* it returns False if the rev is no ancestor
* it throws if there is an invalid revision

bwrsandman:
I am writing a test as we speak

bwrsandman:
Test done. Sorry I have no video of the coding ;)

Byron:
Thank you ! 

Looks great, even without the video :).

Closes #321

PR #324, opened 2015-07-22
gic {init,clone} --separate-git-dir is supported only since 1.7.5

roidelapluie:
Without this commit the update() function of a submodule does not work
in CentOS 6.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -137,7 +137,7 @@ class Submodule(util.IndexObject, Iterable, Traversable):
 
     @classmethod
     def _need_gitfile_submodules(cls, git):
-        return git.version_info[:3] >= (1, 7, 0)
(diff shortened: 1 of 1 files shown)


Byron:
Thank you !
I did my best to figure out when the feature was introduced, and found it difficult to find good documentation about it in the changelog.
Let's hope that patch-level 5 will work everywhere.

roidelapluie:
I did a git blame on https://github.com/git/git/blob/01977f46cb48e2894e335667eaabce274174ac5e/builtin/clone.c

Which shows https://github.com/git/git/commit/b57fb80a7d7d19102b31ab94a28ed43ea1ee07bb

Which is tagged 1.7.5.



In CentOS 6:
$ git --version
git version 1.7.1

Byron:
Thanks again for digging deeper than I did ! With that information I believe there is certainty that GitPython can now make the correct decisions.

PR #326, opened 2015-07-26
Ensure file resources are released

vokimon:
SymbolicReference leaked file descriptor resources when iterating on refs. unittest module complained:

symbolic.py:153: ResourceWarning: unclosed file <_io.TextIOWrapper name='...../gitrepo/.git/packed-refs' mode='rt' encoding='UTF-8'>
symbolic.py:148: ResourceWarning: unclosed file <_io.TextIOWrapper name='....../gitrepo/.git/packed-refs' mode='rt' encoding='UTF-8'>

### DIFF
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -88,25 +88,25 @@ class SymbolicReference(object):
         """Returns an iterator yielding pairs of sha1/path pairs (as bytes) for the corresponding refs.
         :note: The packed refs file will be kept open as long as we iterate"""
         try:
-            fp = open(cls._get_packed_refs_path(repo), 'rt')
-            for line in fp:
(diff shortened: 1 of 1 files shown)


Byron:
Thank you very much !

I am afraid there are many more issues of that kind buried in the code as I was wrongly assuming (back in the days) that an object destructor is deterministically called (such as in C++). This led to me failing to explicitly close files.

vokimon:
what about starting with a grep for 'open'?

Byron:
One will have to spread these efforts among GitDB and SMMap as well. Both should have mechanisms to release resources, but I doubt this works properly.

For now the best way to workaround these resource related issues is to use a `GitCmdObjectDB`, at least when working with non-ref/object data.

PR #329, opened 2015-07-27
Fix bug in tutorial, I'm not sure I'm right so please confirm

cool-RR:


Byron:
The `head` ref of the repository is automatically converted to the commit it points to. Therefore there is no need to be more explicit.
Maybe it helps to be explicit though.

cool-RR:
No, I tried that code and it didn't work. That's why I submitted the merge request.

Issue #340, opened 2015-08-19
Malformatted options


StyXman:
I'm using `git.Git` a lot for calling low level `git` commands; for instance, `git commit-tree`. In particular, I'm trying to use the `-m` option, but `Git.transform_kwargs()` converts `m='foo'` into `-mfoo`, which `git` rejects:

    git commit-tree -mfoo e53dc311b7e3ac674f3a6548a22113307a6af3ef -> 128; stdout: '<OUTPUT_STREAM>'; stderr: 'fatal: Not a valid object name -mfoo'

I'm not sure if the current formatting is like so for any particular reason, and I found about the `split_single_char_options` argument, but there seems to be no way to control that via the API.

Byron:
I totally agree and am not sure myself why it is implemented in that particular way. Changing the default might be a suitable fix (e.g. setting `split_single_char_options` could be turned on by default).

Would you like to submit a PR for this ? Chances are everything will just work fine after that, while fixing the issue.

StyXman:
Well, the unit tests don't pass in my machine (tests py2.7 [`FAILED (errors=10, failures=3)`] and 3.4 [`FAILED (errors=10, failures=2)`]), but the fix is simple enough. Maybe you can try it on your side? See #341.

PR #341, opened 2015-08-19
fix(cmd): make short options with arguments become two separate argum…

StyXman:
…ents for the executable.

So `git.Git.command(f=42)` becomes `git command -f 42` instead of `git command -f42`, which is not recognized by `git`. See #340.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -700,7 +700,7 @@ class Git(LazyMixin):
         finally:
             self.update_environment(**old_env)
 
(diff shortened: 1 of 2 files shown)


Byron:
Thank you !

Please note that the tests require some runup on the local repository, what's done exactly can be seen in the [`.travis.yml`](https://github.com/gitpython-developers/GitPython/blob/master/.travis.yml) file.

PR #354, opened 2015-09-28
Include 'timeout' parameter in Git execute

dpursehouse:
This feature enables to set a timeout while executing a git
command. After this timeout is over, the process will be killed.
If not explicitly specified, the default functionality will not
be affected.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -576,6 +576,8 @@ class Git(LazyMixin):
 
         if sys.platform == 'win32':
             cmd_not_found_exception = WindowsError
+            if timeout:
+                raise GitCommandError('"timeout" feature is not supported on Windows.')
         else:
             if sys.version_info[0] > 2:
                 cmd_not_found_exception = FileNotFoundError  # NOQA # this is defined, but flake8 doesn't know
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -42,7 +42,7 @@ from git.compat import (
 
 execute_kwargs = ('istream', 'with_keep_cwd', 'with_extended_output',
                   'with_exceptions', 'as_process', 'stdout_as_string',
-                  'output_stream', 'with_stdout', 'timeout')
+                  'output_stream', 'with_stdout', 'kill_after_timeout')
 
 log = logging.getLogger('git.cmd')
 log.addHandler(logging.NullHandler())
@@ -476,7 +476,7 @@ class Git(LazyMixin):
                 as_process=False,
                 output_stream=None,
                 stdout_as_string=True,
-                timeout=None,
+                kill_after_timeout=None,
                 with_stdout=True,
                 **subprocess_kwargs
                 ):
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for the PR ! The feature is much appreciated.

When looking at the implementation, I would wonder whether one should always use the same code-path, or have two ones. One will deal with the default (and common) case where `timeout=None` and be exactly what it was before, the other one will handle the timeout. Right now, I am afraid of performance degradation.

Additionally, GitPython claims to run and work on windows, which would stop being the case without additional work. To me it would be sufficient to just not provide a `timeout` parameter on windows, or explicitly raise with `this feature is only implemented for windows`.

What do you think ?

dpursehouse:
@Byron thanks for the feedback.  We're working on a couple of new commits to address the comments, and will upload them next week.

git/cmd.py
```diff
@@ -592,13 +602,45 @@ def execute(self, command,
...
+                if len(line.split()) > 0:
+                    local_pid = (line.split())[0]
+                    if local_pid.isdigit():
+                        child_pids.append(int(local_pid))
+            try:
+                os.kill(pid, SIGKILL)
+                for child_pid in child_pids:
+                    os.kill(child_pid, SIGKILL)
+                kill_check.set()    # tell the main routine that the process was killed
+            except OSError:
```
Byron: If `os.kill` throws `OSError` if the process to be killed doesn't exist, I fear that after `os.kill(pid, SIGKILL)`, some of the child processes might already be down as `pid`. Right now, failure to kill the first child process, will prevent the others from being downed as well. Maybe something like that would help:

```python
for child_pid in child_pids:
    try:
        os.kill(child_pid, SIGKILL)
    except OSError:
        pass
```

git/cmd.py
```diff
@@ -531,6 +533,12 @@ def execute(self, command,
 
         :param with_stdout: If True, default True, we open stdout on the created process
 
+        :param timeout:
+            To specify a timeout in seconds for the git command, after which the process
+            should be killed. This will have no effect if as_process is set to True. It is
+            set to None by default and will let the process run until the timeout is
+            explicitly specified.
```
Byron: I would love to see that this feature is not supported on windows.

git/cmd.py
```diff
@@ -531,6 +533,12 @@ def execute(self, command,
 
         :param with_stdout: If True, default True, we open stdout on the created process
 
+        :param timeout:
```
Byron: What do you think about renaming the parameter to something even more descriptive, like `kill_after_timeout` to be very clear about the `SIGKILL` nature of the operation.

git/cmd.py
Byron: It might be good to make clear that `SIGKILL` can have negative side-effects on a repository. For example, stale locks in case of `git gc` could render the repository incapable of accepting changes until the lock is manually removed.

Byron:
Thanks for the previous batch of changes. I hope you won't mind the additional comments I have left - just now I realized what it means to `SIGKILL` a git process.

In case you make any changes, I would be glad if you drop a comment as well as this will trigger a notification, of which I receive none if additional commits are pushed into the PR.

dpursehouse:
@Byron I just added some more commits to address your latest comments.  Thanks.

Byron:
Thanks a lot for your efforts !

PR #357, opened 2015-10-15
fix(cmd): fixed deadlock when stderr buffer overflow

rikdev:
Fixed deadlock when using stderr=PIPE in Popen and Git generates enough
output to a pipe such that it blocks waiting for the OS pipe buffer to
accept more data (see https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait)

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -310,11 +310,11 @@ class Git(LazyMixin):
             """Wait for the process and return its status code.
 
             :raise GitCommandError: if the return status is not 0"""
-            status = self.proc.wait()
-            if status != 0:
-                raise GitCommandError(self.args, status, self.proc.stderr.read())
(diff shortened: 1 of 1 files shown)


Byron:
Thank you very much ! It's quite amazing how long code like this could survive in a library (as old) as GitPython. (*It's probably because `AutoInterrupt` is already an error-path which shouldn't be called by default, and indeed is called rarely*).

Byron:
Currently I believe GitPython does handle process channels correctly. Maybe you have an example of where this is not the case ?

Those who control the process directly, may as well call `communicate()` themselves.

rikdev:
I use Git with Git LFS plugin. When clone a repository with the function git.Repo.clone_from, Git LFS outputs to stderr a list of downloaded files and overflow OS pipe buffer.

PR #368, opened 2015-12-02
Added CHECKING_OUT as a logged operation

### LINKED ISSUE #330: "Checking out files" not covered by operation names in RemoteProgress
When cloning a repository the following message is printed to stderr repeatedly: 
"Operation name u'Checking out files' unknown - skipping line 'Checking out files:   1% (216/11210)'".

1. Does it makes sense to add the operation 'Checking out files' to the list of operations in the RemoteProgress._parse_progess_line()? (see https://github.com/gitpython-developers/GitPython/blob/master/git/util.py#L162)
2. Shouldn't the error logged by using the logging mechanism and not printed to stderr directly? (https://github.com/gitpython-developers/GitPython/blob/master/git/util.py#L242)  

(Environment: Windows 7, Python 2.7.10, GitPython 1.0.1)

PonteIneptique:
As seen in #330, "Checking out files" is not covered in the operations for RemoteProgress. Here is a simple fix. I have not found where to add test for it, feel free to redirect me.

Closes #330

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -165,7 +165,7 @@ class RemoteProgress(object):
     Handler providing an interface to parse progress information emitted by git-push
     and git-fetch and to dispatch callbacks allowing subclasses to react to the progress.
     """
-    _num_op_codes = 8
+    _num_op_codes = 9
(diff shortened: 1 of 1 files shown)


Byron:
Thanks for this PR, and being considerate regarding the test suite. 

As I can't tell out of the back of my head where exactly this can be easily and concisely tested, I think it's good enough that Travis is happy.

PR #372, opened 2015-12-21
Fixed a non-Windows import

AWhetter:
signal.SIGKILL is not available on Windows so use signal.SIGTERM as a backup
when SIGKILL is not available.


Byron:
Thank you very much.
The failing test in travis must be a spurious one, and I *hope* it will work fine after the merge.

PR #397, opened 2016-03-15
Fix it's vs its in tutorial.rst

wagnermarkd:
Trivial change. "It's" is short for "it is", while "its" is the possessive form of "it".


Byron:
Thanks a bunch!

PR #398, opened 2016-03-15
Split diff line by '\t' for metadata and path

jonathanchu:
#### What's this PR do?
This protects against `.split(None)`, which uses consecutive whitespace as a separator and may overlook paths where a single space is the filename.  While having a filename consisting of only a single space is unlikely in practical usage, it is entirely possible.

For example, take this diff line:

screenshot 2016-03-15 14 18 42

Notice there is a tab after the `change_type` and a space for the path.  The deleted file is a file named ' ' (just one space) at the git root.  To replicate this issue, add and commit this file, then delete and commit.  This will produce the following output from `git diff`:

```
$ git diff --name-status <SHA1> <SHA2>
D
M       path/to/another/file.py
...
```

This would cause the initial `.split(None, 5)` to fail as it will count all consecutive whitespace as a separator, disregarding the ' ' (single space)  filename.


Byron:
Thanks for the elaborate explanation and the fix! Do you think it's feasible to add a test for this as well, maybe with a fixture?

jonathanchu:
And thanks for the quick response @Byron! :)  Sure thing, I was just looking into that after I created this PR actually.  

I'll take a look at the current tests tonight and will see if I can work something in.  I think it should be feasible and definitely worth it.

jonathanchu:
@Byron can you review my test and fixture in e328ffd, if this fits and is in line with the rest of the library?  Thanks!

Byron:
Thanks so much, it looks great!

jonathanchu:
:beers: :)

Issue #396, opened 2016-03-14
Fetching from remote with explicit refspec raises error incorrectly


rrei:
GitPython version: 1.0.2
git/remote.py
Line: 646
OS: Ubuntu 14.04LTS

Passing a refspec to git fetch through a remote, as in
`repo.remotes.origin.fetch("+refs/heads/*:refs/remotes/origin/*")`
will raise an error if the remote has no fetch refspec defined in the config:
```
AssertionError: Remote '%s' has no refspec set.
You can set it as follows: 'git config --add "remote.origin.fetch +refs/heads/*:refs/heads/*"'.
```
Also, note that the error message is incorrectly formatted because the remote's name is replaced only in the second line of the message (a %s appears in the first line of the final message).

The `_assert_refspec()` method should not be called if one or more refspecs were given as arguments to `fetch()`.

Byron:
Thanks for letting me know - the incorrectly formatted message should now be fixed.

However, I am not quite sure why there is no 'fetch' configuration provided here. In a [previous commit](https://github.com/gitpython-developers/GitPython/commit/4a771adc5352dd3876dd2ef3d0c52c8e803fc084) it is stated that gitpython needs this to work.

Just not making the assertion should therefore result in a more difficult to debug error. Maybe you can try removing the assertion in question to see if GitPython will work anyway ?

rrei:
I was testing this briefly and found out that the repo has no fetch refspec because I cloned it as a bare repo. Also, I commented the `_assert_refspec()` call, tried a fetch, and it worked like a charm. 

```
IPython 4.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import git

In [2]: r = git.Repo.clone_from('git://example.com/repos/aletheia.git', 'aletheia.git', branch='develop', bare=True)

In [3]: r.refs
Out[3]: 
[<git.Head "refs/heads/develop">,
 <git.Head "refs/heads/master">,
 <git.TagReference "refs/tags/0.1.0">]

In [4]: r.remotes.origin.fetch('+refs/heads/*:refs/remotes/origin/*')
Out[4]: 
[<git.remote.FetchInfo at 0x7feb9e88a940>,
 <git.remote.FetchInfo at 0x7feb9e88a998>]

In [5]: r.refs
Out[5]: 
[...]

PR #402, opened 2016-03-28
Remove assertion over fetch refspec when explicitly specified

rrei:
As we discussed in issue #396, the repo's fetch refspec is now only required if no explicit refspec was given as argument to `Remote.fetch()` or `Remote.pull()`.


Byron:
Thanks a lot for your contribution ! It's much appreciated, and if you need a release with it please let me know.

rrei:
Glad I could help out, even if only a tiny bit :)
In my code I worked around this by using `repo.git.fetch()` directly, but maybe I'll refactor it later to use the proper `Remote.fetch()` method. So there's no need to do a release just for this ;)

PR #405, opened 2016-04-06
Make sure .read() and friends always return bytes

nvie:
This had tripped me up when consuming from `.read(4096)` in a streaming fashion. All chunks will return bytes except the last one (the empty string), which is a unicode string in Python 3.

This fixes that to always return bytes, in both Python 2 and 3.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -355,7 +355,7 @@ class Git(LazyMixin):
         def read(self, size=-1):
             bytes_left = self._size - self._nbr
             if bytes_left == 0:
-                return ''
+                return b''
             if size > -1:
                 # assure we don't try to read past our limit
                 size = min(bytes_left, size)
@@ -374,7 +374,7 @@ class Git(LazyMixin):
 
         def readline(self, size=-1):
             if self._nbr == self._size:
-                return ''
+                return b''
 
             # clamp size to lowest allowed value
             bytes_left = self._size - self._nbr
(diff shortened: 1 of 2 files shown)


Byron:
Thanks @nvie , much appreciated !
It seems it works well in python 2.x, but apparently the test itself [has wrong expectations](https://travis-ci.org/gitpython-developers/GitPython/jobs/121179592#L476).
Do you think changing this line to `assert s.readline() == b''` will do it ?

nvie:
I've fixed the tests as you indicated. I had some trouble running the tests locally (I just tried running `tox`, but it failed on different tests, not this one). My local tests are still failing, not sure why. I'll await Travis' response now and adjust accordingly.

Thanks for this awesome project, @Byron!

nvie:
Seems all good now :)

Byron:
Thanks @nvie ! Testing locally usually fails when you are on a branch. Reason for this is that GitPython uses its own git repository for read-only testing, but doesn't really deal with the edge-cases arising from it. Bad practice, I know :).

Besides: I still remember when I was first studying this git-flow page that so neatly visualised the branching scheme that so many are using today ! Thus it's great to have your name in this project's commit history :) !

nvie:
Thanks, that's great to hear. As the matter of fact, we're starting to use more and more of GitPython, and we're seeing some wild repositories, so expect more contributions in the future. First one is #408 :)

Issue #364, opened 2015-11-16
Accessing files committed by initial commit with GitPython


sztankatt:
I'm using GitPython for working with a git repository. To list the files added/deleted/changed I use the following code:

    diff_files = []
    for parent in commit.parents:
        diff_files += [x.b_path for x in commit.diff(parent)]

However, the first commit doesn't have any parents, so this approach fails. Any ideas how might I do it in a different way?

Byron:
Maybe the questions are related, but do you think the *edit* portion of [this question on stackoverflow](http://stackoverflow.com/questions/33916648/get-the-diff-details-of-first-commit-in-gitpython) answers your question ?

```python
EMPTY_TREE_SHA = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"

if not i.parents:
    diff = i.diff(EMPTY_TREE_SHA, create_patch=True, **diffArgs)
else:
    diff = i.diff(i.parents[0], create_patch=True, **diffArgs)
```

Byron:
Closed due to inactivity. Please comment to get it reopened.

sztankatt:
Yes, sorry, since the I solved it similarly as you described in your comment.

PR #408, opened 2016-04-12
Add support for diffing against root commit

nvie:
This adds support for diff'ing against the initial commit (or any commit without parents, really). This is something that has classically been a bit hard to do since GitPython does not offer special support for this. (See #364, or http://stackoverflow.com/questions/33916648/get-the-diff-details-of-first-commit-in-gitpython)

I've opted to add this as another special case for the first argument to the `Commit.diff()` call, right next to the currently allowed `None`, the `Index` class, the commit SHA, etc. This plays nice since you'd never use this in combination with those existing options.

It enables:

    initial_commit.diff(NULL_TREE)

…which will produce a diff much like "git show <initial_commit>" would. Of course, it's possible to use a raw `repo.git.diff_tree()` call for this, but that would return a raw string, not a nice DiffIndex object. The workaround suggesting to use the magic "empty tree sha" also does not work, as the diff output is then in the "reverse direction".

Let me know what you think of this! Test cases included.


git/compat.py
```diff
@@ -46,6 +47,20 @@ def mviter(d):
...
+PRE_PY27 = sys.version_info < (2, 7)
+
+
+def safe_decode(s):
+    """Safely decodes a binary string to unicode"""
+    if isinstance(s, six.text_type):
+        return s
+    elif isinstance(s, six.binary_type):
+        if PRE_PY27:
+            return s.decode(defenc)  # we're screwed
```
Byron: As I have never had too many contributors, there is no guide yet. However, would you consider removing this comment , possibly in favour of explaining why python prior to py2.6 is a problem ?

git/diff.py
```diff
@@ -49,6 +49,7 @@ def diff(self, other=Index, paths=None, create_patch=False, **kwargs):
             If None, we will be compared to the working tree.
             If Treeish, it will be compared against the respective tree
             If Index ( type ), it will be compared against the index.
+            If the string 'root', it will compare the empty tree against this tree.
```
Byron: When using the string 'root', I am afraid that it would permanently block usage of a similarly named *treeish*, which could be the name of a tag or ref for example.
How would you circumvent this problem (or is it maybe not a problem at all) ?

git/compat.py
```diff
@@ -8,6 +8,7 @@
 # flake8: noqa
 
 import sys
+import six
```
Byron: I think I looked very carefully, but couldn't find a new dependency. Is six available in Py2.7 and onwards by default ?

Byron:
@nvie This is a nice one ! Thank so much for re-viving GitPython with your contributions.

As you can see, I have left a few notes on the code itself, and am looking forward to what you think !
Besides I very much like the `blame_incremental` feature you added, thank you !

Something you could consider is to add new features to the the `changes.rst`, that way you assure they are not forgotten when I at some point make a release. Speaking of which: If you for some reason should need one, please let me know. As the process is cumbersome, I am very much avoiding it unless there is a major bugfix to be distributed.

Last but definitely not least: Using the `six` library is totally fine, yet I hope it doesn't do anything in the background that could affect GitPython's overall performance. There is quite a vivid memory of me using the performance tests to get python 3 support without slowing things down to a crawl - at the end of this process just a few tweaks were needed and I was quite happy to not have introduced another dependency.

nvie:
(I've reverted my branch and force-pushed now to be a clean change again. Sorry for mixing this one up with the other feature.)

nvie:
> Something you could consider is to add new features to the the changes.rst, that way you assure they are not forgotten when I at some point make a release.

Will do!

> Speaking of which: If you for some reason should need one, please let me know. As the process is cumbersome, I am very much avoiding it unless there is a major bugfix to be distributed.

I can understand. Once everything is merged, I'll point our version to the latest master and make sure everything is stable. Then we can roll an update.

> Last but definitely not least: Using the six library is totally fine, yet I hope it doesn't do anything in the background that could affect GitPython's overall performance. There is quite a vivid memory of me using the performance tests to get python 3 support without slowing things down to a crawl - at the end of this process just a few tweaks were needed and I was quite happy to not have introduced another dependency.

I'm using six here mostly to get a convenient import target for things like `string_type`, `binary_type`, and `range`. I don't think six itself is a slowdown (it only provides a stable import target for things that have been moved around between Python versions AFAIK), but perhaps I'm missing something.

I'll address your feedback on this branch now.

git/compat.py
nvie: You're absolutely right here, this is inappropriate. Fixed in aae2a7328a4d28077a4b4182b4f36f19c953765b.

git/diff.py
nvie: That's a very good argument against it indeed. Rather than using a special string here, I've now opted for a special object, a little bit like the Index class. New usage:

    from git import NULL_TREE
    commit.diff(NULL_TREE)

Fixed in 82b533f86cf86c96a16f96c815533bdda0585f48. What do you think of this API?

nvie:
I've rebased this work against the latest master and added commits to address your feedback. Let me know what you think!

nvie:
OK, I think this is done now, test cases fixed, and dependency on `six` dropped.

git/compat.py
```diff
@@ -56,7 +56,9 @@ def safe_decode(s):
         return s
     elif isinstance(s, six.binary_type):
         if PRE_PY27:
-            return s.decode(defenc)  # we're screwed
+            # Python 2.6 does not support the `errors` argument, so we cannot
+            # control the replacement of unsafe chars in it.
+            return s.decode(defenc)
```
nvie: Btw, @Byron, do you still want to keep supporting Python 2.6? This message is echoed when you run the Python 2.6 tests already:

> _py26 installed: DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6_

Byron: Thanks for letting me know - under these circumstances it will be fine to drop support !

Byron: I just removed py2.6 support.

nvie: Cheers! 👍

Byron: BTW: I was about to merge a few PRs, but work is ...catching up earlier than expected. Will try again tonight, and if these delays become unbearable to you, I'd love to make you a collaborator. Just let me know what you think.
Cheers :) !

nvie: No worries, I've been there. Maintaining OSS libraries while also getting work done is a hard balance to strike. On the plus side, patching/maintaining this project is important for our team now, and we're committed to only invest in it more. So if you'd like to have a contributor on board, I'd be more than happy to help out there!

PR #409, opened 2016-04-13
Add incremental blame support

nvie:
This adds a sibling method to Repo's blame method:

    Repo.blame_incremental(rev, path, **kwargs)

This can alternatively be called using:

    Repo.blame(rev, path, incremental=True)

The main difference is that blame incremental is a bit more efficient and does not return the full file's contents, just the commits and the line number ranges.  The parser is a bit more straight-forward and faster since the incremental output format is defined a little stricter.

See the test case for an example.

I've also added support for passing `**kwargs` to git-blame, to enable passing `-C` or `-M` to it.

### DIFF
--- a/git/compat.py
+++ b/git/compat.py
@@ -8,6 +8,7 @@
 # flake8: noqa
 
 import sys
+import six
 
 from gitdb.utils.compat import (
     PY3,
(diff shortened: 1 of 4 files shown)


Byron:
It seems the commits here are also included in #408 and I might have commented on it. Please feel free to reply anywhere, even though I don't think there is any reason not to merge this one.
Thanks again for your contribution !

Issue #407, opened 2016-04-07
index.add() fails with "maximum recursion depth exceeded"


ppietrasa:
Here's how it looks:

> Traceback (most recent call last):
>   [...]
>   File "/go/ems/emsppi/dev/plspu/git_release_import/git_release_import.py", line 914, in import_patch
>     self.repo.index.add(files_to_add, fprogress=fprogress)
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/index/base.py", line 726, in add
>     entries_added.extend(self._entries_for_paths(paths, path_rewriter, fprogress, entries))
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/util.py", line 54, in wrapper
>     return func(self, *args, **kwargs)
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/index/util.py", line 84, in set_git_working_dir
>     return func(self, *args, **kwargs)
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/index/base.py", line 611, in _entries_for_paths
>     for filepath in self._iter_expand_paths(paths):
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/index/base.py", line 372, in _iter_expand_paths
>     for f in self._iter_expand_paths(glob.glob(abs_path)):
>   [...]
>   File "/go/ems/emsppi/dev/plspu/git_release_import/ext/git/index/base.py", line 372, in _iter_expand_paths
>     for f in self._iter_expand_paths(glob.glob(abs_path)):
>   File "/gow/ems/emsppi/local/lib/python2.7/glob.py", line 27, in glob
>     return list(iglob(pathname))
>   File "/gow/ems/emsppi/local/lib/python2.7/glob.py", line 64, in iglob
>     for name in glob_in_dir(dirname, basename):
>   File "/gow/ems/emsppi/local/lib/python2.7/glob.py", line 82, in glob1
>     names = filter(lambda x: x[0] != '.', names)
>   File "/gow/ems/emsppi/local/lib/python2.7/glob.py", line 82, in <lambda>
>     names = filter(lambda x: x[0] != '.', names)
> RuntimeError: maximum recursion depth exceeded in cmp
>
> the files being added contained a file  "[.exe" amongs them

Amongst the `files_to_add` there was a file called `[.exe` which I believe was the cause of the problem.

[...]

PR #410, opened 2016-04-13
fix(index): avoid recursing endlessly in add()

ppietrasa:
Issue #407


Byron:
Thank you very much, @ppietrasa !

PR #412, opened 2016-04-14
Fix diff patch parsing

nvie:
**NOTE:** This PR builds on #408, so please merge that one first.

---

The a_path and b_path cannot reliably be read from the first diff line as it's ambiguous.  From the git-diff manpage:

  > _"The a/ and b/ filenames are the same unless rename/copy is involved. Especially, **even for a creation or a deletion**, /dev/null is not used in place of the a/ or b/ filenames."_

This patch changes the a_path and b_path detection to read it from the more reliable locations further down the diff headers.  Two use cases are fixed by this:

  - As the man page snippet above states, for new/deleted files the a or b path will now be properly None.
  - File names with spaces in it are now properly parsed.

Working on this patch, I realized the --- and +++ lines really belong to the diff header, not the diff contents.  This means that when parsing the patch format, the --- and +++ will now be swallowed, and not end up anymore as part of the diff contents.  The diff contents now always start with an @@ line.

This may be a breaking change for some users that rely on this behaviour. However, those users could now access that information more reliably via the normal Diff properties a_path and b_path now.

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -5,9 +5,23 @@ Changelog
 1.0.3 - Fixes
 =============
 
-* `Commit.diff()` now supports diffing the root commit via `Commit.diff(NULL_TREE)`.
-* `Repo.blame()` now respects `incremental=True`, supporting incremental blames.  Incremental blames are slightly faster since they don't include the file's contents in them.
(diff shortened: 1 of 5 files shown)


nvie:
Going to rebase this against the latest `master`, to make this more readable.

PR #414, opened 2016-04-19
Add support for getting "aware" datetime info

nvie:
This adds 2 properties to commits.  Their values are derived from the existing data stored on them, but this makes them more conveniently queryable:

- authored_datetime
- committed_datetime

These return "aware" datetimes, so they are effectively companions to their raw timestamp equivalents, respectively `authored_date` and `committed_date`.

These datetime instances are convenient structures since they show the author-local commit date and their UTC offset.

(I've implemented a tzinfo subclass here manually to avoid adding a dependency to dateutil or pytz here.)

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -21,7 +21,8 @@ from .util import (
     Serializable,
     parse_date,
     altz_to_utctz_str,
-    parse_actor_and_date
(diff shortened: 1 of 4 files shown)


Byron:
Thank you, looks good to me !

PR #415, opened 2016-04-19
Fix diff patch parser for paths with unsafe chars

nvie:
This specifically covers the cases where unsafe chars occur in path names, and git-diff -p will escape those.

From the git-diff-tree manpage:

> 3\. TAB, LF, double quote and backslash characters in pathnames are represented as `\t`, `\n`, `\"` and `\\`, respectively. If there is need for such substitution then the whole pathname is put in double quotes.

This patch checks whether or not this has happened and will unescape those paths accordingly.

One thing to note here is that, depending on the position in the patch format, those paths may be prefixed with an `a/` or `b/`.  I've specifically made sure to never interpret a path that actually starts with `a/` or `b/` incorrectly.

Example of that subtlety below.  Here, the actual file path is `b/normal`.  On the diff file that gets encoded as `b/b/normal`.

     diff --git a/b/normal b/b/normal
     new file mode 100644
     index 0000000000000000000000000000000000000000..eaf5f7510320b6a327fb308379de2f94d8859a54
     --- /dev/null
     +++ b/b/normal
     @@ -0,0 +1 @@
     +dummy content

Here, we prefer the `---` and `+++` lines' values.  Note that these paths start with `a/` or `b/`.  The only exception is the value `/dev/null`, which is handled as a special case.

Suppose now the file gets moved `b/moved`, the output of that diff would then be this:

     diff --git a/b/normal b/b/moved
     similarity index 100%
     rename from b/normal
     rename to b/moved

We prefer the `rename` lines' values in this case (the `diff` line is always a last resort).  Take note that those lines are not prefixed with `a/` or `b/`, but the ones in the `diff` line are (just like the ones in `---` or `+++` lines).

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -22,6 +22,20 @@ __all__ = ('Diffable', 'DiffIndex', 'Diff', 'NULL_TREE')
 NULL_TREE = object()
 
 
+def decode_path(path, has_ab_prefix=True):
+    if path == b'/dev/null':
+        return None
+
+    if path.startswith(b'"') and path.endswith(b'"'):
+        path = path[1:-1].decode('string_escape')
+
+    if has_ab_prefix:
+        assert path.startswith(b'a/') or path.startswith(b'b/')
+        path = path[2:]
+
+    return path
+
+
 class Diffable(object):
 
     """Common interface for all object that can be diffed against another object of compatible type.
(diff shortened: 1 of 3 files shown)


Byron:
Thank @nvie , I particularly like your test !
Even though I didn't particularly verify it's truly exhaustive, it does look good to me.

PR #429, opened 2016-05-10
Update requirements doc

inderpreet99:
* Added the Python version requirement
* Updated URLs for other requirement
* Removed whitespace


Byron:
Thank you very much !

PR #433, opened 2016-05-18
Need spaces in Emacs style encoding comment

StevenWColby:
Although it's hard to see, PEP-0263 does have ws delimiting the 'coding' string.
This commit will fix the root cause of (at least) one bug: https://lists.fedoraproject.org/archives/list/eclipse-sig@lists.fedoraproject.org/thread/5XQ5JRHG6DPPMGRDU7TA2AO4EYS2H7AG/


Byron:
Thanks a lot for your contribution !

PR #441, opened 2016-05-24
Split lines by new line characters

jonathanchu:
#### What's this PR do?
Fixes a bug where there could be special characters in the commit message, causing the line splits to be improperly detected when parsing the blame metadata.

This commit opts to split lines by the new line character instead of letting `splitlines()` do this. This helps catch the issue when there are special characters in the line, particular the commit summary section.

#### How should this be manually tested?
`git/test/test_repo.py` has been modified to catch this edge case. Please do run all tests or this one test in particular.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -682,10 +682,12 @@ class Repo(object):
         data = self.git.blame(rev, '--', file, p=True, incremental=True, stdout_as_string=False, **kwargs)
         commits = dict()
 
-        stream = iter(data.splitlines())
+        stream = iter(data.split(b'\n'))
         while True:
             line = next(stream)  # when exhausted, casues a StopIteration, terminating this function
-
+            if line.strip() == '':
+                # Skip over empty lines
+                continue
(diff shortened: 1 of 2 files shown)


git/repo/base.py
```diff
@@ -682,10 +682,12 @@ def blame_incremental(self, rev, file, **kwargs):
...
         commits = dict()
 
-        stream = iter(data.splitlines())
+        stream = iter(data.split(b'\n'))
         while True:
             line = next(stream)  # when exhausted, casues a StopIteration, terminating this function
-
+            if line.strip() == '' or line.strip() == b'':
+                # Skip over empty lines
+                continue
```
nvie: Ohh, this actually has nothing to do with the ^M chars in the fixture. It's because of this:

```python
>>> s = 'foo\nbar\n'
>>> s.splitlines()
['foo', 'bar']
>>> s.split('\n')
['foo', 'bar', '']
               ^^
```

nvie:
Thanks, this is now pulled!

Issue #444, opened 2016-05-26
Progress not showing up before the last second


guyzmo:
I've implemented the `Progress` class to implement a nice progress bar:

* https://github.com/guyzmo/git-repo/blob/master/git_repo/services/service.py#L46-56

I'm then using it, for example here:

* https://github.com/guyzmo/git-repo/blob/master/git_repo/services/service.py#L196-202

And then, when I clone a very large repository (like mozilla's gecko):

    % git hub clone mozilla/gecko-dev
    Pulling from github |████████████████████████████████|
    Successfully cloned `https://github.com/guyzmo/git-repo` into `./git-repo`!

the progress bar is "progressing" (i.e. incrementing from 0 to 100%), but it shows up only at the very end of the fetching, after hanging for a long while. I started to debug it, and it looks like it's related to subprocess stuff. 

Though, I might be implementing it wrong, which is why I'm asking here while I'm still debugging that issue.

Byron:
As far as I know, the progress line parsing is immediate, and thus will produce results in real time, similar to what you see on the command-line.
To understand what's the problem in your case, I recommend having a look at the [parsing code](https://github.com/gitpython-developers/GitPython/blob/b9a7dc5fe98e1aa666445bc240055b21ed809824/git/util.py#L185) to understand how exactly it works.
As an alternative step that you seem to have taken already, I would believe that just printing the remote progress arguments will indicate they are indeed provided in real-time.
If that's the case indeed, the problem would lie with the way the progress bar is used.

However, I am interested in your findings, maybe there is something unexpected here.

guyzmo:
There's definitely something fishy going on, I have replaced my "progress bar" instance with the following code:

and then I ran the command. I made an asciicinema output of the command to better see the pausing, followed by the output:

https://asciinema.org/a/cwoo3r1gnziyxgycbybd1p485

which is not outputting at the same rate as a raw git clone command:

http://asciinema.org/a/0u002zielzbpprrj6e5h20hpa

I already looked at the parsing code, and I agree it's looking totally fine, and it *should* work as intended. So maybe, a good idea would be to try to replicate the issue through a minimal snippet, to see whether my code could be interfering (which I believe should not, but who knows?).

I'll also try to add printouts in the parser to see _where_ it's blocking.

guyzmo:
Ok, I've added a bunch of printouts, along with a timer to find where the pausing is happening (cf below).

[...]

PR #445, opened 2016-05-26
Changing warning to debug logging, to avoid warning showing off when nothing's wrong

guyzmo:
cf issue #444 .


Byron:
Thank you !

PR #446, opened 2016-05-26
Adding support for git remote set-url/get-url API to Remote

guyzmo:
Both commands enable handling of a little known feature
of git, which is to support multiple URL for one remote.
You can add multiple url using the `set_url` subcommand of
`git remote`. Though listing them is also handy, so there's
a nice method to do it, using the configuration.

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -477,10 +477,10 @@ class Remote(LazyMixin, Iterable):
     def urls(self):
         """:return: Iterator yielding all configured URL targets on a remote
         as strings"""
-        scmd = 'get-url'
-        kwargs = {'insert_kwargs_after': scmd}
-        for url in self.repo.git.remote(scmd, self.name, all=True, **kwargs).split('\n'):
-            yield url
+        remote_details = self.repo.git.remote("show", self.name)
+        for line in remote_details.split('\n'):
+            if '  Push  URL:' in line:
+                yield line.split(': ')[-1]
 
     @property
     def refs(self):
--- a/git/remote.py
+++ b/git/remote.py
@@ -434,6 +434,54 @@ class Remote(LazyMixin, Iterable):
             yield Remote(repo, section[lbound + 1:rbound])
         # END for each configuration section
 
+    def set_url(self, new_url, old_url=None, **kwargs):
+        """Configure URLs on current remote (cf command git remote set_url)
+
+        This command manages URLs on the remote.
+
+        :param new_url: string being the URL to add as an extra remote URL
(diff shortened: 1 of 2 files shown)


guyzmo:
might need to write some tests for those?

Byron:
Oh yes please ! It should be relatively straightforward to deduce how to write tests for this by looking at the existing ones. Usually there are helper functions to setup even more complex fixtures.

Byron:
Oh, before I forget, could you also look at how existing methods are documented and match the documentation style ? This would include documenting all parameters, for example.

guyzmo:
I might, but sadly I'm now in the train, going away for four days, during which I'm on a crappy tablet on which I cannot work. Though, I very much would like to see this new methods included in the next patchset so I can remove the monkey patching from my own codebase as well :-)

cf https://github.com/guyzmo/git-repo/blob/master/git_repo/services/service.py#L25-43

guyzmo:
I've been trying to write tests for those, but I'm having issue making it right…

Would you be available in any chat so we could discuss this slightly more efficiently, and avoid "chatting" over issues comments?

guyzmo:
So, I've squashed my commit with a new one, featuring a few changes:

* The `set_url` method is a direct wrapper for `git remote set-url` ;
* the `urls` property is providing an iterator over the remote's URLs (I followed how you named `refs`)

Then I've added a couple of convenience methods, to avoid adding awkward arguments to control `set_url`:

* the `add_url` method is there so you can add an URL to a remote
* and `delete_url` method is to delete one.

Finally, I've added a test covering all the new code, with a couple of edge cases, and wrote pydoc for those functions. I hope it's how you like it, mates!

guyzmo:
I rolled back the `urls` property, because that was failing on Travis, which isn't using latest git.

The `git remote get-url` API has been introduced in [`git 2.7.0`](https://git-scm.com/docs/git-remote/2.7.0), and wasn't available [before](https://git-scm.com/docs/git-remote/2.6.3).

guyzmo:
can you ping me when you think you'll be able to merge, and publish on pypi¸ and what version that will be? So I can update my setup.py accordingly ☺

Byron:
If @nvie doesn't beat me to it, I will get the new release on the way on my next maintenance slot. Maybe tomorrow ... .

guyzmo:
thank you ;-)

preparing my own code to be ready to point to the right version! :-)

nvie:
I'll add 2 more bug fixes tonight to this release. After I'm done with those, I'll roll a new release :)

PR #447, opened 2016-05-26
Use proper syntax for conditional expressions.

Xender:
Hello. :)

This change replaces abusing the "short-circuit" property of logical operations (`cond and A or B`) with dedicated conditional expression syntax (`A if cond else B`).

Just please double-check that I didn't get the `and-or` short-circuitry logic backwards or wrong in some other way! ;)


Byron:
Looks very good, thank you !

PR #451, opened 2016-05-29
Fix traceback because _seen_ops is not initialised

barry-scott:
must call the base class __init__

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -320,10 +320,10 @@ class CallableRemoteProgress(RemoteProgress):
     
     def __init__(self, fn):
         self._callable = fn
-        
+        RemoteProgress.__init__(self)
+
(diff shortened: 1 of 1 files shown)


Byron:
Argh, thanks ! That code is not called by the test-suite yet :( .

git/util.py
```diff
@@ -320,10 +320,10 @@ class CallableRemoteProgress(RemoteProgress):
     
     def __init__(self, fn):
         self._callable = fn
-        
+        RemoteProgress.__init__(self)
```
nvie: This should be `super(CallableRemoteProgress, self).__init__()` instead.

Byron: Thanks - it's done.

nvie:
Cheers!

PR #452, opened 2016-05-29
Pr cmd raise with stderr on error

barry-scott:
I have seen the stderr for both pull() and push() in GitCommandError with this patch.


Byron:
Hey Barry,

Thanks a bunch for your contribution ! Could you also have a look at travis please ? It seems that a few tests are failing. They probably need to be adjusted or something did break indeed.

You will find everything you need to run the tests in the `.travis.yml` file in the root of this project, alternatively you should be able to run `nosetests git/test`, which is what I do quite a lot. `flake8` is easily forgotten here, but can also cause travis to not green-light the PR.

Thanks

git/remote.py
```diff
@@ -570,21 +570,36 @@ def _get_fetch_info_from_stderr(self, proc, progress):
...
+        error_message = None
+        stderr_text = None
+
         for line in proc.stderr:
             line = force_text(line)
             for pline in progress_handler(line):
                 if line.startswith('fatal:') or line.startswith('error:'):
-                    raise GitCommandError(("Error when fetching: %s" % line,), 2)
+                    error_message = "Error when fetching: %s" % (line,)
+                    break
```
Byron: Are you sure it' s enough to break after the first line indicating a problem ? My intuition is that the lines that follow on stderr contain context, that you probably want to catch as well.

git/util.py
```diff
@@ -173,13 +173,17 @@ class RemoteProgress(object):
...
     re_op_absolute = re.compile(r"(remote: )?([\w\s]+):\s+()(\d+)()(.*)")
     re_op_relative = re.compile(r"(remote: )?([\w\s]+):\s+(\d+)% \((\d+)/(\d+)\)(.*)")
 
     def __init__(self):
         self._seen_ops = list()
         self._cur_line = None
+        self._error_lines = []
+
+    def get_stderr(self):
+        return '\n'.join(self._error_lines)
```
Byron: I recommend providing the highest-value data available, which in this case in an array of lines. There should be no assumption about how the user will evaluate that data.
As in GitPython in the very majority of cases there is no `get_` prefix, this accessor would be called something like `error_lines()`.

git/util.py
```diff
@@ -190,6 +194,10 @@ def _parse_progress_line(self, line):
         # Counting objects: 4, done.
         # Compressing objects:  50% (1/2)   \rCompressing objects: 100% (2/2)   \rCompressing objects: 100% (2/2), done.
         self._cur_line = line
+        if len(self._error_lines) > 0 or self._cur_line.startswith( ('error:', 'fatal:') ):
```
Byron: I believe this would be a breakage in the API, as now lines that would previously be passed to update, will be caught in the base-class. Existing sub-types could depend on errors being handed to `update()`.
Do you think that makes sense, or is there another reason this has to be done in the base-class ?

git/remote.py
barry-scott: It works because of the 2nd break at line 591.

Also I see this work in testing.

git/util.py
barry-scott: Once you see error or fatal the progress either never started or will not continue.

The parsing will never pass an error to update. It only passes data from lines that pattern match against
a "progress" line.

barry-scott:
Found flake8 in pip and used to to fix the complaints.
python3 -m pip install flake8

Took a moment to figure out where nosetests came from.
dnf install python3-nose

I see errors from running the tests I'll work on there a bit later. may be tomorrow before I can push a update.

Byron:
Good to hear you can now test locally !
Please note that it's advised to run only the tests you are interested in (e.g. `nosetests git/test/test_remote.py`, as some could fail if they are run on a branch which is not named master.
One good way to workaround is to just work on master.
Next thing to do is to execute the commands that [you find in the .travis.yml](https://github.com/gitpython-developers/GitPython/blob/master/.travis.yml#L21) file to assure the pre-requisites are met. Reason for this requirement is that the tests actually use their own repository (which violates sandboxing).

With all that in place, you should reach a point where all tests run successfully on your own machine, in case everything works.

barry-scott:
I have instrumented to try and find out what is happening in one of the test failures.

What I see is this command that fails with 128.

git clone --progress --separate-git-dir=/home/barry/tmpdir/test_renamescm7sp1x/parent/.git/modules/mymodules/myname -v /home/barry/wc/git/GitPython/git/ext/gitdb/gitdb/ext/smmap /home/barry/tmpdir/test_renamescm7sp1x/parent/mymodules/myname

Notice the /home/barry/wc/git/GitPython/git/ext/gitdb/gitdb/ext/smmap path. 
I have /home/barry/wc/git/GitPython/git/ext/gitdb as am empty directory.
Where does the other files come from in that folder?

barry-scott:
Sorry I missed your reply about testing.

Work on master?! But isn't that a git sin?

Are you saying that I need to run all the commands in the install: section?

barry-scott:
This failure is becuase the GIT_SSH is not making it in to Popen call.
There is odd code that seems to work around a expected failed that I do not understand in the test as well. Taken from tst_git.py around line 205:

            with rw_repo.git.custom_environment(GIT_SSH=path):
                try:
                    remote.fetch()
                except GitCommandError as err:
                    if sys.version_info[0] < 3 and sys.platform == 'darwin':
                        assert 'ssh-origin' in str(err)
                        assert err.status == 128
                    else:
                        assert 'FOO' in str(err), 'err is %r' % (err,)
                        assert err.status == 2

Why expect FOO to be missing an python 2 and darwin?
Its missing with python3 and linux for me.

$ nosetests-3.4 git/test/test_git.py 
......Test TestGit.test_environment failed, output is at '/home/barry/tmpdir/test_environment74vf1sim'
F...........
FAIL: test_environment (git.test.test_git.TestGit)
  File "/home/barry/wc/git/GitPython/git/ext/gitdb/gitdb/test/lib.py", line 87, in wrapper
    return func(self, path)
  File "/home/barry/wc/git/GitPython/git/test/test_git.py", line 213, in test_environment
    assert 'FOO' in str(err), 'err is %r' % (err,)
nose.proxy.AssertionError: err is GitCommandError(['git', 'fetch', '-v', 'ssh-origin'], 128, b'')
-------------------- >> begin captured stdout << ---------------------
failing_script is '/home/barry/tmpdir/test_environment74vf1sim/failing-script.sh'

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
git.cmd: INFO: env GIT_SSH=None
git.cmd: INFO: env GIT_SSH=None
git.cmd: INFO: env GIT_SSH=None
git.cmd: DEBUG: AutoInterrupt wait stderr: b''
--------------------- >> end captured logging << ---------------------

Ran 18 tests in 0.551s

FAILED (failures=1)

barry-scott:
Can you try this and see if the tests pass for you?

I fixed the test that looked fro FOO, test case error.

Given I'm not on master I see that test fail as well.

I am seeing some odd errors from test_repo, but suspect that I have triggered a bug somewhere else.
My .git/logs/refs/heads/pr-cmd-raise-with-stderr-on-error may be corrupt; I commited using GitPython.

Byron:
Please see if my merge is still working. If not, please add a test that fails unless the fix you provide is applied.
Thanks for your contribution.

PR #454, opened 2016-05-30
Fix bug in diff parser output

nvie:
The diff `--patch` output parser was missing some edge case where Git would encode non-ASCII chars in path names as octals, but these weren't decoded properly.

    \360\237\222\251.txt

Decoded via utf-8, that will return:

    💩.txt

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -5,6 +5,8 @@ Changelog
 2.0.4 - Fixes
 =============
 
+* Fix: non-ASCII paths are now properly decoded and returned in
+  ``.diff()`` output
 * Fix: `RemoteProgress` will now strip the ', ' prefix or suffix from messages.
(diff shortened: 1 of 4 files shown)


nvie:
@Byron This is a bug we've seen in production for some time now. Happy for me to release 2.0.4 with this fix right now?

Byron:
No objections, please feel free to make a release.
Something I usually do as well is to keep the milestones in sync, and thus create a new one with the upcoming release, and move all open issues from the just-released milestones into it. The latter can then be closed officially, which will then become another way for the inclined audience to see which issues/PRs made it in there.
Don't feel obliged, as I can do it when I see a disparity, but recently I noticed that I probably have made some mistakes while doing so after the fact.

In any case, thanks for your contribution :) !

nvie:
Sure, will do!

nvie:
2.0.4 is now up on PyPI :)

PR #455, opened 2016-05-30
Skip test that always fails on Travis CI

nvie:
What it says on the tin :)

Almost all tests on branches seem to fail on this one on Travis.


Byron:
Hmm, interesting. I wasn't aware of tests that always fail - all I see is that it works consistently well (?).
https://travis-ci.org/gitpython-developers/GitPython/jobs/133965971
Thanks for the clarification.

PR #456, opened 2016-05-30
Add test case as example of Git commit with invalid data

nvie:
This is a real commit from the microjs.com open source project, see https://github.com/madrobby/microjs.com/commit/7e8457c17850d0991763941213dcb403d80f39f8, which is declared to be encoded in UTF-8, but contains invalid bytes.

This makes GitPython choke on it while decoding.  Rather than choking, this should instead accept the error and replace the invalid bytes by the � (\x80) char.


Byron:
Good catch ! Thanks for the PR !

PR #460, opened 2016-06-01
Make sure os is not even partly destroyed

ddanier:
I had problems with errors like:
Exception TypeError: "'NoneType' object is not callable" in > ignored

After much digging I found, that os.kill is None in this case while os still exists. No clue how Python managed to get this state, but my change fixes the problem.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -287,7 +287,7 @@ class Git(LazyMixin):
                 return
 
             # can be that nothing really exists anymore ...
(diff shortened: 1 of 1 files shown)


Byron:
Thank you very much ! It's a shame python is so unreliable when the process is on it's way down, and even though I believe they have good reasons for that, just removing standard library functions is certainly a terrible way to communicate these.

PR #467, opened 2016-06-07
Don't choke on (legitimately) invalidly encoded Unicode paths

nvie:
We've come across path names that contain bytes that are invalid in UTF-8 encoded strings, even though they're very rare. My assumption here is these commits have been created by an old (buggy?) version of Git, and now live in the tree objects with this data. Since we return only unicode strings for the `a_path` and `b_path` properties, we're not able to decode this string and thus choke when asking for the diff.

This PR fixes that by using "replace" semantics when decoding. This will effectively replace the illegal bytes `\200` (or `\x80`) by \ufffd (= �).

### Follow-up discussion

However, this also means that if you would want to git-blame this file, there's no good way of referencing this path, since it's inherently a bytes path. Normally, when we pass unicode paths to git-blame via GitPython's blame API, the paths get converted to UTF-8 right before issuing the external command. But there's no way of getting the original bytes back after the "replace" operation happened.

Example:

- The input path is `b'illegal-\x80.txt'` (containing illegal byte `\x80`)
- When decoded to UTF-8 with characters replaced, we get the unicode string `u'illegal-\ufffd.txt'` (= "illegal-�.txt")
- When encoding that in UTF-8, we find `b'illegal-\xef\xbf\xbd.txt'`

When we next pass `illegal-\xef\xbf\xbd.txt` to git-blame, it will not be able to find this path. Perhaps it would be a good idea to not only return the decoded path strings, but also provide access to the raw bytes found, i.e. by exposing `a_rawpath` and `b_rawpath`, which would always be bytes? That way, you could still have the friendly "unicode paths" for most use cases, but use bytes if you need to speak the language of Git more accurately.

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -404,15 +404,15 @@ class Diff(object):
             a_mode = old_mode or deleted_file_mode or (a_path and (b_mode or new_mode or new_file_mode))
             b_mode = b_mode or new_mode or new_file_mode or (b_path and a_mode)
             index.append(Diff(repo,
-                              a_path and a_path.decode(defenc),
-                              b_path and b_path.decode(defenc),
+                              a_path and a_path.decode(defenc, 'replace'),
+                              b_path and b_path.decode(defenc, 'replace'),
                               a_blob_id and a_blob_id.decode(defenc),
                               b_blob_id and b_blob_id.decode(defenc),
                               a_mode and a_mode.decode(defenc),
                               b_mode and b_mode.decode(defenc),
                               new_file, deleted_file,
-                              rename_from and rename_from.decode(defenc),
-                              rename_to and rename_to.decode(defenc),
+                              rename_from and rename_from.decode(defenc, 'replace'),
+                              rename_to and rename_to.decode(defenc, 'replace'),
                               None))
 
             previous_header = header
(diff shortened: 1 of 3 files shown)


Byron:
@nvie Thanks for bringing this up ! I believe it's one of the worst decisions made in GitPython to attempt to decode everything using UTF-8 (or sometimes using whatever encoding Git hints us at). Especially for paths, doing this is plain wrong as the filesystem may use arbitrary encodings. This is the reason Rust uses [OsStrings](https://doc.rust-lang.org/std/ffi/struct.OsString.html) to represent these, as opposed to [Strings](https://doc.rust-lang.org/std/string/struct.String.html) which need to be valid UTF-8.
Unfortunately, back in the days Rust wasn't on my Radar, and also I didn't know what I know now.

That said, fixing the API in that regard might be a somewhat major undertaking certainly worth making. Without a major version upgrade, the proposed adjustment might be good enough.

The question is, what you as an API user think about these options - I'd be happy to hear your thoughts.

nvie:
I don't think it's too bad in practice. In most parts where files hit the UI, you'd want to use a `decode(errors='replace')`'ed version of the raw bytes anyway. If there are invalidly encoded bytes, it's really just fine to see the 	� chars in there. So I don't think it's too bad that the result of a Diff returns unicode strings for the paths. In practice, they work in >99.9% of all cases.

However, sometimes there just is a need to use the raw bytes, especially when the result of command A (say, a diff), defines the input of command B (say, a blame). Only in these cases, you need to have the exact bytes. I would still show the `decode(errors='replace')`'ed version of the path in the UI, but in order to ask Git for a blame on that file, I need access to the raw version.

I'll have a stab at adding those extra raw paths as properties.

yarikoptic:
I guess related:  https://github.com/gitpython-developers/GitPython/pull/505  -- I have tried to run tests using TMPDIR with unicode in it...

Issue #464, opened 2016-06-06
GitPython Commit corrupts .git/logs/refs/heads/pr-cmd-raise-with-stderr-on-error


barry-scott:
I think that the GitPython commit is causing a corruption of this file:
    .git/logs/refs/heads/pr-cmd-raise-with-stderr-on-error

I think it is supposed to be 1 line per commit with the first line of the commit at the end.
But I have all of the log message in the file that GitPython attempts to parse and errors out on.

Attached in the full file.

[pr-cmd-raise-with-stderr-on-error.txt](https://github.com/gitpython-developers/GitPython/files/300506/pr-cmd-raise-with-stderr-on-error.txt)

barry-scott:
I have put in a clean pull request that fixes this bug.

Byron:
Will be fixed by #469.

barry-scott:
Fixed by #469

PR #469, opened 2016-06-13
FIx issue #464 corrupt log file

barry-scott:
Must only append the first line of the commit to the log
Otherwise the log file is not parsable by GitPython.


Byron:
Do you think it's possible for you to write a test which goes red if the change is not there ? The [current test-suite](https://github.com/gitpython-developers/GitPython/blob/5c8ff218cb3ee5d3dd9119007ea8478626f6d2ce/git/test/test_reflog.py#L18) should be a good starting-point for a fixture-based test.

barry-scott:
Please don't wait for me to write the test. I have a lot of work on my plate at moment.
Maybe merge this and leave the bug open awiting a test case?

How I detected the bug was:

1. Use GitPython to commit to a repo - causing the fault
2. Use GitPython to scan the log ref file - detecting the fault

barry-scott:
I'm not sure how to write the tests you want.

Can you point to docs on what is required?

git/cmd.py
```diff
@@ -629,7 +636,13 @@ def execute(self, command,
...
         def _kill_process(pid):
             """ Callback method to kill a process. """
-            p = Popen(['ps', '--ppid', str(pid)], stdout=PIPE)
+            if sys.platform == 'win32':
+                CREATE_NO_WINDOW = 0x08000000
+                creationflags = CREATE_NO_WINDOW
+            else:
+                creationflags = None
+
+            p = Popen(['ps', '--ppid', str(pid)], stdout=PIPE, creationflags)
```
Byron: I am afraid that using `creationflags` as positional argument doesn't actually end up in the correct spot. Also I thought that positional arguments after keyword args are not allowed.

git/cmd.py
```diff
@@ -609,6 +609,12 @@ def execute(self, command,
         # end handle
 
         try:
+            if sys.platform == 'win32':
+                CREATE_NO_WINDOW = 0x08000000
```
Byron: `CREATE_NO_WINDOW` is used twice, is there a reason it is not a constant on module or class level ?

git/cmd.py
```diff
@@ -609,6 +609,12 @@ def execute(self, command,
         # end handle
 
         try:
+            if sys.platform == 'win32':
```
Byron: Generally for these kinds of assignment, I believe it's easier to read by starting out like this:
```python
creationflags = None
if sys.platform == 'win32':
	creationflags = CREATE_NO_WINDOW
```

Byron:
Now this PR contains two distinct and independent changes. The most recent one is hard to test, and I'd be fine without one.
As for the first set of changes, I believe I already referred to the respective test-suite. Generally, you want to write a test that fails until your changes are applied. This is the whole trick. The suite in question should already contain everything you need to setup your own.

git/cmd.py
Byron: This concern is reflected in actual breakage as well: https://travis-ci.org/gitpython-developers/GitPython/jobs/148304657#L318

git/cmd.py
barry-scott: I intended to provide a obvious patch. You usually change my process control patches so I left that for you to decide on where you wished to define CREATE_NO_WINDOW.

barry-scott:
I fixed the keyword parameter passing.

barry-scott:
I have added a basic test by changing one of the existing test_repo.py tests.

You can see the test fail by undoing the first line fix and running the test.

Byron:
@barry-scott Thank you, looks very good now ! I am particularly excited about the test, and hope in future you can give *test-first* development a shot as well.

barry-scott:
To get the test as one patch and the the test + fix as a second you will have to fix the dependency of the tests on master. Use of master means that you do not have a reproducible test. Its inputs depend on what the user did to create there master.

I suspect that if you had to create the test data that you use master for you might have found this bugs ages ago.

PR #473, opened 2016-06-14
Fix for parsing non-ASCII chars in status lines

nvie:
It's a bit unclear which test fixture contained the data that tests this: changing the existing ones with `- [up to date] foo -> foo` lines did not turn any tests red.

I've updated this regex twice now very recently and I think just generically accepting any characters except whitespace is most future-compatible. We're running into some very wild ref names in practice, so this seems the most stable fix.

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -5,6 +5,8 @@ Changelog
 2.0.6 - Fixes and Features
 ==========================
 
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a bunch !

The question is to what extend the unit-tests can be trusted in that regard. Taking into consideration that the progress parsing has always been a weak spot, I'd be happy to give it another shot by merging this commit.

PR #474, opened 2016-06-14
Store raw path bytes in Diff instances

nvie:
Previously, the following fields on Diff instances were assumed to be passed in as unicode strings:

  - `a_path`
  - `b_path`
  - `rename_from`
  - `rename_to`

However, since Git natively records paths as bytes, these may potentially not have a valid unicode representation.

This patch changes the Diff instance to instead take the following equivalent fields that should be raw bytes instead:

  - `a_rawpath`
  - `b_rawpath`
  - `raw_rename_from`
  - `raw_rename_to`

**NOTE ON BACKWARD COMPATIBILITY:**  
The original `a_path`, `b_path`, etc. fields are still available as properties (rather than slots).  These properties now dynamically decode the raw bytes into a unicode string (performing the potentially
destructive operation of replacing invalid unicode chars by "�"'s).

This means that all code using Diffs should remain backward compatible. The only exception is when people would manually construct Diff instances by calling the constructor directly, in which case they should now pass in bytes rather than unicode strings.

See also the discussion on https://github.com/gitpython-developers/GitPython/pull/467

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -5,6 +5,9 @@ Changelog
 2.0.6 - Fixes and Features
 ==========================
 
+* API: Diffs now have `a_rawpath`, `b_rawpath`, `raw_rename_from`,
+  `raw_rename_to` properties, which are the raw-bytes equivalents of their
+  unicode path counterparts.
 * Fix: TypeError about passing keyword argument to string decode() on
   Python 2.6.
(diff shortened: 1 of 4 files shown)


nvie:
@Byron What do you think of this?

Byron:
I love it! 

Even though it would have the potential to break people who did manually create `Diff` instances, I believe in this case, the benefits of the many should outweigh the ones of the few.

Thanks for your contribution, again :) !

nvie:
Cheers, I've just released 2.0.6 to PyPI — thanks!

PR #479, opened 2016-06-20
Commit without executing hooks, fixes #468

### LINKED ISSUE #468: Commit without executing hooks
If I do `index.commit(msg)` it will execute [`pre-commit`](https://github.com/gitpython-developers/GitPython/blob/d5739cd466f77a60425bd2860895799f7c9359d9/git/index/base.py#L941) and [`post-commit`](https://github.com/gitpython-developers/GitPython/blob/d5739cd466f77a60425bd2860895799f7c9359d9/git/index/base.py#L946)

That makes it very hard to do the equivalent of `git commit -m "msg" --no-verify` (same as `git commit -m "msg" -n`).

peterbe:
This is a start to fix https://github.com/gitpython-developers/GitPython/issues/468

However, I have no idea where to add tests for this. 
I couldn't find any basic test that just tests the commit, with executing hooks. Tips most appreciated.

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -931,19 +931,24 @@ class IndexFile(LazyMixin, diff.Diffable, Serializable):
         return out
 
     def commit(self, message, parent_commits=None, head=True, author=None,
-               committer=None, author_date=None, commit_date=None):
+               committer=None, author_date=None, commit_date=None,
+               skip_hooks=False):
         """Commit the current default index file, creating a commit object.
         For more information on the arguments, see tree.commit.
 
         :note: If you have manually altered the .entries member of this instance,
                don't forget to write() your changes to disk beforehand.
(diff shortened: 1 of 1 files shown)


Byron:
With the current code-base, it would be a bit cumbersome to test, as you would probably have to check that an actual hook as not been called. Mocking injection seems difficult without a little refactoring.

So I guess merging it will just be fine in this case, thanks for your contribution !

peterbe:
Awesome! That's going to really help me. 
So, when will this be in a pypi release?

Byron:
A [new release](https://pypi.python.org/pypi?name=GitPython&version=2.0.7&:action=display) was just pushed to pypi.
Usually you shouldn't have to wait for a release though, as you should always be able to install from the git repository, specifying a tag, commit or branch.

PR #496, opened 2016-08-01
is_dirty supports path. Fixes #482.

### LINKED ISSUE #482: `is_dirty` for directory
I have a feature request: it will be useful to have something similar to `git status <path>` that returns which files are modified, untracked, etc. - in the give directory only.

Similar to `repo.is_dirty()`, but something that accepts path as well.

The usecase: we use git commitish to generate docker tags, and need to know whether a particular module directory is dirty.

haizaar:


Byron:
@haizaar Thanks a lot ! The PR already looks good to me. The only thing I was wondering is whether we should start supporting multiple paths right away. It's just that `git status` already does that, at least when looking at it's usage spec:
```man
git status [<options>...] [--] [<pathspec>...]
```

What do you think ?

haizaar:
Hi Sebastian,

I would rather wait until someone requests it. Particularly, because the
project is in maintenance mode, we better only add features that are driven
by user demand.

`path` argument can be extended in the future to accept list of paths as
well.

Byron:
Thanks for your contribution, it's much appreciated !

haizaar:
Thanks. When do you plan a new PyPi release?

Byron:
v2.0.8 was just released to pypi.

haizaar:
Thank you!

PR #498, opened 2016-08-17
Use $GIT_DIR when set

DoomHammer:
This allows GitPython and downstream projects (such as [PyGitUp](https://github.com/msiemens/PyGitUp)) to work with [yadm](https://github.com/TheLocehiliosan/yadm).


Byron:
Thanks a bunch !

PR #512, opened 2016-09-12
install ordereddict only on 2.6 with wheel

graingert:


Byron:
@graingert Thanks for your contribution! It's much appreciated, as I would like to get onto wheels too! Unfortunately the build fails, here is the [error-log](https://travis-ci.org/gitpython-developers/GitPython/jobs/159391257#L484). In case you get to it this weekend, it would be much appreciated, otherwise I'd try to mechanically fix the linter issues, even though I don't understand what the code actually does.

Byron:
Alright, here is the patch to pass the linter:
```patch
diff --git a/setup.py b/setup.py
index 15e4571..35b1115 100755
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,8 @@ except ImportError:
 
 from distutils.command.build_py import build_py as _build_py
 from setuptools.command.sdist import sdist as _sdist
+import pkg_resources
+import logging
 import os
 import sys
 from os import path
```

However, I realize that I don't understand the added special cases very much. Can you give me a little more context on the problem you ran into that this patch fixes ?

graingert:
This means that python 2.6 will be able to use wheels because they don't run the setup.py

graingert:
I've allowed edits from maintainers on this branch

Byron:
Alright, it's merged. Thanks for your contribution !
In an ideal world, this rather messy setup.py would not be needed, even though by now I learned to live with it.

PR #521, opened 2016-10-02
BF: remove a submodule with a remote without refs + misc fixes around

yarikoptic:
now if there is a remote without any refs (this happens e.g. with git-annex repo with special remotes), it would refuse to remove the submodule
(with 2.0.8 it also didn't work but with a different msg "AssertionError: Remote datalad did not have any references")

PR #522, opened 2016-10-02
RF: coveralls (not used/relied on really) -> codecov

yarikoptic:
codecov in our (datalad, etc) experience provides a better service,
great support, and super-nice intergration with chromium and firefox
for reviewing coverage of pull requests.  In light of the
@with_rw_directory fiasco detected/fixed in #521 I would strongly
recommend to (re-)enable and use coverage reports

edit 1: I think that if you accept this PR codecov might automagically enable itself on this repo, but I could be wrong and someone with authority might need to login into codecov via github and enable

coverage as seen from my clone: https://codecov.io/gh/yarikoptic/GitPython/commit/94ca83c6b6f49bb1244569030ce7989d4e01495c
and trust me integration with browsers is great!    ha -- actually basic (since there were no reports against master -- can't show the difference)  report already appeared here (magic!)

If you want to see it in action -- install codecov browser plugin and look e.g. at https://github.com/datalad/datalad/pull/969/files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ git:
 install:
   - git submodule update --init --recursive
   - git fetch --tags
-  - pip install coveralls flake8 ddt sphinx
+  - pip install codecov flake8 ddt sphinx
 
   # generate some reflog as git-python tests need it (in master)
   - ./init-tests-after-clone.sh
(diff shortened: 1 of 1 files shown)


ankostis:
That seems cool.  @Byron has to decide and implement the change.

yarikoptic:
Please let me know if I could be of additional help in persuasion :-)

Byron:
No persuasion required :) ! I just want to make sure we get what we need to keep GitPython up and running !

PR #528, opened 2016-10-11
Improved way of listing URLs in remote

guyzmo:
instead of using git remote show that can trigger a connection to the
remote repository, use git remote get-url --all that works by only
reading the .git/config.

change should have no functional impact, so no test needed.

Signed-off-by: Guyzmo <guyzmo+github@m0g.net>

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -78,7 +78,7 @@ install:
 build: false
 
 test_script:
-    - nosetests --with-coverage
+    - nosetests -v --with-coverage
 
 #on_success:
 #    - IF "%PYTHON_VERSION%"=="3.4" (coveralls)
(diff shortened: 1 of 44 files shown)


guyzmo:
(I noticed the issue when doing a `.urls` while on the train without internet connection 😖)

ankostis:
@guyzmo Linux checks fail due #513 BUT also due to `git.test.test_remote:TestRemote.test_multiple_urls ()` failing due to ([I suppose](http://stackoverflow.com/a/32991784/548792)) TravisCI having an older version of git; check my own build where I don't merge with the buggy master branch: https://travis-ci.org/ankostis/GitPython/jobs/166806778

guyzmo:
looks like it's merged, thanks 👍

PR #529, opened 2016-10-11
repo, cmd: DROP UNEEDED Win path for chcwd & check for '~' homedir

ankostis:
+ Do not abspath twice when constructing cloned repo.
+ Add `git.repo.base` logger.

According to the TCs, the changes in this PR are OK.
@Byron is there any reason for not deleting this Window-only code-path about changing-cwd?

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -62,7 +62,10 @@ from git.compat import (
 import os
 import sys
 import re
+import logging
 from collections import namedtuple
+
+log = logging.getLogger(__name__)
 
 DefaultDBType = GitCmdObjectDB
 if sys.version_info[:2] < (2, 5):     # python 2.4 compatiblity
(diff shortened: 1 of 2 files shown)


ankostis:
Has already been merged in 8ea7e265d15 with #530.

git/repo/base.py
```diff
@@ -871,46 +874,15 @@ def _clone(cls, git, url, path, odb_default_type, progress, **kwargs):
...
-                os.chdir(prev_cwd)
-                path = prev_path
-            # END reset previous working dir
-        # END bad windows handling
+        proc = git.clone(url, path, with_extended_output=True, as_process=True,
+                         v=True, **add_progress(kwargs, git, progress))
+        if progress:
+            handle_process_output(proc, None, progress.new_message_handler(), finalize_process)
+        else:
+            (stdout, stderr) = proc.communicate()  # FIXME: Will block of outputs are big!
```
Byron: It's quite amazing that `communicate()` actually cannot be trusted or is simply not implemented correctly :/.

Byron:
@ankostis This looks like a massive simplification, which in any case is a good thing. Given that Appveyor will alert us of breakage, I think it is fine to go ahead with it.

git/repo/base.py
ankostis: Well, you are quite right - `communicate()` CAN be trusted.

I did a small experiment of reading a Gbyte, both in PY27 and PY35, and they do not block: 

```python
>>> import subprocess as sb
>>> proc=sb.Popen('dd if=/dev/zero bs=1024 count=1000000', bufsize=0, stdin=sb.PIPE, stdout=sb.PIPE, stderr=sb.PIPE)
>>> a,b=proc.communicate()
```

I don't remember when I got this mistrust for `communicate()`, maybe when I had tried PY33, or maybe when epxerimenting with interactive processes (like the `git.cmd.Git.cat-file` persistent command); there the `comminicate()` is not suitable because it will start consuming streams till the death of the process, but the process is awaiting for more input from python-side before ending, so it's a deadlock.

One or two changes in my *windows* fixes were under this wrong assumption that `communicate()` should not be trusted.  But mostly I replaced code accessing directly `proc.stderr/stdout` from the main thread - this is definitely freeze-prone code.

Byron: A nice experiment ! It doesn't seem to output both on stdout and stderr though, which might be causing the problem we see. The [implementation of communicate](https://github.com/python/cpython/blob/c30098c8c6014f3340a369a31df9c74bdbacc269/Lib/asyncio/subprocess.py#L181) looks asynchronous, but who knows what's hidden in the details.

ankostis: Even when `stdout/stderr` is interplexed, `communicate()` does not block (at least in PY3):
 
```python
>>> import subprocess as sb
>>> proc = sb.Popen(['python','-c',"import sys\nfor i in range(1000000):\n    print('a'*1024); print('b'*1024, file=sys.stderr)"], stdin=sb.PIPE, stdout=sb.PIPE, stderr=sb.PIPE)
>>> %time a,b=proc.communicate()
Wall time: 48.3 s
>>> len(a), len(b)
(1026000000, 1026000000)
```

For some "unicode" reason cannot run the above experiment in PY2.

PR #536, opened 2016-10-18
Allow for setting git options, that are persistent across subcommand calls

bpoldrack:
Currently `_git_options` in class `Git` are reset after each subcommand call. While this is appropriate in most cases, I'm in need to have options for the git executable, that are used with all calls to git on a certain repository and I can't provide this option for each and every call. Therefore I'd like to introduce the possibility to persistently set such options for a given instance of `Git`.

This is my first PR here, so I'm not exactly sure, what you want me to provide. `CONTRIBUTING.md` reads: "Write a test that fails unless your patch is present." 
Where am I supposed to put such a test?
Just add it to `git/test/test_git.py`?

### DIFF
--- a/git/test/test_git.py
+++ b/git/test/test_git.py
@@ -172,7 +172,7 @@ class TestGit(TestBase):
 
         # reset to empty:
         self.git.set_persistent_git_options()
-        self.assertRaises(GitCommandError,  self.git.NoOp)
+        self.assertRaises(GitCommandError, self.git.NoOp)
 
     def test_single_char_git_options_are_passed_to_git(self):
         input_value = 'TestValue'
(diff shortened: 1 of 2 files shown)


git/test/test_git.py
```diff
@@ -160,6 +160,20 @@ def test_options_are_passed_to_git(self):
...
+        self.git.set_persistent_git_options(version=True)
+        git_version = self.git.NoOp()
+        self.assertEquals(git_version, git_command_version)
+        # subsequent calls keep this option:
+        git_version_2 = self.git.NoOp()
+        self.assertEquals(git_version_2, git_command_version)
+
+        # reset to empty:
+        self.git.set_persistent_git_options()
+        self.assertRaises(GitCommandError,  self.git.NoOp)
```
bpoldrack: Right, forgot this one. Thanks.

Byron:
@bpoldrack Thanks a lot for your contribution ! I am doing maintenance today, and am confident to manage to put out a new release as well.

PR #537, opened 2016-10-19
FIX #535: expand also GIT_DIR var on Repo-construction

### LINKED ISSUE #535: Use of GIT_REPO env-var for setting Repo path in 2.0.9
[This change](https://github.com/gitpython-developers/GitPython/commit/26bb778b34b93537cfbfd5c556d3810f2cf3f76e) in `2.0.9` is a bit problematic:

- Even if a `path` is explicitly provided to the constructor, the `GIT_DIR` environment variable is used instead

- For things like git hooks, `GIT_DIR` is set to `.git` by default. This will result in [`self.git_dir`](https://github.com/gitpython-developers/GitPython/blob/master/git/repo/base.py#L156) being set to `.git` (`os.path.abspath` instead of `os.path.normpath` would fix this). Consequently, subsequent commands fails (eg: `repo.git.status()`)

ankostis:
+ Ignore "empty" GIT_DIR vars.
+ Improve documentation on the constructor `path` parameter.
Minor repo-code and doc correcions.
  + Expansion of paths also `osp.normalize()` them.
  + Make Repo-fields --> class-fields to avoid initializations on
construct.
  + Explain and rename `git.repo.fun.find_git_dir()` is for submodules
(`find_submodule_git_dir()`).

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -100,8 +100,8 @@ class Repo(object):
                 repo = Repo("$REPOSITORIES/Development/git-python.git")
 
             - In *Cygwin*, path may be a `'cygdrive/...'` prefixed path.
-            - If `None, current-directory is used.
-            - The :envvar:`GIT_DIR` if set and not empty takes precendance over this parameter.
+            - If it evaluates to false, :envvar:`GIT_DIR` is used, and if this also evals to false,
+              the current-directory is used.
         :param odbt:
             Object DataBase type - a type which is constructed by providing
             the directory containing the database objects, i.e. .git/objects. It will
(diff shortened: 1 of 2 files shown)


git/repo/base.py
```diff
@@ -124,6 +124,8 @@ def __init__(self, path=None, odbt=DefaultDBType, search_parent_directories=Fals
                 repo = Repo("~/Development/git-python.git")
                 repo = Repo("$REPOSITORIES/Development/git-python.git")
 
+            if `None, current-directory is used.
+            The :envvar:`GIT_DIR` if set and not empty takes precendance over this parameter.
```
ethereon: The environment variable taking precedence over the specified path seems error prone and undesirable. For instance, with `GIT_DIR` set, you can no longer do something like this:

```
repo_1 = git.Repo('/path/to/repo1')
repo_2 = git.Repo('/path/to/repo2')
```
I think `GIT_DIR` as a fallback makes sense (if `path` is `None` for instance). However, is there any reason to prefer it if an explicit path is provided?

ankostis: Discussion moved to original issue #535

Byron:
Thanks a lot @ankostis :) !

PR #544, opened 2016-10-23
Stop importing on runtime dependent projects as submodules

ankostis:
As discussed in the last comments of #511, the projects stops relying on git's _submodules_ to import its dependencies.  That means that for TravisCI to succeed, the correct revisions must exist in PyPi, and hence, more frequent releases.

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -15,24 +15,6 @@ import os.path as osp
 __version__ = 'git'
 
 
-#{ Initialization
-def _init_externals():
-    """Initialize external projects by putting them into the path"""
-    if __version__ == 'git':
-        sys.path.insert(0, osp.join(osp.dirname(__file__), 'ext', 'gitdb'))
-
-    try:
-        import gitdb
-    except ImportError:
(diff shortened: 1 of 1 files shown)


ankostis:
Indeed, CI-testing becomes impossible, unless we pip install `master` from git each time; but for the releases, this will not be correct, unless all sub-projects have correct masters.

Byron:
@ankostis This PR has been stale for a while now - what do you suggest we should do with it? Giving it more time is absolutely an option, and merging it certainly is too. Even though I am the biggest fan of this as GitPython will just work out of the box in the checkout location.

ankostis:
I would suggest not to merge it.
I have stopped efforts on dependent projects (my rewrite of mmap was 98% ready but have no more interest to finish it  since it is not actually used by GitPython). And GitDB is probably used in few "write-only" cases, so choose the solution with least effort required.

Byron:
@ankostis Thanks for letting me know. As per your recommendation, I am closing it.

Issue #553, opened 2016-12-02
GitPython `repo.index.commit()` spawns persistent git.exe instance, holds handles to repo


yellowtomato8675:
I am trying to use GitPython for some repo manipulation, but ran into issues with my app, with handles open where i wouldn't expect.

Bug-jarring the issue, it seems that calling `repo.index.commit()` results in a several (4, consistently) git.exe processes being spawned, each holding a handle to the repo's root directory `newRepo`.  When the test below goes to delete this TempDir, the processes are still up, causing a failure on context-manager `__exit__()`.  My app occasionally needs to do a similar cleanup, so hits the same issue.

On the one hand, it looks like there is no contect-manager capable repo-wrapper, meaning it makes sense if _some_ resources is open, it will not typically be cleaned/GC'd before the tempdir `__exit__()`.  On the other hand - if `Repo` is going to behave like that, it really should not persist resources that have such side-effects.

Here is a working unittest:

    import unittest
    import git
    import tempfile
    import os.path
    
    class Test(unittest.TestCase):
    
        def testCreateRepo(self):
            with tempfile.TemporaryDirectory(prefix=(__loader__.name) + "_") as mydir:
    
                # MAKE NEW REPO 
                repo = git.Repo.init(path=os.path.join(mydir, "newRepo"), mkdir=True)
                self.assertTrue(os.path.isdir(os.path.join(repo.working_dir, ".git")), "Failed to make new repo?")
                
                # MAKE FILE, COMMIT REPO
                testFileName = "testFile.txt"
                open(os.path.join(repo.working_dir, testFileName) , "w").close()
                repo.index.add([testFileName])
                self.assertTrue(repo.is_dirty())
                
                #### 
                # COMMENTING THIS OUT --> TEST PASSES
                repo.index.commit("added initial test file") 
                self.assertFalse(repo.is_dirty())
                #### 
                
                # adding this does not affect the handle
                git.cmd.Git.clear_cache()
                
                
                print("done") # exception thrown right after this, on __exit__
                # i can also os.walk(topdown=False) and delete all files and dirs (including .git/)
                # it is just the  newRepo/  folder itself who's handle is held open
                
    if __name__ == '__main__':
        unittest.main()

PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\%USER%\\AppData\\Local\\Temp\\EXAMPLE_gitpython_v3kbrly_\\newRepo'

[...]

PR #541, opened 2016-10-21
Fixes to support Python 2.6 again.

andy-maier:
GitPython 2.0.9 has some issues on Python 2.6 (see issue #540).

This PR fixes them. For details, see the commit message.
The Travis runs now pass on all Python versions including 2.6.

Please review.

PR #555, opened 2016-12-08
Retrofit `repo` class as context-man to cleanup global mman on repo-delete

ankostis:
Some changes for this release:
See #553 for difficulties when working on Windows, where open file-handles actually "lock" the underlying files, and repos cannot be deleted.

More changes:

- Small pythonism on  #541, improves  docstring of `cmd._call_process()`.
- Export `gitpython.util.rmtree()` for deleting repos in *Windows* where blobs are readonly.

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -49,7 +49,8 @@ from git.util import (                  # @NoMove @IgnorePep8
     LockFile,
     BlockingLockFile,
     Stats,
-    Actor
+    Actor,
+    rmtree,
 )
 
 #} END imports
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -822,27 +822,30 @@ class Git(LazyMixin):
             is realized as non-existent
 
         :param kwargs:
-            is a dict of keyword arguments.
-            This function accepts the same optional keyword arguments
-            as execute().
-
-        ``Examples``::
+            It contains key-values for the following:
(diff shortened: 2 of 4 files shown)


Byron:
@ankostis Do you think I should wait till tomorrow until I cut the release to be able to include this one? Or would you prefer to just have this released separately at a later time?

Byron:
@nvie I just created a new release.
@ankostis I will follow this PR and make a release when needed. Thanks to the `Makefile`, that became easy and convenient (interestingly, I wasn't aware it existed anymore :)).

ankostis:
Sorry @Byron , I've just seen your message. 

I'm not exactly sure what happened and you could not include my PR in 2.1.1. 
I believe it ready to be merged.
is this  right, or this PR relies on features from a different GitDB/smmap?

Byron:
Even though the implementation seems to be in place, it would certainly be nice to include an example in the `test_docs.py` to show people how to use it. I for one wouldn't know myself actually, just because my experience with context managers is very limited.
Do you think you could add an example? I could then include it in the tutorial and write some prose around it :).

ankostis:
Apologies for not having the time,
but I really believe this little PR wroth to have it.

Too overloaded to fix the TCs in the next weeks :-(

Byron:
I think so too :)! Sorry for the incredible delays.

PR #565, opened 2016-12-28
Delete requirements.txt

ignatenkobrain:


Byron:
I am not sure what to think of this PR if you don't provide any motivation.
As all tests fail as well, so apparently more work is needed to get a merge.

Byron:
I am closing this PR as there was no reply in 3 months. If @yarikoptic sees value though, please let me know and we figure something out.

yarikoptic:
that is ok with me -- no immediate need/desires ;)

PR #572, opened 2017-01-10
Log lines for debugging

expobrain:


Byron:
@expobrain Is this PR linked to some issue I am unaware of? If not, could you explain its motivation? Thank you.

expobrain:
Hi @Byron, this PR allows to see not only the git-push/git-fetch command but also the output of it.

It's useful i.e. in case a git hook rejects the commit and you want to see the message in the logs.

expobrain:
Sorry, closed the PR by mistake

git/util.py
```diff
@@ -52,7 +52,18 @@
            'RemoteProgress', 'CallableRemoteProgress', 'rmtree', 'unbare_repo',
            'HIDE_WINDOWS_KNOWN_ERRORS')
 
+
+try:  # Python 2.7+
+    import logging.NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
```
Byron: I don't understand what the NullHandler is used for. So far it was sufficient to obtain a logging using `logging.getLogger('git.util')` and assume parties interested in logging have set it up to contain/inherit handlers.

Byron:
@expobrain Sorry for being so very delayed in my replies. I left a comment, but once clarified, it should be fine to merge.

git/util.py
expobrain: I juts following the same pattern of enable logging like it is done in other part of `GitPython` as well (see `git/cmd.py` for example). The only difference is that unit tests under Python 2.6+ are failing because the `logging` package in 2.6 doesn't have the `NullHandler` thus this little piece of code.

Byron:
Sorry for the great delay, and for the hold-up due to this Nullhandler. Now that it is clarified I realize this PR should not be merged, as it adds `info`-level logging to the implementation of the parser.
The way clients are expected to work with that, should they need logging of the received line, is to pass an implementation of the `RemoteProgress` type. There one could override `_parse_progress_line` and log it, for example.

PR #574, opened 2017-01-15
Doc fix: Using camellcase package name in "pip install" command

vokimon:
The pip install command in documentation refers the package in lower case "gitpython" instead of the actual package name in camellcase "GitPython"


Byron:
Thanks a lot!

PR #578, opened 2017-02-02
git expects boolean value to be lower case

barry-scott:
str(bool) returns capitialised True and False
that is not what git expects or compatible with
getboolean.


barry-scott:
This is a bad PR. I'll put a god one together.

PR #582, opened 2017-02-10
fix(cmd): checking process.DEVNUL were needlessly opening `os.devnull`

ankostis:
Also this fixes resource-leak warning on Windows Python-3.5.3+:

```
D:\python-3.5.2.amd64\lib\site-packages\git\cmd.py:583: ResourceWarning:
unclosed file <_io.BufferedWriter name='nul'>
  else getattr(subprocess, 'DEVNULL', open(os.devnull, 'wb')))
```

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -580,7 +580,7 @@ class Git(LazyMixin):
 
         stdout_sink = (PIPE
                        if with_stdout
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for this one!
I do hope that the subprocess consumes the file-handle too. Last time I checked, it looked like it did, so merely opening the file should not cause leakage.

PR #588, opened 2017-02-24
check for GIT_WORK_TREE

tbhartman:
This uses GIT_WORK_TREE environment variable if present.

In response to gitpython-developers/GitPython/issues/587.

Not all tests are passing on my master, but no new tests are failing with my commit.

I have added a test.


Byron:
Thanks a lot for your contribution, and especially for providing a test as well!

PR #602, opened 2017-03-06
Ignore all lines of subsequent hunks until last one is found

ghickman:
Git version 2.11.1+ introduced extra lines into the subsequent hunk
sections for incremental blame output.  The documentation notes that
parsers of this output should ignore all lines between the start and end
for robust parsing.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -713,11 +713,14 @@ class Repo(object):
                            committed_date=int(props[b'committer-time']))
                 commits[hexsha] = c
             else:
-                # Discard the next line (it's a filename end tag)
-                line = next(stream)
-                tag, value = line.split(b' ', 1)
-                assert tag == b'filename', 'Unexpected git blame output'
-                orig_filename = value
+                # Discard all lines until we find "filename" which is
+                # guaranteed to be the last line
+                while True:
+                    line = next(stream)
+                    tag, value = line.split(b' ', 1)
+                    if tag == b'filename':
+                        orig_filename = value
+                        break
 
             yield BlameEntry(commits[hexsha],
                              range(lineno, lineno + num_lines),
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -716,7 +716,7 @@ class Repo(object):
                 # Discard all lines until we find "filename" which is
                 # guaranteed to be the last line
                 while True:
-                    line = next(stream)
+                    line = next(stream)  # will fail if we reach the EOF unexpectedly
                     tag, value = line.split(b' ', 1)
                     if tag == b'filename':
                         orig_filename = value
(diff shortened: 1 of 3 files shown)


git/repo/base.py
```diff
@@ -713,11 +713,14 @@ def blame_incremental(self, rev, file, **kwargs):
...
                 commits[hexsha] = c
             else:
-                # Discard the next line (it's a filename end tag)
-                line = next(stream)
-                tag, value = line.split(b' ', 1)
-                assert tag == b'filename', 'Unexpected git blame output'
-                orig_filename = value
+                # Discard all lines until we find "filename" which is
+                # guaranteed to be the last line
+                while True:
```
Byron: I believe this loops break-condition is not clearly defined just now.
Do you think 'end-of-stream' could be another one?

What do you think of something like this:
```python
for line in stream:
	tag, value = line.split(b' ', 1)
	if tag == b'filename':
		orig_filename = value
		break
		
# Finally handle the case of the stream being exhausted too early
```

Byron:
Thanks a lot! In addition to the comment, how feasible/easy is it for you to add a test based on the kind of input you are handling? I believe there are fixtures for this already, maybe it's possible to adjust one too.

git/repo/base.py
ghickman: I've added a comment in 77b20be to be more explicit about an unexpected EOF condition.  With a for loop we'd need to raise that exception ourselves in an `else` block so I think the while loop works better here?

ghickman:
@Byron – yep I think that sort of test should be easy to add with an extra fixture that has the new style git output.  I'll add that as soon as I get the time :)

git/repo/base.py
Byron: Ah, I see now! I didn't know that `next` throws - iterators do that, right. Have been doing Rust for too long ;).

Byron:
@ghickman Great, I will be waiting for that one then - by now github sends emails on new commits, so I should see it and merge shortly thereafter.

git/test/fixtures/blame_incremental_2.11.1_plus
```diff
@@ -0,0 +1,33 @@
+82b8902e033430000481eb355733cd7065342037 2 2 1
```
Byron: Thanks a lot! I am ready to merge!
One last question though: should there or should there not be empty lines in the fixture?

ghickman: I don't think there should be empty lines in the fixture.

If it helps I generated the fixture with `git blame --incremental -p 82b8902e033430000481eb355733cd7065342037 -- AUTHORS > git/test/fixtures/blame_incremental_2.11.1_plus` (which I believe matches how `blame_incremental` was generated).

Byron:
Awesome! Thanks a lot for this one!

PR #607, opened 2017-03-09
Python 3.6 invalid escape sequence deprecation fixes

scop:
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -208,7 +208,7 @@ class FetchInfo(object):
     NEW_TAG, NEW_HEAD, HEAD_UPTODATE, TAG_UPDATE, REJECTED, FORCED_UPDATE, \
         FAST_FORWARD, ERROR = [1 << x for x in range(8)]
 
-    re_fetch_result = re.compile('^\s*(.) (\[?[\w\s\.$@]+\]?)\s+(.+) -> ([^\s]+)(    \(.*\)?$)?')
+    re_fetch_result = re.compile(r'^\s*(.) (\[?[\w\s\.$@]+\]?)\s+(.+) -> ([^\s]+)(    \(.*\)?$)?')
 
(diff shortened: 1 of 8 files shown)


scop:
Done.

git/remote.py
```diff
@@ -391,7 +391,7 @@ def __init__(self, repo, name):
 
     def __getattr__(self, attr):
         """Allows to call this instance like
-        remote.special( \*args, \*\*kwargs) to call git-remote special self.name"""
+        remote.special( \\*args, \\*\\*kwargs) to call git-remote special self.name"""
```
isaacsanders: Why not use a raw string here? (I am fairly new to Python, so this is a question seeking a real answer if there is an answer, not a passive aggressive question)

SethMichaelLarson: Typically you don't see docstrings being raw strings.

isaacsanders: Does it negatively effect their use as docstrings?

isaacsanders: It seems like you might want to have the docstring be WYSIWYG, and the raw string appears to be meant for that, to me...

SethMichaelLarson: I don't have any say in this repo so I don't mind either way. Just a preference and for homogeneity.

scop: BTW, FWIW and all that, I don't know why the asterisks in this particular docstring are backslash escaped in the first place.

Byron:
@scop I am very sorry for the delay, and thanks a lot for making this PR! I had a look at the link you provided, but was unable to understand the relation to this PR. Could you briefly explain the problem it is solving? My intuition is that the docstrings look incorrect when rendered.

Byron:
Great, thanks a lot!

PR #608, opened 2017-03-09
Add most recent Python versions in Travis CI

SylvainDe:
Add more recent Python versions including development branches and nightly build.

The motivation came from reading brettcannon's article : https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/ . Trying to activate the newest Python versions on CI jobs is in most case a win-win situation: if everything works fine, there is nothing to worry about. If an issue is spotted, it is good to know about it to fix it on your side or to open a bug on CPython ( https://bugs.python.org/ ).

Also, if a failures is spotted, you can use the allow_failures option in your matrix build (more information about this in the link above).

Also, nightly and 3.7-dev may be a bit too recent/unstable so I can remove then if need be.

More information about how this PR happened : https://github.com/SylvainDe/CIthon .

More information about the various Python versions available on Travis : https://docs.travis-ci.com/user/languages/python/ .

More information about the Python 3.6 version schedule : https://www.python.org/dev/peps/pep-0494/ .
More information about the Python 3.7 version schedule : https://www.python.org/dev/peps/pep-0537/ .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
+  - "3.6-dev"
+  - "3.7-dev"
+  - "nightly"
   # - "pypy" - won't work as smmap doesn't work (see gitdb/.travis.yml for details)
 #matrix:
 #  allow_failures:


Byron:
The only thing I added is to allow failures - maybe that's the wrong move as I would be supposed to be warned if something breaks in development versions. On the other hand, for now I am fine having the knowledge, without being forced to act on it.
And of course, I hope nothing will break :) , while acknowledging the usefulness of an early warning system.

SylvainDe:
Sure, I'm glad you managed ​to get the most out of it. Allowing failures is
a great way to keep some of the benefits without making it too much of a
burden.

Le 9 avr. 2017 1:26 PM, "Sebastian Thiel" <notifications@github.com> a
écrit :

The only thing I added is to allow failures - maybe that's the wrong move
as I would be supposed to be warned if something breaks in development
versions. On the other hand, for now I am fine having the knowledge,
without being forced to act on it.
And of course, I hope nothing will break :) , while acknowledging the
usefulness of an early warning system.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/gitpython-developers/GitPython/pull/608#issuecomment-292780008>,
or mute the thread

Issue #622, opened 2017-05-02
Bad performance when obtaining ref sha while a big amount of Git Reference


wusisu:
In my production case, I have more than 144k References in the Git directory.
(and `packed-refs` file has over 12 MiB)

I need to iterate all branches and tags and get their hexsha.
However, It takes more than 2500 seconds in iteration.
```
for ref in repo.refs:
    yield ref.commit.hexsha
```

I think this takes too much time, and I wonder If I need to implement my own
`Reference.list_item` method to speed it up?

best wishes.

PR #624, opened 2017-05-04
reference: iter_item method add a parameter to return the ref's destination

wusisu:
This solve my problem #622 .

I have no idea and need advice about if this PR should be merged into GitPython.

### DIFF
--- a/git/refs/reference.py
+++ b/git/refs/reference.py
@@ -94,10 +94,10 @@ class Reference(SymbolicReference, LazyMixin, Iterable):
         return '/'.join(tokens[2:])
 
     @classmethod
-    def iter_items(cls, repo, common_path=None):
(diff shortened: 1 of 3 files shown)


git/refs/reference.py
```diff
@@ -94,10 +94,10 @@ def name(self):
         return '/'.join(tokens[2:])
 
     @classmethod
-    def iter_items(cls, repo, common_path=None):
+    def iter_items(cls, repo, common_path=None, with_destination=False):
         """Equivalent to SymbolicReference.iter_items, but will return non-detached
         references as well."""
```
Byron: Could you please describe what `with_destination` does in the doc-string? Since it is not available in the `SymbolicReference`, it's probably correct to put it here and make a note about this specialty in `SymbolicReference.iter_items()`.

Issue #627, opened 2017-05-13
Configparser does not respect order of config files with includes


DanielSiepmann:
When you have an include statement in the user `.gitconfig` there content will take precedence over the local `.git/config` of the repository.

E.g. the following code:
`entry = repo.config_reader().get_value('gitlab', 'url')`

With the following config setup:

`~/.gitconfig`:
```ini
[include]
    path = ~/.gitconfig-local
[gitlab]
    url = url1
```

`~/.gitconfig-local`:
```ini
[gitlab]
    url = url2
```

`.git/config`:
```ini
[gitlab]
    url = url3
```

Will return `url2` I would expect to receive `url3` instead as local config should take precedence as the include was in a more global config.

Looks like the issue resides inside the `read ` function e.g. https://github.com/gitpython-developers/GitPython/blob/c823d482d03caa8238b48714af4dec6d9e476520/git/config.py#L427 will just add the includes to the list, which is just wrong, as they should be read instantly to overwrite just the already known configs.

As the function is just to big, I'm not able to provide a fix.

Versions in use:
```
>>> import git; git.__version__
'2.1.3'
```

```
python --version
Python 2.7.13
```

Byron:
@DanielSiepmann Thanks for the detailed description, and for the fix! Please have a look at my review - I hope to be able to merge the fix soon and create a new release accordingly.

PR #628, opened 2017-05-14
BUGFIX: 627 fix include order

DanielSiepmann:
* Fix order of parsing configured include files.
* More specific files should overrule included files.
* Include files at same level where include directive happens.

Resolves: #627

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -410,6 +410,7 @@ class GitConfigParser(with_metaclass(MetaParserBuilder, cp.RawConfigParser, obje
             # Read includes and append those that we didn't handle yet
             # We expect all paths to be normalized and absolute (and will assure that is the case)
             if self._has_includes():
+                include_paths = []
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
There might be a pythonic way of solving this, but as I'm a beginner im happy to provide at least a fix for the issue.

git/config.py
```diff
@@ -424,8 +425,11 @@ def read(self):
                     if include_path in seen or not os.access(include_path, os.R_OK):
                         continue
                     seen.add(include_path)
-                    files_to_read.append(include_path)
+                    include_paths.append(include_path)
                     num_read_include_files += 1
+                include_paths.reverse()
```
Byron: I think this block should be rewritten like this:
```python
files_to_read = include_paths + files_to_read
```
This makes clearer that included files should be read and merged first.

PR #638, opened 2017-06-26
Repo: handle worktrees better

vathpela:
This makes Repo("foo") work when foo/.git is a file of the form created
by "git worktree add", i.e. it's a text file that says:

gitdir: /home/me/project/.git/worktrees/bar

and where /home/me/project/.git/ is the nominal gitdir, but
/home/me/project/.git/worktrees/bar has this worktree's HEAD etc and a
"gitdir" file that contains the path of foo/.git .

Signed-off-by: Peter Jones <pjones@redhat.com>


vathpela:
I'm really not sure what's going on in the AppVeyor test cases here.  It's not obvious to me that these are related to this change set, and locally on a Linux machine all tests with tox pass except all py34 ones, in which pip fails to install the test prerequisites correctly.  That's clearly just py34 not working, rather than being this patch.

I have no idea why test_git_submodules_and_add_sm_with_new_commit() and test_diff_with_staged_file() failures would be caused by this, and they do seem to be present in AppVeyor runs on the PRs before this one.  Staring at the code and the backtrace, my best guess is that polish_url() should be returning a backslash-version instead of slashes, and so the filesystem paths aren't working?  But I don't know how that all works in cygwin/windows, and I don't have a windows machine to reproduce locally.

Byron:
Hi @vathpela , thanks a lot for your contribution, and for making it such a gorgeous one!

I can feel the pain that appveyor can cause, which is why I chose to go down the easy path and try to skip the test there, just to save time to be spent more productively :). Also it looks like Appveyor was breaking for a while, as you noted, which shouldn't have happened in the first place.

Once the CI's are happy, this one will be merged.
Thanks again

PR #639, opened 2017-07-04
WIP: use `cygpath.exe` to convert Windows paths and respect mount-points

ankostis:
Currently the `/cygdrive/` mount-point is hard-coded in GitPython's conversions `git/utils.py` since version 2.1.3 that Cygwin support was added.  This PR purports to use Cygwin's `cygpath -w | -u` utility for path conversions.  
An added benefit is support out-of-the-box for *MSYS2* which is a *Cygwin* clone based on the excellent *pacman* package manager that (at last) mounts Windows-drives under root(`/`).

It is still a WorkInProgress (WIP) because although [all *travis* TCs pass](https://travis-ci.org/ankostis/GitPython/builds/250043838) on the POSIX part, *appveyor* now fails on all non-Cygwin builds; prior to this PR, it was only Cygwin builds that were failing in a couple of TCs: https://ci.appveyor.com/project/ankostis/gitpython/build/1.0.443).

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -6,6 +6,10 @@ Changelog
 ====================================
 
 * support for worktrees
+* fix(cygwin): use ``cygpath.exe`` to convert *Windows* paths and respect
+  different mount-points (e.g. *MSYS2* which is a *Cygwin* clone mounts drives
+  under root).
+
 
 2.1.3 - Bugfixes
 ====================================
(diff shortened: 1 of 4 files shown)


Byron:
@ankostis I happen to be on a merge-spree right now and wonder if this one is ready. Also I know you could just merge yourself when ready, I am just wondering :). Thank you.

ankostis:
Sorry, I cannot respond confidently.
I have imported this as hotfix in [my code](/JRCSTU/allinone/commit/96c1395b273d97a4147da4fc6bcf10bee8264f42) without any side-effect working on MSYS2 & Cygwin (less confident) for the last 2-3 months.

But still not all TCs pass, and these need fixing, but have no time to do it.
I would wait for them to be fixed; we already have [2 "worktree" TCs failing in Cygwin](https://ci.appveyor.com/project/Byron/gitpython/build/1.0.230/job/5qbeihnce9avar9s), so better not make it any worse than that.

Apologies for me not being available.

PR #641, opened 2017-07-13
open func  support "utf-8"

nooperpudd:
Support some Chinese name branches like `origin/feature/测试`.


Byron:
@nooperpudd My apologies for the late reply! Unfortunately I merged plenty of PRs in the wrong order, which is probably the reason yours is conflicting right now. As I would really like to see it merged, maybe you find the time to have another look.
I promise to be fast merging it once the tests on travis pass again.
Thank you so much for your consideration!

nooperpudd:
@Byron  ok, I will fix the problem.

Byron:
@nooperpudd As always, sorry for getting back to this so late. I have had a look at travis, and besides possibly unrelated failure, one is clearly attributed to `codecs` working slightly differently in Python > 2.7: https://travis-ci.org/gitpython-developers/GitPython/jobs/293061325#L2480.
Could you have a look again? Thanks a lot.

PR #654, opened 2017-08-04
worktrees: make non-packed refs also work correctly.

vathpela:
Turns out aec58a9 did the right thing for /packed/ refs, but didn't work
correctly on /unpacked/ refs.  So this patch gives unpacked refs the
same treatment.

Without the fix here, the test added will cause this traceback:

ERROR: Check that we find .git as a worktree file and find the worktree
  File "/home/pjones/devel/github.com/GitPython/git/test/lib/helper.py", line 92, in wrapper
    return func(self, path)
  File "/home/pjones/devel/github.com/GitPython/git/test/test_repo.py", line 938, in test_git_work_tree_dotgit
    self.assertIsInstance(repo.heads['aaaaaaaa'], Head)
  File "/home/pjones/devel/github.com/GitPython/git/util.py", line 893, in __getitem__
    raise IndexError("No item found with id %r" % (self._prefix + index))
IndexError: No item found with id 'aaaaaaaa'

Woops.

Signed-off-by: Peter Jones <pjones@redhat.com>

### DIFF
--- a/git/refs/remote.py
+++ b/git/refs/remote.py
@@ -37,6 +37,10 @@ class RemoteReference(Head):
         # and delete remainders manually
         for ref in refs:
             try:
+                os.remove(osp.join(repo.common_dir, ref.path))
+            except OSError:
+                pass
+            try:
                 os.remove(osp.join(repo.git_dir, ref.path))
             except OSError:
                 pass
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -26,6 +26,14 @@ from .log import RefLog
 __all__ = ["SymbolicReference"]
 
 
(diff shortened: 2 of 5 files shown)


vathpela:
Don't apply this one just yet; I've discovered some more things I don't like, and it's better to fix them with a more systemic approach.

vathpela:
Literally no idea why this is failing on AppVeyor.  I'll try to have a look in the next couple of days, but it's pretty busy over here right now.

Byron:
Thanks a ton for your contribution, and my apologies for getting to this PR that late! Windows is constantly failing by now, so I choose to ignore it :/.

PR #662, opened 2017-08-21
Fix leaking environment variables

Plazmaz:
When cloning a repo, GitPython will leak environment variables in error messages. For instance, this code:
```
import git
repo = git.Repo('https://www.github.com/Plazmaz/${PATH}')
repo.clone('testrepo/$PATH')
```
will output something like:
```
  File "test.py", line 2, in <module>
    repo = repo.Repo('https://www.github.com/Plazmaz/${PATH}', unsafe=True)
  File "GitPython/git/repo/base.py", line 133, in __init__
    raise NoSuchPathError(epath)
git.exc.NoSuchPathError: <THE CONTENTS OF $PATH>
```
This behavior has unwanted security implications. To counter this, I've added an `unsafe` variable, which will allow for environment variables to be expanded, otherwise, this behavior is disabled. By default, this variable is set to True. However, when used with environment variables, a warning is displayed. Hopefully, this will eventually be set to False by default. When running the same code, but with unsafe set to False, here's the output:
```
  File "test.py", line 2, in <module>
    repo = repo.Repo('https://www.github.com/Plazmaz/${PATH}', unsafe=False)
  File "GitPython/git/repo/base.py", line 133, in __init__
    raise NoSuchPathError(epath)
git.exc.NoSuchPathError: 
Documents/https:/www.github.com/Plazmaz/${PATH}
```


git/repo/base.py
```diff
@@ -50,8 +51,11 @@
 __all__ = ('Repo',)
 
 
-def _expand_path(p):
-    return osp.normpath(osp.abspath(osp.expandvars(osp.expanduser(p))))
+def _expand_path(p, unsafe=True):
```
ankostis: Better call it what it is, `expand_vars`, and explain the reason for using this flag in the invoking site.

And I would split it, not to have to compare lines to discover the difference:
```python
p = osp.expanduser(p)
if expand_vars:
    p = osp.expandvars(p)
return osp.normpath(osp.abspath(p))
```

git/repo/base.py
```diff
@@ -90,7 +94,7 @@ class Repo(object):
     # Subclasses may easily bring in their own custom types by placing a constructor or type here
     GitCommandWrapperType = Git
 
-    def __init__(self, path=None, odbt=DefaultDBType, search_parent_directories=False):
+    def __init__(self, path=None, odbt=DefaultDBType, search_parent_directories=False, unsafe=True):
```
ankostis: Wouldn't this default break existing clients expecting variable-expansion?
I guess *hooks* would suffer.

Plazmaz: The default is True, meaning variables will be expanded by default. However, this behavior is _bad_, especially seeing as a standard practice is storing secrets and keys in environment variables. The reason for the flag is to give a bit of a grace period, and allow users to transition.

Plazmaz:
Regarding compatibility, this is OFF by default for now. It will work with existing projects just fine.

Byron:
Without a new major release, this flag can't be OFF by default, and in maintenance mode, GitPython probably is not going to do that.

PR #664, opened 2017-08-25
util: move expand_path from repo/base and use it in Git class init

Horgix:
Starting using GitPython, I quickly encountered a surprising behavior: a `Repo` instance could be initialized with a path like `~/path/to/some/repo` but a `Git` instance couldn't. And indeed, `Repo` expands the path it is given while `Git` doesn't.

This PR is an attempt at fixing this by moving the `_expand_path()` method previously defined in `repo/base.py` to and `expand_path()` method in the `util/` part, and using it both where it was already used and newly in the `Git` constructor.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,5 +19,6 @@ Contributors are:
 -Timothy B. Hartman <tbhartman _at_ gmail.com>
 -Konstantin Popov <konstantin.popov.89 _at_ yandex.ru>
 -Peter Jones <pjones _at_ redhat.com>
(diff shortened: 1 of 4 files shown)


Horgix:
Any chance to get some feedback on this? 🙂

Byron:
Thanks a lot for your contribution, and sorry for the late reply. This looks good to me!

Horgix:
No worries for the delay, thanks for merging 🙂

PR #668, opened 2017-09-08
DUMMY -- do not merge, want to check if appvayor is healthy

yarikoptic:


yarikoptic:
QED -- two last runs in appvayor fail with current master

PR #677, opened 2017-09-30
Retaining env on clone_from

### LINKED ISSUE #676: Environment values passed to Repo.clone_from() are not kept in the resulting repo
r = Repo.clone_from("git@github.com:bpiotr/whatever_repo.git", "/tmp/repo", env={"key": "value"})
    r.git.environment()
    {}

This is not what I expected when I called `Repo.clone_from` and documentation does not elaborate on the transient nature of the `env` here so IMHO this is a bug. 

BTW it costed me quite a while to figure out that `clone_from` actually uses `env` but any subsequent operations on the result needs it to have it set again.

bpiotr:
fixes #676

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,5 +22,6 @@ Contributors are:
 -Anson Mansfield <anson.mansfield _at_ gmail.com>
 -Ken Odegard <ken.odegard _at_ gmail.com>
 -Alexis Horgix Chotard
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for your contribution! It’s great to see an issue created, and the fix provided with the next PR :)!

Issue #678, opened 2017-10-02
create_submodule fails on existing submodule


jakirkham:
Based on the [docs for `create_submodule`]( http://gitpython.readthedocs.io/en/stable/reference.html?highlight=push#git.repo.base.Repo.create_submodule ), it sounds like it proceeds through `Submodule.add`. According to the [docs for `Submodule.add`]( http://gitpython.readthedocs.io/en/stable/reference.html?highlight=push#git.objects.submodule.base.Submodule.add ), it sounds like `add` should still work fine if the submodule already exists. However that does not appear to be the case. Have included an example of the issue experienced below along with details about how to create an environment where this issue can be reproduced.

Example:

<details>

```python
In [1]: import os

In [2]: import git

In [3]: r = git.Repo.clone_from("https://github.com/conda-forge/feedstocks", "feedstocks")

In [4]: r.create_submodule(name="addict", path=os.path.join("feedstocks", "addict"), url="https://github.com/conda-forge/addict-feedstock", branch="master")
Out[4]: ---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/zopt/conda2/envs/test/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

/zopt/conda2/envs/test/lib/python3.6/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    393                             if callable(meth):
    394                                 return meth(obj, self, cycle)
--> 395             return _default_pprint(obj, self, cycle)
    396         finally:
    397             self.end_group()

/zopt/conda2/envs/test/lib/python3.6/site-packages/IPython/lib/pretty.py in _default_pprint(obj, p, cycle)
    508     if _safe_getattr(klass, '__repr__', None) is not object.__repr__:
    509         # A user-provided repr. Find newlines and replace them with p.break_()
--> 510         _repr_pprint(obj, p, cycle)
    511         return
    512     p.begin_group(1, '<')

/zopt/conda2/envs/test/lib/python3.6/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    699     """A pprint that just redirects to the normal repr function."""
    700     # Find newlines and replace them with p.break_()
--> 701     output = repr(obj)
[...]

PR #679, opened 2017-10-02
Store submodule name

jakirkham:
Fixes https://github.com/gitpython-developers/GitPython/issues/678

Seems there are a plethora of things that expect to be able to get a submodule's name. However it is not defined if `Submodule.add` finds the submodule already exists. This is one attempt at trying to solve this issue.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -358,7 +358,9 @@ class Submodule(IndexObject, Iterable, Traversable):
         if sm.exists():
             # reretrieve submodule from tree
             try:
-                return repo.head.commit.tree[path]
+                sm = repo.head.commit.tree[path]
+                sm._name = name
+                return sm
             except KeyError:
                 # could only be in index
                 index = repo.index


Byron:
Thanks for your contribution! At this time I unfortunately also don’t know if there are better ways to fix this, but believe it’s OK to assume it’s not making anything worse if the tests are not breaking.
Ideally there would also be a test that requires the code-change in order to be passing, but I understand the test-suite is messy and impossible to understand, so I wouldn’t ask for it :)!

jakirkham:
> At this time I unfortunately also don’t know if there are better ways to fix this, but believe it’s OK to assume it’s not making anything worse if the tests are not breaking.

After trying a few things locally, this seemed to allow the submodule to behave normally. Though would think the name should be coming from `.gitmodules`. Just wasn't clear to me why that wasn't happening or how to fix it.

> Ideally there would also be a test that requires the code-change in order to be passing, but I understand the test-suite is messy and impossible to understand, so I wouldn’t ask for it :)!

Sorry. Should have included a test. TBH wasn't sure if this seemingly hacky fix would be accepted or so quickly. 😄 Thanks for that. I'll give the test some thought.

PR #681, opened 2017-10-03
Configurable chunk size

### LINKED ISSUE #648: Allow max_chunk_size to be configurable
[Here](https://github.com/gitpython-developers/GitPython/blob/cf8dc259fcc9c1397ea67cec3a6a4cb5816e3e68/git/cmd.py#L659) `max_chunk_size` will usually contain (depending on the platform) `8192`, which is a very high value for a chunk and won't trigger a `.write` (keep reading) until quite some output is printed to stdout. I propose to make that value configurable (maybe via an argument).

My use case:

I have the following class:

	class PipeIO(io.BytesIO):
		def __init__(self, cb):
			io.BytesIO.__init__(self)
			self.updater_cb = cb

		def write(self, b):
			buf = io.BytesIO.getbuffer(self).tobytes()
			progress = re.findall("(?<=\()\d+\/\d+(?=\))", buf.decode("utf-8"))

			try:
				self.updater_cb(progress[-1])
			except:
				pass

			return io.BytesIO.write(self, b)

And I use it like this:

    stdout = PipeIO(self.my_updater_callback)
    self.repo.git.filter_branch("-f", "--env-filter", s, output_stream = stdout)

This will cause my callback to parse the content of the buffer every time [`.write`](https://github.com/gitpython-developers/GitPython/blob/cf8dc259fcc9c1397ea67cec3a6a4cb5816e3e68/git/util.py#L127) is called, which is very useful if I want to provide progress updates while running a long git command (like `filter-branch`).

Possible solution:

Make it possible to call raw commands with one extra argument:

    self.repo.git.filter_branch("-f", "--env-filter", s, output_stream = stdout, max_chunk_size = 32)

bpiotr:
Fixes #648

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -44,10 +44,10 @@ from .util import (
 )
 
 
-execute_kwargs = {'istream', 'with_extended_output', 'with_exceptions',
-                  'as_process', 'stdout_as_string', 'output_stream',
-                  'with_stdout', 'kill_after_timeout', 'universal_newlines',
-                  'shell', 'env'}
+execute_kwargs = {'istream', 'with_extended_output',
+                  'with_exceptions', 'as_process', 'stdout_as_string',
+                  'output_stream', 'with_stdout', 'kill_after_timeout',
+                  'universal_newlines', 'shell', 'env', 'max_chunk_size'}
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot! I am loving the tests, too!

I would happily merge it, it’s just `flake8` having some silly white-space issue: https://travis-ci.org/gitpython-developers/GitPython/jobs/282786566#L1047 .

alexandernst:
This is still not merged because of a single whitespace at the end of line 647. @bpiotr can you fix it or give me push permissions to your repo so I can push the fix there?

bpiotr:
@alexandernst OK, it's fixed now.

alexandernst:
@Byron can we merge this and make a release?

Byron:
Thanks a ton for your contribution! There is another PR in review (#755), and when done, a new release can be cut.

PR #682, opened 2017-10-05
Fix encoding issue with stderr_value and kill_after_timeout

pabelanger:
We don't properly encode our error message under python3.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>


Byron:

PR #683, opened 2017-10-05
Allow user to use kill_after_timeout / as_process with clone_from

pabelanger:
The git.cmd.execute function has the ability to use the
kill_after_timeout variable, however currently the clone function has
as_process hardcoded to True.

If progress is None, we can likley rely directly on git.cmd.execute()
for process handling, which then exposes kill_after_timeout properly.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -900,8 +900,10 @@ class Repo(object):
 
     @classmethod
     def _clone(cls, git, url, path, odb_default_type, progress, **kwargs):
+        as_process = False
         if progress is not None:
             progress = to_progress_instance(progress)
+            as_process = True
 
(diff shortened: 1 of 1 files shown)


pabelanger:
I am not sure about the failures in AppVeyor, they don't look related

Byron:
Thanks for your contribution.
To me it looks like if there is no progress, the function will not wait anymore for the clone to be finished. This, however, is a requirement, the function is expected to be synchronized.
Even though I would think that some tests should fail (on travis) if there is a problem, I would feel more comfortable if there would be a test that explicitly verifies this requirement.
What do you think?

PR #686, opened 2017-10-09
Only gc.collect() under windows

### LINKED ISSUE #553: GitPython `repo.index.commit()` spawns persistent git.exe instance, holds handles to repo

### LINKED ISSUE #605: Performance degradation in 2.1.3
Hi,

The recent commit f1a82e4 has caused a significant performance
degradation in the use of GitPython.  As an example, one unit
test that we have which performs operations on about 42
repositories normally runs in 6.7 seconds, but with the two new
gc.collect() calls in f1a82e4 it now takes 24.2 seconds.

The gitdb.util.mman.collect() call does not seem to be expensive,
only gc.collect().

It's worth noting that most objects in python are freed by use of
reference counting, while gc.collect() invokes the garbage
collector, which is only needed to free objects with reference
cycles.  Normally, the garbage collector runs periodically and
does not need to be explicitly invoked.

Since the garbage collector should only be effective on objects
which no longer have references from active frames, it's not
clear why these two calls to gc.collect() are required.  It would
be nice to know if there are any other ways to address what they
are attempting to fix.  If they truly are required in some
circumstances, it would be good to know what those are and if
there are cases where we do not need to invoke the expense.  And
if they are not required, we should remove them.

Thanks!

jeblair:
Under Windows, tempfile objects are holding references to open files
until the garbage collector closes them and frees them.  Explicit
calls to gc.collect() were added to the finalizer for the Repo class
to force them to be closed synchronously.  However, this is expensive,
especially in large, long-running programs.  As a temporary measure
to alleviate the performance regression on other platforms, only
perform these calls when running under Windows.

Fixes #605

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -209,9 +209,17 @@ class Repo(object):
     def close(self):
         if self.git:
             self.git.clear_cache()
-            gc.collect()
+            # Tempfiles objects on Windows are holding references to
+            # open files until they are collected by the garbage
+            # collector, thus preventing deletion.
+            # TODO: Find these references and ensure they are closed
+            # and deleted synchronously rather than forcing a gc
+            # collection.
+            if is_win:
+                gc.collect()
             gitdb.util.mman.collect()
-            gc.collect()
+            if is_win:
+                gc.collect()
 
     def __eq__(self, rhs):
         if isinstance(rhs, Repo):


Byron:
Thanks a lot for this fix!
Performance tests exist, but unfortunately were never integrated so that they could actually fail a build when a regression is detected.

PR #695, opened 2017-11-05
Fixes issue #694

### LINKED ISSUE #694: Remote URL cannot be accessed for git < 2.7 if SSH isn't working
If the system git is < 2.7, and the URL is one that git expects to use SSH access on, but SSH isn't working for any reason, then `self.repo.git.remote("show", self.name)` [here](https://github.com/gitpython-developers/GitPython/blob/master/git/remote.py#L539) in remote.py will fail with: 

```python
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git remote show origin
  stderr: 'Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.'
```
or related errors if SSH is not installed or on the path.  

This can be fixed by falling back to `git config --get remote.%s.url' % self.name` which just reads the URL directly from the configuration.

rgerkin:
Fixes #694.

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -542,7 +542,7 @@ class Remote(LazyMixin, Iterable):
                         if '  Push  URL:' in line:
                             yield line.split(': ')[-1]
                 except GitCommandError as ex:
-                    if 'correct access rights' in str(ex):
+                    if any([msg in str(ex) for msg in ['correct access rights','cannot run ssh']]):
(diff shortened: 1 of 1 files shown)


Byron:
Sorry for taking so long, and thanks for contributing the fix, it's much appreciated!

PR #699, opened 2017-11-27
Merged 2.1.7 into master (just VERSION changes)

yarikoptic:
I guess just that master diverged from the release tag, that left VERSION stating that current master is 2.1.6 .  Detected while observing installation from git master in our datalad travis setup:
https://travis-ci.org/datalad/datalad/jobs/308052470#L1012

This PR just merges 2.1.7 tag back into master, so also `git describe --tags` provides adequate information about current state of the affairs in master


yarikoptic:
failures are unrelated (I will submit those tiny fixes shortly to fix at least travis), and so ppl do not get confused (I need to instruct them not to pip install from master), I will merge this one

PR #700, opened 2017-11-27
BF: process included files before the rest

### LINKED ISSUE #627: Configparser does not respect order of config files with includes

yarikoptic:
Alternative solution to #628  - may be I have missed why it was more complex than 1 line change? ;)
- [x] Closes #627
- [x] includes test
- [x] also some minor refactoring to tests.lib to provide `SkipTest`, and `skipIf`

It also includes disabled (well -- raises SkipTest) for the fact that in current logic processing of included files is done not right at the point of inclusion.  That makes it behave differently from stock git config


yarikoptic:
I guess I will wait for #702 to be merged first so we could get travis pass here as well

Byron:
@yarikoptic As always, thanks so much for chiming in :)! I merged the referenced 'make travis happy' branch in, and it seems to work just fine.

Issue #525, opened 2016-10-03
Fix "known" Windows errored test cases


ankostis:
Currently, *Windows*  TestCases on Appveyor established on #519 are ["artifically" ok](https://ci.appveyor.com/project/ankostis/gitpython/build/1.0.199).
Because in the code, the Test Cases known to fail on have been marked with the [`HIDE_WINDOWS_KNOWN_ERRORS`](https://github.com/gitpython-developers/GitPython/blob/master/git/test/lib/helper.py#L38) variable so that we get an ALL GREEN response; that way we can *detect when new commits break additional TCs on Windows*.

Currently there are 10 failing TCs (search them with the above var-name), and you can view the errors in this Appveyor job:

  https://ci.appveyor.com/project/ankostis/gitpython/build/1.0.196

They are mostly related either to unicode issues, or to leaked files/resources that cannot be erased.  In order to reproduce on your PC, change locally [`git/test/lib/helper.py:#L38`](https://github.com/gitpython-developers/GitPython/blob/master/git/test/lib/helper.py#L38) from this:

```python
    HIDE_WINDOWS_KNOWN_ERRORS = os.environ.get('HIDE_WINDOWS_KNOWN_ERRORS', True)
```
into this:
```python
    HIDE_WINDOWS_KNOWN_ERRORS = os.environ.get('HIDE_WINDOWS_KNOWN_ERRORS', False)
```

This issue is to remember about these TCs, and to ask for help to fix them.

Byron:
@ankostis Even though I won't be of much help here, I have just provided you with full write permissions on all the gitpython repositories that matter. Previously, I believe you might have been missing write access to smmap, for no other reason than me having forgotten to set it up right away :).

PR #702, opened 2017-11-28
BF (codename "happy travis"): trying to address lints etc to make Travis green again

yarikoptic:
Some minor RFings did creep in, hopefully I didn't change any logic (tests will show ;) ).


yarikoptic:
Travis is happy now, fighting with windows

yarikoptic:
my question would be -- did hooks ever work on windows??? or tests were testing only that they "fail to work"? ;)

yarikoptic:
the last succesful build on appveyor was  https://ci.appveyor.com/project/Byron/gitpython/build/1.0.161  -- at that point there where no tests for succesful hooks execution.  So AFAIK they never worked, and I am offloading that for future to https://github.com/gitpython-developers/GitPython/issues/703 .  In this one I will skip those tests tests to establish a reliable baseline for now

ankostis:
Kudos to tour efforts and happy fighting ;-)

Regarding hooks, at the time of #519 and for some months later all remaining test-cases were passing on Windows, e.g. https://ci.appveyor.com/project/ankostis/gitpython/build/1.0.433
You may read #525 where i explain the trick to artificially pass Appveyor, and use it to mark also the hooks-TCs as permanently failing.

BTW you may check if your recent efforts have fixed some of those long disabled TCs.

yarikoptic:
@ankostis -- join the "fun"!  but I think/hope I am "done"

hooks -- as of https://ci.appveyor.com/project/ankostis/gitpython/build/1.0.433 you reference, there were no test for testing **successful** execution of a hook on windows .

Thanks for the pointer to `HIDE_WINDOWS_KNOWN_ERRORS ` -- I will use it instead of plain is_win (just will wait for tests to finish or fail to get a baseline ;))

As for the 'check' -- someone should just check if there is a similar to travis `allow_failures`, if so -- then there could be a dedicated matrix run which would do that for us, without requiring manual messing around... but that later -- for now just want it to get green.

yarikoptic:
But in the future (after we are green again), it would be nice if no PR was merged if tests fail...

yarikoptic:
woohoo -- some of the windows environments are green now! rebuilding to see if some spurious one didn't creep in.  any immediate clue about what could be wrong in those anaconda windows ones @ankostis ?

yarikoptic:
ok, @Byron , @ankostis et al, it is ready for the review.  I will leave the appveyor crusade for later/someone else ATM (note a handy addition to .appveyor.yml) - at least some environments are in clear there as well, and travis is all green.  Would be nice to establish the baseline of passing CI in master.

ankostis:
I think it is not anaconda to blame but rather Cygwin's Git!
Since `git-2.8+`, I've seen huge problems with joining local paths, causing most of the 16 errors in the last 2 environments.

For instance it doesn't even work the command `pip install git+https://github.com/foo/bar.git`.
I believe it is this problem: http://lists-archives.com/git/894111-cygwin-git-cannot-install-python-packages-with-pip.html

yarikoptic:
Yikes

ankostis:
`pip install git+http` scheme failure on cygwin reported as pypa/pip#4895.

yarikoptic:
Well... would be nice to get it reviewed/merged at this stage of success so that at least some baseline is established back again, and then tackle cygwin in a more dedicated PR ;)

git/test/lib/helper.py
```diff
@@ -139,7 +139,7 @@ def repo_creator(self):
             try:
                 try:
                     return func(self, rw_repo)
-                except:
+                except:  # noqa E722
```
hugovk: How about `except Exception:` like elsewhere?

git/test/lib/helper.py
```diff
@@ -296,7 +296,7 @@ def remote_repo_creator(self):
                     with cwd(rw_repo.working_dir):
                         try:
                             return func(self, rw_repo, rw_daemon_repo)
-                        except:
+                        except:  # noqa E722
```
hugovk: How about `except Exception:` like elsewhere?

git/test/lib/helper.py
yarikoptic: We could. But at this out-most level of handling execution, I thought to remain as "greedy" as possible.  May be someone (among used dependencies) still manages to raise non-Exception "exceptions" up the chain from somewhere/somehow -- so I thought to remain catching them as well.  In the other places it was more "specific" to not expect some complete weird stuff being thrown up.  But if you insist, we could have it Exception here as well

Byron:
@yarikoptic Thanks so much for hanging into this! Especially your crusade to get the windows CI green is particularly brave!! @ankostis certainly has a tale to tell there, too!
I am merging this as it already is getting us some steps towards a somewhat sane Christmas release :D! Full disclaimer: I am on vacation now to the end of the year, and intend to get my inbox empty and spend the time it takes.
@yarikoptic @ankostis I understand I was the bottleneck for the upcoming release, and can't even promise betterment. However, if you are not yet able to cut a release on your own, but would like to, I am happily assigning you the rights needed and add your GPG signature keys to the ones officially endorsed by this repository. Just let me know, and we will make it happen.

yarikoptic:
Thanks @Byron , I hope you will have a nice (and productive in all regards) vacation!
Here is my primary key, (public portion) should be available from public servers
```
pub   rsa4096/A2DE235062DA33FA 2010-06-07 Yaroslav Halchenko (Yarik) <yoh@onerussian.com>
 Primary key fingerprint: C5B9 05F0 E8D9 FD96 68FF  366F A2DE 2350 62DA 33FA
```
I usually (for other projects) "tag -s" releases locally, push the tags, create release out of them on github

I am yarikoptic on pypi

Byron:
@yarikoptic Thanks - I added you to the list of maintainers on PyPi and your key to the `release-verification-key.asc`, so you should be able to make your own signed releases from now on. Just to stay in training, I would like to do `v2.1.8` though :).

yarikoptic:
oh yeah, I would prefer not to release unless I really have to ;) so I am ok to stay in training for as long as it could be ;-)

PR #736, opened 2018-03-18
Allow mmap not just for py2.6, 2.7 and 3.6+ but also 3.0+

hugovk:
This version check only checks the minor number: the `y` in Python `x.y`.

So mmap is available for Python 2.6 and 2.7, and 3.6+, but not for 3.0 to 3.5.

But I think the intent was to have memory maps for Python 2.6 and up, including all of Python 3.x (and 4.x etc.).

As Python 2.6 is the minimum supported version, the `allow_mmap` is always `True` for supported versions.

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -8,7 +8,6 @@ from io import BytesIO
 import os
 from stat import S_ISLNK
 import subprocess
-import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


Byron:
What a great catch! Thanks a lot for your contribution!

Issue #540, opened 2016-10-21
2.0.9 fails on Python 2.6


andy-maier:
I was thankful to see your reaction to my issue #453, but unfortunately GitPython 2.0.9 has some issues on Python 2.6. GitPython 2.0.8 worked like a charm for us.

The issues are:
- uses of dictionary comprehension in `git/cmd.py` around line 800 (dict comprehensions were introduced in py27).
- use of SkipTest and skipIf from `unittest.case` in `git/util.py` and in several test cases (these were added to `unittest` in py27). Also, the Python `unittest` docs suggest they be imported from `unittest` and not from `unittest.case`, but that is minor.

I have fixed those locally and will create a PR for your review.

hugovk:
PR #541 was merged on 8 Dec 2016. Can this issue be closed?

Whilst on the topic, is Python 2.6 still something you need to support, or might it be time to consider dropping it more fully?

Here's the pip installs for GitPython from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  82.08% |        466,982 |
| 3.6            |  10.07% |         57,285 |
| 3.5            |   4.88% |         27,754 |
| 3.4            |   2.58% |         14,663 |
| 2.6            |   0.23% |          1,333 |
| 3.7            |   0.10% |            571 |
| 3.3            |   0.06% |            321 |
| 3.8            |   0.00% |              5 |

Source: `pypinfo --start-date -39 --end-date -12 --percent --pip --markdown GitPython pyversion`

Byron:
@hugovk  Thanks a lot for publishing some hard facts about its usage here! I wasn't even aware of download information once again being available.

I think you are providing a strong argument to indeed drop python 2.6 support, the next time we break it. Alternatively, if someone feels strongly about getting rid of it to improve the code climate for example, PRs are always welcome.

As the PR mentioned in this issue was merged, I think it's fair to close this ticket, too.

hugovk:
Yeah, there's quite a lot of interesting stuff you can now get out of PyPI and pypinfo:  
https://github.com/ofek/pypinfo/blob/master/README.rst#usage

Here's some examples:
```console
$ pypinfo GitPython country
Served from cache: False
Data processed: 8.22 GiB
Data billed: 8.22 GiB
Estimated cost: $0.05

| country | download_count |
| ------- | -------------- |
| US      |        548,405 |
| IE      |         13,000 |
| DE      |         12,170 |
[...]

PR #737, opened 2018-03-18
Drop support for EOL Python 2.6 and 3.3

hugovk:
As a follow on from https://github.com/gitpython-developers/GitPython/issues/540#issuecomment-373905431, this drops support for Python 2.6 and 3.3, both of which are EOL and no longer receiving security updates, or any updates, from the core Python team.

They're also little used. Here's the pip installs for GitPython from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  82.08% |        466,982 |
| 3.6            |  10.07% |         57,285 |
| 3.5            |   4.88% |         27,754 |
| 3.4            |   2.58% |         14,663 |
| 2.6            |   0.23% |          1,333 |
| 3.7            |   0.10% |            571 |
| 3.3            |   0.06% |            321 |
| 3.8            |   0.00% |              5 |

Source: `pypinfo --start-date -39 --end-date -12 --percent --pip --markdown GitPython pyversion`

Dropping Python 2.6 means some old compatibility and 2.6-specific code can be removed, and can also be updated to use more modern features of Python 2.7+, some of which has been applied here.

Python 3.3 is also dropped here. Whilst there's less benefit to be had from code changes (compared to 2.6), it's unsupported and used even less that 2.6 (in this project and in general), so there's no strong reason to keep it and we might as well free up some CI slots.


Byron:
Thank you so much for this massive cleanup!
To me it would have been utterly impossible to do it! It's contributors like you who keep this project going!

I will have a closer look in the following days, and invite @yarikoptic and @ankostis to do the same.

**What do you all think how this change should be released?** In theory it would have to be a major version jump, as we might break some people. On the other hand, the API didn't change so the release might be confusing and repelling to everyone else.

From that point of view, it might not be more than a patch release.

hugovk:
You're welcome, and thanks for the maintenance!

---

Some projects bump major versions to indicate dropping support, some don't. 

If following semver, it depends if you consider it a backwards incompatible changes:

> 8. Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.

https://semver.org/#spec-item-8

I probably would bump it, but don't have strong opinions. 

---

It's good 2.18 was already released with `python_requires` (https://github.com/gitpython-developers/GitPython/commit/c22f1b05fee73dd212c470fecf29a0df9e25a18f), which will help pip install the right version for people still on old Pythons. See:
* https://hackernoon.com/phasing-out-python-runtimes-gracefully-956f112f33c4
* https://github.com/pypa/python-packaging-user-guide/issues/450
* https://github.com/pypa/python-packaging-user-guide/compare/master...tonybaloney:python_version_tutorial

Byron:
And since we don't change the public API in that sense, I think a major version bump shouldn't be required. Also I would rather upset those behind ~1000 downloads, rather than confusing (or loosing) those behind the 500k downloads.

I will wait a week for more commends and/or opinions to come in, and if there are no objects, will release a new version next weekend.

test-requirements.txt
```diff
@@ -3,4 +3,4 @@
 coverage
 flake8
 nose
-mock; python_version<='2.7'
+mock; python_version=='2.7'
```
yarikoptic: so version here could never include ".patch" (ie to be `2.7.5`)? if it could then would need to retain `<=`

doc/source/changes.rst
```diff
@@ -207,11 +207,11 @@ It follows the `semantic version scheme <http://semver.org>`_, and thus will not
   * Repo.working_tree_dir now returns None if it is bare. Previously it raised AssertionError.
   * IndexFile.add() previously raised AssertionError when paths where used with bare repository, now it raises InvalidGitRepositoryError
 
-* Added `Repo.merge_base()` implementation. See the `respective issue on github <https://github.com/gitpython-developers/GitPython/issues/169>`_
+* Added `Repo.merge_base()` implementation. See the `respective issue on GitHub <https://github.com/gitpython-developers/GitPython/issues/169>`_
 * `[include]` sections in git configuration files are now respected
 * Added `GitConfigParser.rename_section()`
 * Added `Submodule.rename()`
-* A list of all issues can be found `on github <https://github.com/gitpython-developers/GitPython/issues?q=milestone%3A%22v0.3.6+-+Features%22+>`_
+* A list of all issues can be found `on GitHub <https://github.com/gitpython-developers/GitPython/issues?q=milestone%3A%22v0.3.6+-+Features%22+>`_
```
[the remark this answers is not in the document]
hugovk: Yes, that would be sensible.

Would you like me to do that as part of this PR, or should it wait? If now, what's version will the next release be?

test-requirements.txt
hugovk: Not too sure. Shall I revert this to be on the safe side?

PR #740, opened 2018-03-27
 Enable git diff-tree with empty second tree-ish param 

gangefors:
Executing `git diff-tree` and omitting the second tree-ish argument
wasn't possible before. This PR enables this behaviour by allowing
the caller to send in an empty string.

Allowing an empty string keeps backwards compatibility for all other
allowed input types.


gangefors:
The appveyor builds that fail has failed before and are not related to this PR.

gangefors:
It might be that the thing I'm trying to achieve here is the same result I get if I use `git.diff.NULL_TREE`. Feel free to ignore and close if that is the case.

Byron:

PR #744, opened 2018-04-03
git: index: base: use os.path.relpath

### LINKED ISSUE #743: Adding files to repository that is located directly in the root
This is a weird scenario, but if you init git repo in the root directory(e.g. D:\\) and try to add a file using gitpython, it will throw a ValueError:
```
Absolute path `D:\\file` is not in git repository at `D:\\`
```
The reason for this is improper validation in _to_relative_path method located at git/index/base.py:
```
    def _to_relative_path(self, path):
        """:return: Version of path relative to our git directory or raise ValueError
        if it is not within our git direcotory"""
        if not osp.isabs(path):
            return path
        if self.repo.bare:
            raise InvalidGitRepositoryError("require non-bare repository")
        relative_path = path.replace(self.repo.working_tree_dir + os.sep, "")
        if relative_path == path:
            raise ValueError("Absolute path %r is not in git repository at %r" % (path, self.repo.working_tree_dir))
        return relative_path
```
I.e. when path is `/file` and working_tree_dir is `/`, relative_path stays equal to path, because `self.repo.working_tree + os.sep` is equal to `//`. The way to solve this is to simply use os.path.relpath methond. Please see pull-request attached.

efiop:
Fixes #743

Signed-off-by: Ruslan Kuprieiev <kupruser@gmail.com>

### DIFF
--- a/git/test/test_index.py
+++ b/git/test/test_index.py
@@ -838,6 +838,21 @@ class TestIndex(TestBase):
         r.index.add([fp])
         r.index.commit('Added [.exe')
 
+    def test__to_relative_path_at_root(self):
+        root = osp.abspath(os.sep)
+
+        class Mocked(object):
+            bare = False
+            git_dir = root
+            working_tree_dir = root
+
+        repo = Mocked()
+        path = os.path.join(root, 'file')
+        index = IndexFile(repo)
+
+        rel = index._to_relative_path(path)
+        self.assertEqual(rel, os.path.relpath(path, root))
+
     @with_rw_repo('HEAD', bare=True)
     def test_pre_commit_hook_success(self, rw_repo):
         index = rw_repo.index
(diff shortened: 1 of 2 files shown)


git/test/test_index.py
```diff
@@ -838,6 +838,19 @@ def test_add_a_file_with_wildcard_chars(self, rw_dir):
...
+
+        class Mocked(object):
+            bare = False
+            git_dir = root
+            working_tree_dir = root
+
+        repo = Mocked()
+        path = os.path.join(root, 'file')
+        index = IndexFile(repo)
+        index._to_relative_path(path)
```
Byron: Is there an assertion missing? It looks like `_to_relative_path` returns the relative path, which is not verified here.

efiop: Ah, right. Thank you, fixed in a new version.

Byron:
Thanks a lot! Looking good!

PR #773, opened 2018-06-14
Update test_docs.py

markrowsoft:
Using "import as" is normally a time saver but for usability of the documentation, please consider removing osp and join with fully qualified calls for better snippet readability.

I had some co-workers end up using popen commands vs using this library when I recommended GitPython as they did not understand how it worked from skimming the docs.

Sorry for not writing a test that fails... only a syntax change for readability.


Byron:
Thanks a lot for your contribution! Along with that other PR which fixes the indentation, the tutorial should be so much easier to read and understand.

Issue #619, opened 2017-04-07
Raising of GitCommandError fails in archive function if stdout was a cStringIO object


wohauser:
I tried to check out an archive for a not existing path from the repository and I expect a GitCommandError exception.

I got following TypeError exception:
"Expected bytes or text, but got <cStringIO.StringO object at 0x7fabac6f97a0>"

I did following:
fpt = StringIO() 
-- (used as a temporary storage for the tar archive)
self._repository.archive(fpt, treeish=version_identifier, format="tar", path=sub_path) 
-- (where sub_path points to an not existing path)

The cause of the problem was, the Constructor of the GitCommandError could not handle the cStringIO object. (a normal file pointer may be faulty too). 
I suppose it should set its stdout variable to None in case of not handled types. Or it should be adjusted when raising the exception.

Byron:
Using the latest version of GitPython (39335e6242c93d5ba75e7ab8d7926f5a49c119a3) on OSX on python 2.7, I was unable to reproduce the issue. 

![screen shot 2017-04-09 at 16 38 38](https://cloud.githubusercontent.com/assets/63622/24838187/3f7af16a-1d43-11e7-9cb9-657c513051fa.png)

Please feel free to provide more information to help reproducing it - maybe this functionality is platform dependent. In the meanwhile, I am closing this issue.

wohauser:
Try it with a path filter using a path that not exists in the repo.

e.g.
r.archive(fpt,format="tar",path="tttt")

results in 
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/git/repo/base.py", line 969, in archive
    self.git.archive(treeish, *path, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/git/cmd.py", line 425, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/git/cmd.py", line 877, in _call_process
    return self.execute(call, **exec_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/git/cmd.py", line 688, in execute
    raise GitCommandError(command, status, stderr_value, stdout_value)
  File "/usr/local/lib/python2.7/dist-packages/git/exc.py", line 74, in __init__
    super(GitCommandError, self).__init__(command, status, stderr, stdout)
  File "/usr/local/lib/python2.7/dist-packages/git/exc.py", line 54, in __init__
[...]

PR #782, opened 2018-07-24
Fix TypeError in git.execute(..., output_stream=file)

the-owl:
This fixes issue #619 - raise GitCommandError(not TypeError) when output_stream is set
in `git.execute(...)`.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -794,7 +794,7 @@ class Git(LazyMixin):
             else:
                 max_chunk_size = max_chunk_size if max_chunk_size and max_chunk_size > 0 else io.DEFAULT_BUFFER_SIZE
                 stream_copy(proc.stdout, output_stream, max_chunk_size)
-                stdout_value = output_stream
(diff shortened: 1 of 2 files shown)


git/cmd.py
```diff
@@ -794,7 +794,7 @@ def _kill_process(pid):
             else:
                 max_chunk_size = max_chunk_size if max_chunk_size and max_chunk_size > 0 else io.DEFAULT_BUFFER_SIZE
                 stream_copy(proc.stdout, output_stream, max_chunk_size)
-                stdout_value = output_stream
+                stdout_value = proc.stdout.read()
```
Byron: That's probably the only way to do this properly, even though it seems funky.  Ideally `stdout_value` should be the value read, but in this case, it will always be an empty string as `read()` is called right after `stream_copy(...)`.
However, looking at it, the whole implementations seems to have gotten somewhat convoluted over the years, so it's probably alright after all.

Byron:
Thanks a lot! This fix is much appreciated.

Issue #784, opened 2018-07-26
Stop stringifying None in clone kwargs


nvllsvm:
Kwargs to `git.Repo.clone` and `git.Repo.clone_from` with `None` values get cast to strings and passed to the git command. This leads to behavior I'd consider non-Pythonic as it's exposing a Python internal null value to something which has no understanding of it.

For example, `git.Repo.clone_from(url, path, mirror=None)` results `git clone --mirror=None -v url path`. This is not valid as `--mirror` takes no arguments.

Instead, a `None` value should result in one of two things in the git command:

- Exclude the kwarg entirely
- Include the kwarg, but treat it as a flag without arguments (ex. `--mirror` without the `=`).

PR #785, opened 2018-07-27
Exclude kwarg when None

nvllsvm:
Resolves https://github.com/gitpython-developers/GitPython/issues/784


Byron:
Thanks a lot! Much appreciated!

PR #788, opened 2018-08-07
Include test files in sdists

toddrme2178:
These files are missing from sdists but are needed for the unit tests as far as I can determine.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,7 +5,13 @@ include AUTHORS
 include README
 include requirements.txt
 
+include .gitattributes
+include .gitignore
+include .gitmodules
+
+recursive-include .git *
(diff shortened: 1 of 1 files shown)


Byron:
Right now I am not on a computer in the traditional sense and can’t verify GitPython source-distributions are still working with the change.
Once that is done, I will happily merge.

toddrme2178:
One possibility might be to create a small git project just for the tests
and include that in the source tree.  That has the advantage and
disadvantage of being independent of the main source tree (advantage that
it is consistent across any change to the source tree, disadvantage that
changes to the source tree can highlight errors you didn't think to
check).  However, I would understand if you think this is too much work and
prefer to keep things the way they are.

>

PR #793, opened 2018-09-10
Document support for Python 3.7

michael-k:


michael-k:
Turns out that despite tests running against 3.7-dev, GitPython is not compatible with Python 3.7.  The reason is [PEP 479](https://www.python.org/dev/peps/pep-0479/) which was [released as part of Python 3.7](https://docs.python.org/3/whatsnew/3.7.html#changes-in-python-behavior).

michael-k:
@Byron Commit d6d89ad, the second one of the two included in this PR, should fix the mentioned problem in a backwards compatible way.

GitPython uses `raise StopIteration` inside generators directly or indirectly (calling `next` without handling the exception) to stop them.  As [PEP 479](https://www.python.org/dev/peps/pep-0479/) this was never the right thing to do and Python 3.7 does no longer support this.

Most projects that run their test suite against Python 3.7-dev (aka 3.7.0a4) on travis are also compatible with Python 3.7.  There are a few examples and this project is one of them.  This is what my comment was about.  See https://travis-ci.com/michael-k/GitPython/builds/84335704 which ran the test suite against Python 3.7 and https://travis-ci.com/michael-k/GitPython/builds/84338644 which did the same thing, but without using `StopIteration`.

However I missed that nightly also needs an update.  I've fixed that and rebased.

AnneTheAgile:
@michael-k but the CI job on the latest commit said it failed for two pythons. Did you see that? 
For example py2.7 first failure is; 
https://ci.appveyor.com/project/Byron/gitpython/builds/19502926/job/riw6uo91dbk292fh
````
ERROR: test_init (git.test.test_repo.TestRepo)
  File "C:\projects\gitpython\git\test\test_repo.py", line 267, in test_init
    rc = r.clone(clone_path)
  File "C:\projects\gitpython\git\repo\base.py", line 982, in clone
    return self._clone(self.git, self.common_dir, path, type(self.odb), progress, **kwargs)
  File "C:\projects\gitpython\git\repo\base.py", line 948, in _clone
    finalize_process(proc, stderr=stderr)
  File "C:\projects\gitpython\git\util.py", line 333, in finalize_process
    proc.wait(**kwargs)
  File "C:\projects\gitpython\git\cmd.py", line 415, in wait
    raise GitCommandError(self.args, status, errstr)
GitCommandError: Cmd('C:\cygwin\bin\git.exe') failed due to: exit code(128)
  cmdline: C:\cygwin\bin\git.exe clone -v /cygdrive/c/users/appveyor/appdata/local/temp/1/repos/foo/bar.git c:\users\appveyor\appdata\local\temp\1\repos\foo\bar.git_clone
  stderr: 'Cloning into 'c:\users\appveyor\appdata\local\temp\1\repos\foo\bar.git_clone'...
fatal: Invalid path '/cygdrive/c/users/appveyor/appdata/local/temp/1/repos/foo/bar.git/c:\users\appveyor\appdata\local\temp\1\repos\foo\bar.git_clone': No such file or directory
````
I'm not quite sure what's up with that.

michael-k:
@AnneTheAgile yeah, but is this related to this PR?  They are also [failing on master](https://ci.appveyor.com/project/Byron/gitpython/builds/19495100) and the log looks similar enough to me.  (I didn't compare the logs in detail.)

It looks like not a single test run of master on AppVeyor was successful.

PR #796, opened 2018-09-14
Respect _common_dir when finding repository config file

delego-wlritchi:
Among other things, remotes are now correctly identified when in a separate worktree.


Byron:
Thanks a lot for your contribution! Even though I am not sure if the test fails without the change of logic, I believe it's save enough to merge anyway as there is at least no further breakage.

PR #797, opened 2018-09-18
do not glob a path if it exists 'unglobbed'

### LINKED ISSUE #680: Index.add fail silently when path contains character `[`
Hello,

on windows when the path passed as an argument given to index.add contains a `[`,  the function just return an empty list, whether or not the path exists.

```
r = git.Repo('.')
r.index.add(['test[test'])
Out[88]: []
```

Kind Regards and thanks for this projects.

sdementen:
fix #680

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -377,13 +377,13 @@ class IndexFile(LazyMixin, diff.Diffable, Serializable):
                     continue
             # end check symlink
 
-            # if abs_path exists, yield it, otherwise resolve globs if possible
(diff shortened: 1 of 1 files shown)


Byron:
@sdementen Thanks a lot for your contribution! Unfortunately some tests seem to legitimately fail. Do you think that can be fixed?
If that should work out, it would be great to have a specific test as well that would fail the current change is not applied.
Thank you

Byron:

PR #801, opened 2018-09-28
read workdir from git.config as referenced in man 1 git-config

Farom:
The setting in .git/config for GIT_WORK_TREE were ignored.
I tried to implement this as git-config says.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -140,7 +140,20 @@ class Repo(object):
             # removed. It's just cleaner.
             if is_git_dir(curpath):
                 self.git_dir = curpath
-                self._working_tree_dir = os.getenv('GIT_WORK_TREE', os.path.dirname(self.git_dir))
+                # from man git-config : core.worktree
+                # Set the path to the root of the working tree. If GIT_COMMON_DIR environment
+                # variable is set, core.worktree is ignored and not used for determining the
+                # root of working tree. This can be overridden by the GIT_WORK_TREE environment
+                # variable. The value can be an absolute path or relative to the path to the .git
+                # directory, which is either specified by GIT_DIR, or automatically discovered.
+                # If GIT_DIR is specified but none of GIT_WORK_TREE and core.worktree is specified,
+                # the current working directory is regarded as the top level of your working tree.
+                self._working_tree_dir = os.path.dirname(self.git_dir)
+                if os.environ.get('GIT_COMMON_DIR') is None:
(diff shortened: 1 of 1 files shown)


git/repo/base.py
```diff
@@ -140,7 +140,20 @@ def __init__(self, path=None, odbt=GitCmdObjectDB, search_parent_directories=Fal
...
+                # variable. The value can be an absolute path or relative to the path to the .git
+                # directory, which is either specified by GIT_DIR, or automatically discovered.
+                # If GIT_DIR is specified but none of GIT_WORK_TREE and core.worktree is specified,
+                # the current working directory is regarded as the top level of your working tree.
+                self._working_tree_dir = os.path.dirname(self.git_dir)
+                if os.environ.get('GIT_COMMON_DIR') is None:
+                    gitconf = self.config_reader("repository")
+                    if gitconf.has_option('core', 'worktree'):
+                        self._working_tree_dir = gitconf.get('core', 'worktree')
+                self._working_tree_dir = os.getenv('GIT_WORK_TREE', self._working_tree_dir)
```
Byron: This looks like in the end, the old behaviour is always taking precedence, as `_working_tree_dir` is written unconditionally.

Byron:
Thanks a lot for your contribution! It's much appreciated.
I also added a comment about something puzzling, maybe you can have a look.
Thank you

git/repo/base.py
Farom: I do not think so:

Set _working_tree_dir to git_dir.
if env(GIT_COMMON_DIR) is unset, set  _working_tree_dir to core.worktree, if it exists.
If env(GIT_WORK_TREE) is set, use it or use the value before.

Do you think line 156 would be more pythonic this way:?

```
if os.environ.has_key('GIT_WORK_TREE'):
     self._working_tree_dir = os.getenv('GIT_WORK_TREE')
```

Byron: Thank you @ukleinek , thanks @Farom , I think I understand now. Given that I was easily tricked by it, maybe that indicates we should indeed put the code behind an `if` condition. That way, we don't do all the work if, in the end, we are willing to throw it away for the value which is in `GIT_WORK_TREE`.
Once this change is made, I believe we can merge this one.

Byron:
One change request was made, after which this PR can be merged. See the resolved conversation in `git/repo/base`.

Byron:

PR #808, opened 2018-11-08
read workdir from git.config as referenced in man 1 git-config

fscherf:
Hi @Byron,

I am a co-worker of @Farom and we need his PR #801 merged for one of our future projects. He had no time to add the last feedback so i edited his patch.


fscherf:
i know the tests fail, but is this my fault? Both fails seem like ci config problems.

Byron:
Sorry for the late reply, and thanks a lot for the contribution.

I do hope to get the tests fixed at some point, being in a red state for too long makes me quite fearful of merging more than typo-fixes in future.

PR #810, opened 2018-11-20
Fix docstring in cmd module

dschaller:


Byron:
Thanks a lot! Much appreciated.

PR #822, opened 2019-01-09
Fixed error in documentation

Syoc:
The renamed_file function contains the following which ends up on readthedocs: 
  :note: This property is deprecated, please use ``renamed_file`` instead.

Removed the line


Byron:
Thanks a lot for catching this, and for the fix!

PR #825, opened 2019-01-15
Travis CI: 'sudo' tag is now deprecated in Travis

cclauss:
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 python:
   - "2.7"
   - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,8 @@ matrix:
   include:
     - python: 3.7
       dist: xenial
-      sudo: required
     - python: "nightly"
       dist: xenial
-      sudo: required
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,10 +9,8 @@ matrix:
   include:
     - python: 3.7
       dist: xenial
```
jeking3: You might be able to move `dist: xenial` out one level and define it globally.  If so, you can move the python declarations out to the global list and eliminate the matrix include section all together.

cclauss: I can confirm that both solutions work. While xenial is generally available, it is not the Travis default so I usually take the more conservative route of only using it where it is needed.

cclauss: #831 eliminates one test when run on xenial so I think it is best to run more tests on Pythons that can run on the Travis CI default distro and fewer tests only on Python 3.7 and nightly.

jeking3:
Python 3.7 passes builds on Windows...

PR #830, opened 2019-01-21
Define PermissionError in Python versions < 3.3

cclauss:
Simplify the logic by defining [__PermissionError__](https://docs.python.org/3/library/exceptions.html#PermissionError) in Python 2 in alignment with Python porting best practice [___use feature detection instead of version detection___](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -43,6 +43,10 @@ from .util import (
     stream_copy,
 )
 
+try:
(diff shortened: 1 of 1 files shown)


cclauss:
Closing in favor of #826.

PR #840, opened 2019-02-08
allow calling index.add, index.move and index.remove with single items

### LINKED ISSUE #813: repo.index.add -> check if input is a list
For ~~stupid~~ people like me:
Could you please add a check if the input to `repo.index.add(...)` is a list, and not a string?
It took me a while to find out that this was the real error behind:

> ValueError: Absolute path '\\' is not in git repository at 'C:\\Users\\xxx\\yyy\\zzz'

:-)

MarcelBeining:
added testing for it
closes #813

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -569,16 +569,23 @@ class IndexFile(LazyMixin, diff.Diffable, Serializable):
         """ Split the items into two lists of path strings and BaseEntries. """
         paths = []
         entries = []
-
-        for item in items:
-            if isinstance(item, string_types):
-                paths.append(self._to_relative_path(item))
-            elif isinstance(item, (Blob, Submodule)):
-                entries.append(BaseIndexEntry.from_blob(item))
-            elif isinstance(item, BaseIndexEntry):
-                entries.append(item)
-            else:
-                raise TypeError("Invalid Type: %r" % item)
+        
+        if isinstance(items, string_types):
+            paths.append(self._to_relative_path(items))
+        elif isinstance(items, (Blob, Submodule)):
+            entries.append(BaseIndexEntry.from_blob(items))
+        elif isinstance(items, BaseIndexEntry):
+            entries.append(items)
+        else:            
+            for item in items:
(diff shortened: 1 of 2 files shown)


MarcelBeining:
hm guys, I am not that into testing to understand why the tests failed in some environments, maybebe you can help me. =/ thanks

The only thing I see is that in AppVeyor, the test fails when I try to remove the added file again from the index with  
`stderr: 'fatal: pathspec 'git\test\test_index.py' did not match any files`
however why is this file there in some environments but not there in others?

git/index/base.py
```diff
@@ -569,16 +569,23 @@ def _preprocess_add_items(self, items):
...
-            if isinstance(item, string_types):
-                paths.append(self._to_relative_path(item))
-            elif isinstance(item, (Blob, Submodule)):
-                entries.append(BaseIndexEntry.from_blob(item))
-            elif isinstance(item, BaseIndexEntry):
-                entries.append(item)
-            else:
-                raise TypeError("Invalid Type: %r" % item)
+        
+        if isinstance(items, string_types):
```
Byron: I think the repetition can and should be removed. My suggestion is to check if the input type is something you can iterate, and if not, pack it into a list. The rest of the code would then remain unchanged.
Does that make sense?

git/index/base.py
```diff
@@ -801,13 +808,18 @@ def _items_to_rela_paths(self, items):
...
         may be absolute or relative paths, entries or blobs"""
         paths = []
-        for item in items:
-            if isinstance(item, (BaseIndexEntry, (Blob, Submodule))):
-                paths.append(self._to_relative_path(item.path))
-            elif isinstance(item, string_types):
-                paths.append(self._to_relative_path(item))
-            else:
-                raise TypeError("Invalid item type: %r" % item)
+        if isinstance(items, (BaseIndexEntry, (Blob, Submodule))):
```
Byron: I think alternative implementation suggested above should also be applied here.

Byron:

PR #841, opened 2019-02-13
Fix commit.py documentation

### LINKED ISSUE #806: commit requires author to be Actor, not string
Documentation is wrong about the author parameter: it has to be an Actor, not a string.

CptMikhailov:
Closes #806: author parameter requires Actor type, not string.

Updates documentation to reflect this.


Byron:
Thank you

PR #844, opened 2019-02-16
Added usage example to Repo __init__.py call for Windows users

alexqfredrickson:
This is a non-functional change intended to offer guidance to Windows users.  It's a little confusing that, say, `C:\GitHub\GitPython` is not a valid Git repository - the Repo object seems to expect a `.git` directory during initialization.


Byron:
Thank you!

PR #853, opened 2019-03-15
Fix setup.py and use of requirements files

jeking3:
There were requirements scattered about in both setup.py and requirements.txt - this consolidated them all into requirements files.  It also adds python 3.5 and 3.6 to the tests.


jeking3:
@Byron is it important to maintain support for Python 3.4?  It doesn't understand some of the setup.py changes.  I could mark the project as requiring Python 3.5 or later.  This PR adds 3.6 and 3.7 testing to appveyor and it consolidates all the requirements logic for setup.  Next PR will set some of the currently failing appveyor tests as allowed failures (miniconda variants).  Thanks.

PR #859, opened 2019-04-05
Fix #858

### LINKED ISSUE #858: remote.Remote.urls fails
It seems the fix for #694 only works if the current workin directory of the running program is within the git repository.

If the current working dir is outside the git repo, then the following (starting at `git/remote.py:557`)
```
                        result = Git().execute(
                            ['git', 'config', '--get', 'remote.%s.url' % self.name]
                        )
```
fails with:
```
git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git config --get remote.origin.url
```

sstancu:
This patch ensures that `git config --get remote.the_remote.url` is always ran inside the correct current working directory, and fixes #858

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,5 +29,6 @@ Contributors are:
 -Tim Swast <swast _at_ google.com>
 -William Luc Ritchie
 -David Host <hostdm _at_ outlook.com>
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for the fix! It's much appreciated!

PR #860, opened 2019-04-05
added support for non-ascii directories and file names

xarx00:
This fix allows support for non-ascii filenames, both in workdir and repos. For instance, when your workdir contains non-ascii (0-255) characters, `git.Repo.clone_from()` didn't work.

Implementation:
It turned out that the amount of necessary changes is not big. The main problem was in that git.compat.defenc was set to sys.getdefaultencoding() instead of sys.getfilesystemencoding().
`sys.getfilesystemencoding()` requires Python 2.3. The `from builtins import str` fixes several str(epath) etc. conversions that were causing UnicodeEncodeError. I have no idea what python2 version is necessary for that.

I have tested the fix on Windows7 Pro x64, both with Python 2.7.13 and 3.7.3.


Byron:
Thanks a lot, and I would love to merge this. However, it appears the PR breaks [a single test in Py 2.7](https://travis-ci.org/gitpython-developers/GitPython/jobs/516428362#L1018), and I am not entirely sure these are related. So I am merging to test it in master, and revert as needed.

Byron:
[Unfortunately the build still fails on top of master](https://travis-ci.org/gitpython-developers/GitPython/jobs/561334516#L634) for the same reason as before. I have no idea why this is happening.
What do you recommend?

PR #862, opened 2019-04-13
Fix `AttributeError` when searching a remote by name

cesarizu:
Running code like `'origin' in git.Repo('path/to/existing/repository').remotes`
raises an AttributeError instead of returning a boolean. This commit fixes that
behaviour by catching the error when doing an identity match on `IterableList`.

Minimum test case with version 2.1.11:

```python
>>> 'origin' in git.Repo('path/to/existing/repository').remotes
  File "/home/cesar/.local/lib/python3.7/site-packages/git/util.py", line 867, in __contains__
    rval = list.__contains__(self, attr)
  File "/home/cesar/.local/lib/python3.7/site-packages/git/remote.py", line 451, in __eq__
    return self.name == other.name
AttributeError: 'str' object has no attribute 'name'
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,5 +32,6 @@ Contributors are:
 -A. Jesse Jiryu Davis <jesse _at_ emptysquare.net>
 -Steven Whitman <ninloot _at_ gmail.com>
 -Stefan Stancu <stefan.stancu _at_ gmail.com>
+-César Izurieta <cesar _at_ caih.org>
 
 Portions derived from other open source works and are clearly marked.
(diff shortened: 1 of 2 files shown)


cesarizu:
I just tried that same test on master and I still see the failure. Also, it doesn't touch the `IterableList#__contains__` function (modified in this PR) at all, so it seems unrelated. Or am I missing something here?

cesarizu:
In any case, I've just rebased the branch and it's all green now.

Byron:
Thanks a bunch!

Issue #871, opened 2019-05-06
Reading progress of Git operation using GitPython stuck (or not printing)


t89:
I'm trying to access the progress of time consuming Git operations using GitPython. I tried the sample solution taken from the official documentation, and also tried passing in a method following the exact signature of the update method below. Everytime I call `fetch()`, `push()`, `pull()` **with** the parameter `progress=<anything>`, the programm is stuck and the `update` method does not get called. If I call those operations without setting the `progress` parameter, it works flawlessly.

- `$ git --version` is 2.21.0
- Calling `sys.stdout.flush()` after `print()` does not help either
- I use `assert` to assure my repo objects are available and in the expected state
- `ProgressPrinter()` yields not `None`
- I tried calling the functions from the main thread and multithreaded
- I took a look at the [implementation (line 350)][1] of `RemoteProgress` and also the [implementation (line 815)][2] of `push()` and do not see a reason, why it would not continue execution
- I found out, that when I assign my `ProgressPrinter` instance and pass the assigned variable, the programm is not stuck anymore. Yet the `update()` method does not get called and no progress is printed:

```python
# Not stuck anymore, yet no progress
pp = ProgressPrinter()
fetch_info = origin.fetch(progress=pp)
```
___
Core of my implementation:

```python
from git import RemoteProgress

class ProgressPrinter(RemoteProgress):
    def update(self,
               op_code,
               cur_count,
               max_count=None,
               message=''):
        print("Is this even called?")
```
And later on:
```python
origin = repo.remotes.origin
assert origin.exists()
fetch_info = origin.fetch(progress=ProgressPrinter())

```

Any recommendations on how to investigate this problem furthermore? I've been debugging this for several days now and feel like I am missing something. 

  [1]: https://github.com/gitpython-developers/GitPython/blob/52ee33ac9b234c7501d97b4c2bf2e2035c5ec1fa/git/util.py#L350
  [2]: https://github.com/gitpython-developers/GitPython/blob/52ee33ac9b234c7501d97b4c2bf2e2035c5ec1fa/git/remote.py#L815

jeking3:
Try an older version of git (v2.20.0) as in v2.21.0 of git they changed the output strings to allow for localization, and that's caused at least one bug, let's see if it caused others...

t89:
I switched to both the default macOS Git `2.14.ish` as well as `2.20.1`. Both versions are not getting stuck anymore. Yet there is still no progress being printed.

[...]

PR #872, opened 2019-05-07
Implement update call when the object is "up to date" #871

t89:
In the current state the `update()` method of `RemoteProgress` is not called, if the progress terminates with ** = [up to date]**.

With this implementation, the method gets called once, containing all useful pieces of information. By passing `current_count = 1` and `max_count = 1` we assure that any percentage calculation terminates with **100%**.

**TODO**:
- Please check, if the `op_code = 0` makes sense in this case
- I opted for a more future proof, yet slower solution in the `elif`:
  - to trade durability for performance, use `.startswith(' = [up to date]')`

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -390,6 +390,21 @@ class RemoteProgress(object):
         if len(self.error_lines) > 0 or self._cur_line.startswith(('error:', 'fatal:')):
             self.error_lines.append(self._cur_line)
             return []
+        elif 'up to date' in self._cur_line:
+            # Checking this way instead of startswith, because debugging for
+            # startswith(' = [up to date]') is going to be a major pain if just
+            # a single space or bracket changes.
+
+            # Strip the initial ' = [up to date]' from the line
+            message_string = line.split('date]', 1)[-1]
+
+            # Trim whitespace
+            message_string = ' '.join(message_string.split())
+
+            self.update(0,
+                        1,
+                        1,
+                        message_string)
 
         sub_lines = line.split('\r')
         failed_lines = []


jeking3:
I am able to do a clean "git clone" followed by "tox" and it tests py27, py36, py37 locally and successfully.

t89:
I ran `$ git pull main master`, installed the (test-)requirements.txt using `pip install -r` and ran `tox` / `nosetests` both terminate with the same result:

> FAILED (SKIP=41, errors=10, failures=3)

I can reproduce this two machines. Suggestions?

t89:
I also investigated [your testfile](https://github.com/gitpython-developers/GitPython/blob/11fd713f77bb0bc817ff3c17215fd7961c025d7e/git/test/test_remote.py#L65) regarding the updated method and I know too little about GitPython's internals to write the test. Like I mentioned in my initial pull-request, I kindly ask the maintainer of the repo to insert the correct `op_code`, etc. pp.

t89:
Any feedback on this topic? _**"It works on my machine"**_ isn't really helpful at all. A part of my current project relies on this functionality and I would like to put in the time to make this pull-request easily mergeable for you.

jeking3:
It may be that you have a very old or very new version of git?  What does git --version say?

Byron:
Hi all, and sorry for the late reply. May I also suggest to have a look at
the travis file, as it performs a few commands that prepare the local clone
to be sufficient to run the tests. Yes, the tests are not isolated and
brutally rely on the local clone for read-only operations.
At this point I am not sure anymore if my machine still qualifies to run
them correctly, however, travis should still be green and can be triggered
with every push.

jeking3:
You may also be running into a bug in the unit test framework - I don't think it can handle anything other than the master branch without errors.  I haven't reported it yet (I actually forgot I ran into it months ago), but if you tried to run tox on a branch other than master, that might be why.  Perhaps the previous comment has guidance to work around that?

t89:
Wanted to let you know that I still intend to fix this issue, at the moment I don't have the sparetime thoug. Current workaround is freezing the GitPython dependency oin the current version and overriding the method with my own.

I'll let you know, once I come back to this.

Byron:

PR #873, opened 2019-05-07
BF: install_requires not just requires for setup() call

yarikoptic:
Originally detected while running DataLad tests on CRON  https://github.com/datalad/datalad/issues/3395 although mystery remains why not earlier.

To say the truth I am not fan of the approach (defer to reading requirements.txt) introduced in ce21f63 but I guess it is a separate issue


yarikoptic:
I think this one is obvious so I will just merge it

PR #875, opened 2019-05-19
Handling of relative submodule paths

PhilLab:
This MR is merely a start for a discussion about how to solve #730 

Is incomplete, I still get errors


Byron:
Thanks a lot for your initiative!
Would it help to add a test that represents the case to be fixed?

Byron:

Issue #876, opened 2019-05-21
Error "BadName: Ref 'HEAD' did not resolve to an object" encountered when trying to push repository containing deep submodule nesting


s4v4g3:
On Windows, If you try to push a repository that contains a deep nesting of submodules you can get a "BadName: Ref 'HEAD' did not resolve to an object" when GitPython tries to assemble the path that contains the git dir for the submodule.

A concrete example is the submodule path 
`C:\git\vsm-utilities\vsm-helper\csdieinfo\host_py_rm_pdb_internal\host_py_rm_pdb_internal_efr32x21x000f1024xm32_secpu`

that has a `.git` file containing

`gitdir: ../../../../.git/modules/vsm-helper/modules/csdieinfo/modules/host_py_rm_pdb_internal/modules/host_py_rm_pdb_internal_efr32x21x000f1024xm32_secpu`

In find_submodule_git_dir() in fun.py, the path to the git dir is assembled by using os.path.join, which results in a very long path.   In fact, os.path.isdir() returns False for this long path, even though the path length in absolute terms is under the size limit.  But because the path contains relative designators, the path length is too long for os.path.isdir() to handle and thus it returns False.

I found that calling os.path.normpath() on the long path would normalize the path to return a path that os.path.isdir() recognizes as a real directory and would fix the problem.

Byron:
Thanks for the information! I hope there is a chance for a PR to be crafted, maybe that way the fix can get merged upstream.

Byron:
Closed by #877

PR #877, opened 2019-05-21
Fix for issue #876

s4v4g3:
Normalize path after joining submodule path and the relative path to the git dir, to eliminate path length errors on Windows


jeking3:
There was an environmental error on the Windows python 2.7 job.  Please amend --no-edit and force push your PR to kick another build.  I would do it myself but only the owner of the AppVeyor account can do that.

Issue #332, opened 2015-07-30
Encoding problem


StyXman:
I'm not sure this is a proper way to use `TagReferences`, but it's definitely unexpected. This time I'm using `GitPython` installed by pypi.

I have this nice tag:

```python
In [8]: tag
Out[8]: <git.TagReference "refs/tags/PROMOTED_1501131729_MKT15_01_12_QU_1">
```

I can get a lot of info out of it:

```python
In [9]: tag.object.hexsha
Out[9]: u'dca63c5c7e6aab3cd4934e60230ec3419ab87071'

In [12]: tag.name
Out[12]: 'PROMOTED_1501131729_MKT15_01_12_QU_1'

In [13]: tag.object
Out[13]: <git.TagObject "dca63c5c7e6aab3cd4934e60230ec3419ab87071">

In [14]: tag.ref
TypeError: PROMOTED_1501131729_MKT15_01_12_QU_1 is a detached symbolic reference as it points to 'dca63c5c7e6aab3cd4934e60230ec3419ab87071'
```

But this fails:

```python
In [15]: tag.commit
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-15-2431a6e80cf9> in <module>()
----> 1 tag.commit

/home/mdione/local/lib/python2.7/site-packages/git/refs/tag.pyc in commit(self)
     29         elif obj.type == "tag":
     30             # it is a tag object which carries the commit as an object - we can point to anything
---> 31             return obj.object
     32         else:
     33             raise ValueError("Tag %s points to a Blob or Tree - have never seen that before" % self)

/home/mdione/local/lib/python2.7/site-packages/gitdb/util.pyc in __getattr__(self, attr)
--> 237         self._set_cache_(attr)
    238         # will raise in case the cache was not created
    239         return object.__getattribute__(self, attr)

/home/mdione/local/lib/python2.7/site-packages/git/objects/tag.pyc in _set_cache_(self, attr)
     54         if attr in TagObject.__slots__:
     55             ostream = self.repo.odb.stream(self.binsha)
---> 56             lines = ostream.read().decode(defenc).splitlines()
     57
     58             obj, hexsha = lines[0].split(" ")       # object <hexsha>

/usr/lib/python2.7/encodings/utf_8.pyc in decode(input, errors)
     14
     15 def decode(input, errors='strict'):
---> 16     return codecs.utf_8_decode(input, errors, True)
     17
[...]

Issue #237, opened 2015-01-19
Exception if GIT_COMMITTER_NAME contains UTF-8 encoded Umlaut


guettli:
If I run the python script using gitpython with this env var:

export GIT_COMMITTER_NAME="Thomas Müller"

I get this exception:

```

Error
Traceback (most recent call last):
  File "/home/foo_eins_d/src/scmtools/scm/tests/test_git.py", line 25, in test_git_status
    repo.index.commit('initial commit')
  File "/home/foo_eins_d/local/lib/python2.7/site-packages/git/index/base.py", line 900, in commit
    head, author=author, committer=committer)
  File "/home/foo_eins_d/local/lib/python2.7/site-packages/git/objects/commit.py", line 349, in create_from_tree
    new_commit._serialize(stream)
  File "/home/foo_eins_d/local/lib/python2.7/site-packages/git/objects/commit.py", line 392, in _serialize
    altz_to_utctz_str(self.committer_tz_offset))).encode(self.encoding))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 18: ordinal not in range(128)
```

Version: GitPython==0.3.5

Byron:
This is an interesting one ! git-python does handle non-ascii encodings, and defaults to utf-8.
In your case, there either is a non-obvious implicit conversion, or `self.encoding` ended up being 'ascii'.

I will review how encodings are handled to make sure something like this can't easily happen.

In the meanwhile, can you post the test in `test_git.py` here to make sure I can reproduce the issue ?

Byron:
You can watch the archived development stream [on youtube](http://youtu.be/pr6Qoaz2eE4)

guettli:
thank you very much for the fast fix.

saheel1115:
@Byron A similar issue occurs in GitPython-1.0.1 (tested with Python 2.7 and 3.4) when I try to do Repo.git.blame() and the blame output contains some funny characters.  Is this issue fixed somewhere upstream, so that I use a later version?

Byron:
@saheel1115 `Repo.git.blame()` executes the git command and attempts to interpret the returned bytes as UTF-8 string. If that is not expected to work, you can keep the bytes unchanged like so. `Repo.git.blame(stdout_as_string=False)`.

saheel1115:
@Byron Cool, sounds perfect. Thanks!

jrydberg:
[...]

PR #893, opened 2019-07-15
fix encoding problems on windows with not using default encoding but …

brizjin:
Solves encoding problem on windows with using fetch, pull, push commands. It similar to issues: [332](https://github.com/gitpython-developers/GitPython/issues/332), [237](https://github.com/gitpython-developers/GitPython/issues/237), explanation of problem I found [here](https://stackoverflow.com/questions/33283603/python-popen-communicate-str-encodeencoding-utf-8-errors-ignore-cr) in accepted answer. Also was encoding problem with calling repo.branches, quite simillar, file open in default locale encoding (cp-1251) for me, and getting branches names was broken.

### DIFF
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -90,7 +90,7 @@ class SymbolicReference(object):
         """Returns an iterator yielding pairs of sha1/path pairs (as bytes) for the corresponding refs.
         :note: The packed refs file will be kept open as long as we iterate"""
         try:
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for your efforts! 
Unfortunately there is [some breakage](https://travis-ci.org/gitpython-developers/GitPython/jobs/558977661#L715), maybe you can have a look.
Thank you

Byron:

Issue #719, opened 2018-01-29
Support additional worktrees


hilkoc:
As described in this issue: #344 . 
Since it's now almost 3 years later, it would be great if support for additional worktrees can be added.

KellyDeveloped:
There isn't a wrapper for worktrees yet, but I found using Repo.git.worktree("add", folder, branch) will load the worktree for you.

From there you can navigate into the worktree and use it like you world any other worktree. 

To unload the worktree, simply remove the folder and then call Repo.git.worktree("prune").

This works with the latest version.

Full example:

import os
import git

os.mkdir(<desired-folder>)

repo = git.Repo()
print (repo) # You'll see the original branch path that you're on here

repo.git.worktree("add", <folder>, <branch>)

test_repo = git.Repo(<folder>)
print (test_repo) # Now you'll see that with this new git.Repo object that it's referencing your worktree
print (repo) # Just to confirm that this is pointing to the original location

# Run a simple command to make sure everythings working - might be better to use the wrapper library to confirm though
print (repo.git.status())
print (test_repo.git.status())

hilkoc:
Thanks for this.
I managed to get it working. The main problem was this:
When you're in an additional worktree,
```
remote = repo.remote(name='origin')
remote.push()
```
does not work.

So instead, all the remote commands need to be called with repo.git.push() or repo.git.pull() etc.

enigma9119:
@KellyDeveloped  @hilkoc 

I am curious how you got this working. The example where you used Repo.git.worktree("add", folder, branch) and then initialized with git.Repo(<folder>) doesn't seem to work. git.Repo() always throws an InvalidGitRepositoryError when the folder path is a worktree.

My main objective is to get all the file names from the previous commit, which I can do using "repo.git.diff('HEAD~1', name_only=True)". This works with git clones, but with worktrees, git.Repo() throws an InvalidGitRepositoryError.

Do you know how I could get this working? or is there another way to get file names from git commits?

KellyDeveloped:
> 
> 
> @KellyDeveloped @hilkoc
> 
> I am curious how you got this working. The example where you used Repo.git.worktree("add", folder, branch) and then initialized with git.Repo() doesn't seem to work. git.Repo() always throws an InvalidGitRepositoryError when the folder path is a worktree.
> 
[...]

PR #894, opened 2019-07-16
use git rev-parse to look for config file

bdauvergne:
It fixes config file finding when in a separate worktree.

If fixes the issue with worktree reported in https://github.com/gitpython-developers/GitPython/issues/719#issuecomment-362232645

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -66,7 +66,7 @@ class Repo(object):
     'git_dir' is the .git repository directory, which is always set."""
     DAEMON_EXPORT_FILE = 'git-daemon-export-ok'
 
-    git = None  # Must exist, or  __del__  will fail in case we raise on `__init__()`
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for the contribution - I would love to merge, but can't as long as there is test breakage. Maybe it's an easy fix.

bdauvergne:
I tried a work-around but i'm not sure of it, there is some circular dependency between the search got the .git directory in Repo and the initialization of the self.git command wrapper (the git command should be able to find the .git itself, but the late initialization has maybe some use).

bdauvergne:
And it works (I don't know the problem with appveyor, I do not think it's related to my patch).

Byron:
Thanks so much! I am glad you could find the time.
In the next days, GitPython 3 will be released and include the fix.

PR #899, opened 2019-07-23
Fix Git.transform_kwarg

FrNecas:
While working on our project I found that `old_repo.git.rev_list('HEAD', max_parents=0)` started to work differently after the new release of GitPython and our tests started to fail. After some debugging I found the problem. Kwargs were not transformed correctly if a value was set to 0 due to wrong if condition:

```
>>> 0 not in (False, None)
False
>>> 0 is not False and 0 is not None
True
```

The commit also contains a test for the edge case.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -893,7 +893,7 @@ class Git(LazyMixin):
         else:
             if value is True:
                 return ["--%s" % dashify(name)]
-            elif value not in (False, None):
+            elif value is not False and value is not None:
                 return ["--%s=%s" % (dashify(name), value)]
         return []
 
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot! Fantastic test and fix! I hope putting it into v3.0.0 which drops python 2 as well is working for you.

FrNecas:
Thanks for merging. We will use my branch as a dependency until v3.0.0 is released on PyPI to have it working correctly for now and then we will switch back to the PyPI release.

kangp3:
Would it be possible to backport this fix to `v2.X` as well? We are relying on this package and unfortunately haven't dropped Python 2 support just yet.

Byron:
@kangp3 Thus far, maintaining two versions wasn't planned, but I will see if I can get a one-off release done.

PR #902, opened 2019-08-05
Correcting a file name

Arthur-Milchior:
There was an inconsistency. However, I don't know whether the file name or the text should be corrected.


Arthur-Milchior:
Hi,

Sorry, I wanted to do a pull request with a single edition. Pushing to master also told you that I wanted to push my last change, which was not the case. This is the reason for this force-pushed. That's not beautiful, but I hope, it should not appear in git tree.

Byron:
Thanks a lot! Every fix is appreciated :). No worries about force-pushing.

PR #903, opened 2019-08-05
Method stating which commit is being played during an halted rebase

Arthur-Milchior:
This will be useful to me at least. This way, I know that I can tell
my script to omit some specific commits. If you accept to merge it, I
may also do similar method for merges and cherry pick.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,5 +33,6 @@ Contributors are:
 -Steven Whitman <ninloot _at_ gmail.com>
 -Stefan Stancu <stefan.stancu _at_ gmail.com>
 -César Izurieta <cesar _at_ caih.org>
+-Arthur Milchior <arthur _at_ milchior.fr>
 
 Portions derived from other open source works and are clearly marked.
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -1061,3 +1061,14 @@ class Repo(object):
 
     def __repr__(self):
         return '<git.Repo "%s">' % self.git_dir
+
+    def currentlyRebasingOn(self):
+        """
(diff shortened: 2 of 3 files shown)


Byron-TW:
Thanks a lot, good idea! Great you took the extra time to make a contribution.

I would happily merge, but believe it needs some work before that could happen.

* camel case method name should be snake case
* Heads or commits have native types, do you think it would be more convenient to return such object instead of string?
* please add tests for both code paths

I understand that this would mean quite some extra work, and hope you can make it work.

Besides, I am happily accepting additional quality of life methods if they meet the quality requirements.

Thanks for your consideration:)

Arthur-Milchior:
What do you mean by «both code paths» ? I added a single function, and already added tests for them
I don't know what is a «quality of life method», sorry :s

Arthur-Milchior:
The other two requested change are already done

Byron:
Thanks for the changes. The tests would have to be added in [`test_repo.py`](https://github.com/gitpython-developers/GitPython/blob/master/git/test/test_repo.py#L1), however, I can't see any change to a file related to tests. `Both code paths` refers to both the success case, and the one where there is no active rebase.

Screenshot 2019-08-12 at 11 01 11

My remark about 'quality of life methods' refers to your comment about adding more methods like this, which I assume are to help you work with GitPython more easily, thus improving your 'quality of life', metaphorically speaking :D.

Arthur-Milchior:
I added the test in a new file, and forgot to add the file to git. My bad :sa: 
Now the test is in test_repo, as you asked for.

Byron:
Thanks a lot for your contribution! That does the trick!

PR #922, opened 2019-09-16
Fix UnicodeDecodeError in gpgsig

### LINKED ISSUE #915: UnicodeDecodeError in gpgsig
Hi @Byron !
So, someone find this bug using my tool, and it's propagated to GitPython. The problem is in the decoding of a non-ascii character.

repo: https://github.com/gentoo/gentoo
commit: `13e644bb36a0b1f3ef0c2091ab648978d18f369d`

code:
```
from git import Repo, Commit

gr = Repo('/tmp/gentoo')
c = gr.commit('13e644bb36a0b1f3ef0c2091ab648978d18f369d')

print(c.authored_date)
``` 
This returns:
```
  File "/Users/dspadini/Documents/pydriller/tmp.py", line 341, in <module>
    print(c.authored_date)
  File "/Users/dspadini/Documents/pydriller/venv/lib/python3.7/site-packages/gitdb/util.py", line 253, in __getattr__
    self._set_cache_(attr)
  File "/Users/dspadini/Documents/pydriller/venv/lib/python3.7/site-packages/git/objects/commit.py", line 144, in _set_cache_
    self._deserialize(BytesIO(stream.read()))
  File "/Users/dspadini/Documents/pydriller/venv/lib/python3.7/site-packages/git/objects/commit.py", line 502, in _deserialize
    self.gpgsig = sig.rstrip(b"\n").decode('ascii')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 75: ordinal not in range(128)
```
The problem is in [this line](https://github.com/gitpython-developers/GitPython/blob/master/git/objects/commit.py#L501). The line to be decoded is the following:
```
b'-----BEGIN PGP SIGNATURE-----\nVersion: GnuPG v2.1\nComment: Signed-off-by: J\xc3\xb6rg Bornkessel <hd_brummy@gentoo.org>\n\n.........\n-----END PGP SIGNATURE-----\n'
```
As you can see, at the beginning we have `J\xc3\xb6rg`. This fails the decoding.

So, I tried to change `.decode('ascii')` to `.decode('UTF-8')` and **it works**. 
Also, changing `.decode('ascii')` to `.decode('ascii', 'ignore')` **works**.

However, I am not sure whether I should do it. Why is `ascii` in the first place (instead of UTF-8)?
Are we gonna break tests with this change?

ishepard:
Fix #915. Originally opened in ishepard/pydriller#58.

It turns out that `commit.py` already has a `default_encoding` field set to 'UTF-8'. This was probably a leftover from previous versions.

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -484,7 +484,8 @@ class Commit(base.Object, Iterable, Diffable, Traversable, Serializable):
         buf = enc.strip()
         while buf:
             if buf[0:10] == b"encoding ":
-                self.encoding = buf[buf.find(' ') + 1:].decode('ascii')
+                self.encoding = buf[buf.find(' ') + 1:].decode(
+                    self.encoding, 'ignore')
             elif buf[0:7] == b"gpgsig ":
                 sig = buf[buf.find(b' ') + 1:] + b"\n"
                 is_next_header = False
@@ -498,7 +499,7 @@ class Commit(base.Object, Iterable, Diffable, Traversable, Serializable):
                         break
                     sig += sigbuf[1:]
                 # end read all signature
-                self.gpgsig = sig.rstrip(b"\n").decode('ascii')
+                self.gpgsig = sig.rstrip(b"\n").decode(self.encoding, 'ignore')
                 if is_next_header:
                     continue
             buf = readline().strip()


Byron:
Thanks a lot for your contribution, fantastic:)!

PR #926, opened 2019-09-28
Remove control character stripping.

uri-canva:
The logic to strip away ansi control sequences only works if they're at the end of the line, if the control sequences are at the beginning of the line the whole line will be stripped away, and all progress information will be lost.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -364,6 +364,7 @@ class RemoteProgress(object):
                  '_seen_ops',
                  'error_lines',  # Lines that started with 'error:' or 'fatal:'.
                  'other_lines')  # Lines not denoting progress (i.e.g. push-infos).
+    re_ansi_escape = re.compile(r'\x1B[@-_][0-?]*[ -/]*[@-~]')
     re_op_absolute = re.compile(r"(remote: )?([\w\s]+):\s+()(\d+)()(.*)")
     re_op_relative = re.compile(r"(remote: )?([\w\s]+):\s+(\d+)% \((\d+)/(\d+)\)(.*)")
 
@@ -392,17 +393,8 @@ class RemoteProgress(object):
 
         # find escape characters and cut them away - regex will not work with
         # them as they are non-ascii. As git might expect a tty, it will send them
-        last_valid_index = None
-        for i, c in enumerate(reversed(line)):
-            if ord(c) < 32:
-                # its a slice index
-                last_valid_index = -i - 1
-            # END character was non-ascii
-        # END for each character in line
-        if last_valid_index is not None:
-            line = line[:last_valid_index]
-        # END cut away invalid part
-        line = line.rstrip()
+        line = self.re_ansi_escape.sub('', line)
+        line.strip()
 
         cur_count, max_count = None, None
         match = self.re_op_relative.match(line)
--- a/git/util.py
+++ b/git/util.py
@@ -394,7 +394,7 @@ class RemoteProgress(object):
         # find escape characters and cut them away - regex will not work with
         # them as they are non-ascii. As git might expect a tty, it will send them
         line = self.re_ansi_escape.sub('', line)
-        line.strip()
+        line = line.strip()
 
         cur_count, max_count = None, None
         match = self.re_op_relative.match(line)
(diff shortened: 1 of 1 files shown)


git/util.py
```diff
@@ -395,17 +396,8 @@ def _parse_progress_line(self, line):
...
-                    # its a slice index
-                    last_valid_index = -i - 1
-                # END character was non-ascii
-            # END for each character in sline
-            if last_valid_index is not None:
-                sline = sline[:last_valid_index]
-            # END cut away invalid part
-            sline = sline.rstrip()
+            sline = self.re_ansi_escape.sub('', sline)
+            sline.strip()
```
Byron: `strip()` will not mutate the string. Thus it should be `sline = sline.strip()`.
If it is important, do you think it's possible to have a test for it? Otherwise, maybe that line is not needed at all.

Byron: I just saw that the line stripping was done previously as well, let's just re-add it then as in `sline = sline.rstrip()`. Apparently there is no test that would need it, and it's tech-debt I wouldn't ask you to pay either 😅.

uri-canva: Absolutely, good catch!

uri-canva:
Hmm, tests are failing, it is a bit different in behaviour, before if there was `abc<control sequence>def` the resulting string would be `abc`, now it's `abcdef`. Will have to test it further.

uri-canva:
Seems to have been a bad rebase on my part, now that it's rebased on master after the other PRs were merged it's working fine.

git/util.py
```diff
@@ -392,17 +393,8 @@ def _parse_progress_line(self, line):
...
-                # its a slice index
-                last_valid_index = -i - 1
-            # END character was non-ascii
-        # END for each character in line
-        if last_valid_index is not None:
-            line = line[:last_valid_index]
-        # END cut away invalid part
-        line = line.rstrip()
+        line = self.re_ansi_escape.sub('', line)
+        line.strip()
```
Byron: Thank you! Once `line = line.strip()` is back I can merge it.

uri-canva: Sorry, I must have dropped the change during the rebasing. Because the tests only work on `master` I lose track of where the commits are sometimes. 😆

uri-canva:
Ah it's the stripping that breaks the tests.

uri-canva:
Changed it back to `rstrip` instead. I don't really understand what is going on, it works on my machine both in the macos host and the linux docker image with `strip`, so I couldn't debug it. I guess leaving it as it is is the safest option.

Byron:
Thanks so much for your continued efforts, it's much appreciated :)!

uri-canva:
Thank you for the wonderful library!

Byron:
About 10 years later, I am not quite sure if it's wonderful anymore ;).

Somehow now [it breaks on master](https://travis-ci.org/gitpython-developers/GitPython/builds/591318027) in a spot that seems unrelated. Fascinating. Could it be the difference between `strip()` and `rstrip()`?

uri-canva:
It does seem completely unrelated: it doesn't have anything to do with progress. I understand if it would break tests related to progress. But anyway definitely revert it if it's the problem!

Byron:
It looks like the `rstrip()` call was the problem. Probably it should worry me that I have no idea how this could affect that particular test :/.
The conservative voice tells me to roll back the commit entirely, but prefer to trust the respective tests and rather add additional ones as needed.

uri-canva:
Yeah I think rolling back makes sense, then figure out how to repro the failure on the PR with a new test, then fix it. Without a new test that fails on the PR there's no way to tell if the problem will reappear after the PR is merged.

Byron:
Thanks, @uri-canva , the respective commit was reverted.

I agree, probably it would be beneficial to have a test for the initial motivation of this PR, i.e. the code removing too much of the line if the escape sequence appears in an unexpected spot, and then provide a fix.

uri-canva:
Oh yes, that too. I meant it mostly the other way around: none of the tests failed on this PR, but then some tests failed on master, so it would be good for that hole to be filled.

PR #927, opened 2019-09-28
Take advantage of universal newlines.

uri-canva:
Universal newlines means that the streams will split on `\r` just like any other kind of newline, so there is no need to handle them differently.


uri-canva:
DeepSource failed because I changed some lines with pre-existing issues.

Byron:
What a catch! Thanks a lot!
Quite amazing that despite me having implemented complex 'sub-line' handling, there is no test that actually verifies it 😅

PR #928, opened 2019-09-28
Remove assert that can fail erroneously.

uri-canva:
`git` will sometimes print out empty progress lines.


Byron:
Thanks a lot! I also wonder why these empty lines would be an issue, and since there is no test for that, the line might as well be removed.

PR #929, opened 2019-09-28
Parse rejected deletes.

uri-canva:
Successful branch deletions look like

```
-	:refs/heads/test-delete	[deleted]
```

however rejected branch deletions look like

```
!	(delete):refs/heads/test-delete	[rejected] (stale info)
```

for some reason. Parse those as well.


Byron:
Fantastic! Thanks so much!

PR #936, opened 2019-10-13
Allow single item for index.addremovemove

### LINKED ISSUE #813: repo.index.add -> check if input is a list

MarcelBeining:
reopens #840 
finally closes #813


Byron:
Looks great, thanks a lot for your contribution!

PR #940, opened 2019-10-15
Fix if else

pratikanurag:


Byron:
Thanks a lot for all your contributions! They are much appreciated!

Issue #919, opened 2019-09-11
Fix some (critical) issues discovered by deepsource


Byron:
https://deepsource.io/gh/gitpython-developers/GitPython

imkaka:
Hey @Byron,  I will take this. :)

Byron:
@imkaka Thanks a lot for your help.

When it's about making decisions, it is entirely up to you to make the final call. It should be possible to silence deepsource using an annotation, or to disable entire classes of errors/warnings.

Regarding _Q1_, also I don't know why the version is made explicit there. Maybe this is how they avoid drastic changes in behaviour? A quick check revealed that there is version 3.0 already, maybe consider upgrading before you begin.

Generally, you can feel free to make fixes the way you find appropriate :).

Good luck!

imkaka:
Do we only support Python 3 Now? then we should change the `runtime_version='3.x.x'`?

Byron:
That's correct. Python 2 support was dropped not too long ago.

imkaka:
We have more issues after changing the version, I will try to fix this incrementally :)

imkaka:
In every file there are various *Unused Imports* as well as *Wild Card Imports*, Can anyone suggest do we use transitional imports?

Example: 
* `a.py`
```py
from gitdb import X, Y
```

* `b.py`
```py
from a import X
```

What to do with Unused Imports?

jaipradeesh:
This might help: https://deepsource.io/docs/configuration/skip-cq.html

imkaka:
I am following the same doc, but there is no option to silence deepsource globally, will work on it. :100:

imkaka:
To silent globally, we have to log in to `deepsource` and select `actions` for each issue we want to disable as described in [this]( https://deepsource.io/blog/releases-issue-actions/).

I don't see the `action` button as I don't credential maybe.

Byron:
Thanks for the hints! It would be great to be able to ignore things globally using a file instead, keeping all relevant configuration with the project code.

For now, I have disabled all import related rules, and this is how it looks for me:
<img width="366" alt="Screenshot 2019-10-28 at 08 48 54" src="https://user-images.githubusercontent.com/63622/67660838-fb9b9e80-f95f-11e9-9661-5eb7f032519e.png">

jaipradeesh:
Thanks for the input @Byron. Will keep you posted.

PR #942, opened 2019-10-16
Update .deepsource.toml

imkaka:
Relates to #919


Byron:
Thanks a lot! I am curious to see how this changes the results of deepsource, and if it gets better or worse.

Issue #852, opened 2019-03-13
repo.index.diff("HEAD", create_patch=True) always empty


jadinm:
I am using GitPython v2.1.11 with Python 3.6. My git version is 2.17.1.

I see inconsistent behavior of the diff method of a repo index if create_patch option set to True.
I am running the following script in a git repository with staged and unstaged changes.

```python
from git import Repo

repo = Repo(".")

print(repo.index.diff("HEAD"))
print(repo.index.diff("HEAD", create_patch=True)) # Problem here

print(repo.index.diff(None))
print(repo.index.diff(None, create_patch=True))
```

But it prints:
```
[<git.diff.Diff object at 0x7fd54c328678>, <git.diff.Diff object at 0x7fd549e405e8>, <git.diff.Diff object at 0x7fd54c328558>]
[]
[<git.diff.Diff object at 0x7fd54c328678>]
[<git.diff.Diff object at 0x7fd54c328678>]
```

Somehow, the diff is empty if I set `create_patch` to True and the first argument to anything but `None`.
Do you know why it happens ?

jadinm:
I found out the problem.

When running ``repo.index.diff("HEAD", create_patch=True)``, the command used is:
```bash
$ git diff -R <reference-hash> --cached --abbrev=40 --full-index -M -p --no-color
```

And this produces:
```
diff --git b/other_file.txt a/other_file.txt
deleted file mode 100644
index d733472942281cd2f19f52ab4b5ddecb4151bb72..0000000000000000000000000000000000000000
--- b/other_file.txt
+++ /dev/null
@@ -1 +0,0 @@
-Other file
```

But the regex that is used for parsing (i.e., ``Diff.re_header``) expects ``a/other_file.txt`` to be before ``b/other_file.txt``.
The inversion is due to the "-R" flag in the command line.

-------------------
A temporary fix for my use of GitPython is to set R to True to prevent the reversal.

```python
print(repo.index.diff("HEAD", create_patch=True, R=True))
```

Byron:
Thanks a lot for sharing the issue, and for posting the workaround here as well!

can-taslicukur:
I am using version 3.1.42 and experiencing this issue currently.

```python
from git import Repo

repo = Repo(".")
print(repo.index.diff("HEAD"))
print(repo.index.diff("HEAD", create_patch=True))
```

prints

```python
[<git.diff.Diff object at 0x10cbb08b0>]
[]
```

PR #948, opened 2019-10-21
Fix repo.index.diff("HEAD", create_patch=True) always returning an empty list

thetwoj:
This PR resolves the issue reported in #852 

As @jadinm correctly identified in #852, this bug centers around the fact that the regex utilized by `_index_from_patch_format()` in `git/diff.py` always expects the first line of the diff to have the `a` file first, then `b`:
```
diff --git a/the_file.py b/the_file.py
```
When the `other` passed to `repo.index.diff(other)` is not `None` the `-R` argument is added to the git command, resulting in the positions of the `a` and `b` files being flipped in the output and causing the `_index_from_patch_format()` function invoked by the `create_patch` option to not find a regex match:
```
diff --git b/the_file.py a/the_file.py
```

This fix makes the regex slightly more permissive, allowing it to match either order of the `a` and `b` files in the output of `git diff`.

For test coverage of this change I added assertions to an existing test within `test_diff.py` that set up everything needed to reproduce this bug.

Massive props to @jadinm, their troubleshooting and feedback on the issue significantly shortened the investigation required to get this one sorted out 🎉

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -237,7 +237,7 @@ class Diff(object):
     # precompiled regex
     re_header = re.compile(br"""
                                 ^diff[ ]--git
-                                    [ ](?P<a_path_fallback>"?a/.+?"?)[ ](?P<b_path_fallback>"?b/.+?"?)\n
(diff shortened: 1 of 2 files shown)


thetwoj:
Thanks @Byron! I appreciate your willingness to entertain these PRs and how promptly you review them, it makes contributing fun :+1:

PR #952, opened 2019-10-23
Fix/deepsource issues

imkaka:
* [x] Covers all critical issues.
* [x] Some Major issues also.

I will continue with [this](#919 ) further.
* refers #919


git/refs/log.py
```diff
@@ -84,12 +84,12 @@ def message(self):
         return self[4]
 
     @classmethod
-    def new(cls, oldhexsha, newhexsha, actor, time, tz_offset, message):
+    def new(cls, oldhexsha, newhexsha, actor, time_, tz_offset, message):
```
Byron: Is changing the `time` parameter name truly required? It would break people using the _keyword_ way of calling the method. If possible, I would prefer to keep it as is and silence deep source in other ways.

git/refs/log.py
```diff
@@ -71,7 +71,7 @@ def actor(self):
         return self[2]
 
     @property
-    def time(self):
+    def time_(self):
```
Byron: Is changing the `time` property name truly required? It would break people using it. If possible, I would prefer to keep it as is and silence deep source in other ways.

git/refs/log.py
imkaka: Ok, I will revert it to make it silent,  It was not a critical issue.

git/refs/log.py
```diff
@@ -217,9 +217,9 @@ def entry_at(cls, filepath, index):
             the index is negative
         """
         fp = open(filepath, 'rb')
-        if index < 0:
-            return RefLogEntry.from_line(fp.readlines()[index].strip())
-        else:
+        with open(filepath, 'rb') as fp:
```
Byron: This looks like it will open the file twice, leaking one file handle in the process. The first `open` occours in the line above.

imkaka: Yeah, Thanks

PR #961, opened 2019-11-22
Avoids env var warning when path contains $/%; fix #832

### LINKED ISSUE #832: Use of environment vars in paths is deprecated, Warning in Python 3.7
```
warnings.warn("The use of environment variables in paths is deprecated" +
/usr/local/lib/python3.7/dist-packages/git/repo/base.py:127: UserWarning: The use of environment variables in paths is deprecated for security reasons and may be removed in the future!!
```

However, I'm passing absolute paths to the Repo class, this creates a lot of warning messages in the script output

rgkimball:
I ran into the same issue as @kamadoatfluid in #832, as I have a mounted disk path including a `$` (valid file & folder character in most operating systems, as is `%`). Since this is intended to catch environment variables per #662, I propose the use of regex instead to prevent irrelevant warnings from cluttering logs/console output. 

This PR supports all expansion conventions demonstrated in [`os.path.expandvars`](https://docs.python.org/3/library/os.path.html#os.path.expandvars), and also aligns with environment variable naming specifications. Illegal variables simply will cause errors and therefore don't present a security risk: word characters only, beginning with `a-z` or `_`.

Supported formats:
- `% APPDATA %` or `%APPDATA%`
- `$MYVAR`
- `${MYVAR}`

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -77,6 +77,7 @@ class Repo(object):
     re_whitespace = re.compile(r'\s+')
     re_hexsha_only = re.compile('^[0-9A-Fa-f]{40}$')
     re_hexsha_shortened = re.compile('^[0-9A-Fa-f]{4,40}$')
+    re_envvars = re.compile(r'(\$(\{\s?)?[a-zA-Z_]\w*(\}\s?)?|%\s?[a-zA-Z_]\w*\s?%)')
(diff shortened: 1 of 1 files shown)


Byron:
Sorry for the late response and thanks a lot for the improvement!

Issue #963, opened 2019-11-27
Using GitPython inside a pre-commit hook breaks (GIT_AUTHOR_DATE breaks)


dries007:
I suspect the culprit of the crash is the GIT_AUTHOR_DATE. If I remove this before starting my script it works. I already pass `expand_vars=False`, but that isn't passed far enough down to make a differance.

My current workaround is to nulke the enviroment variables when invoking my script by using `env -i <script>` from the commit hook.

My ideal solution would be a flag to disable the copy of the os environ on [git/cmd.py:699](https://github.com/gitpython-developers/GitPython/blob/master/git/cmd.py#L699). This will not only eliminate this issue, but also any potential other interferance (think GIT_DIR and friends).

Replicate via:

``` bash
export GIT_INDEX_FILE='.git/index'
export GIT_AUTHOR_DATE='@1574863356 +0100'
export GIT_EXEC_PATH='/usr/lib/git-core'
export GIT_AUTHOR_EMAIL='admin@dries007.net'
export GIT_PREFIX=''
export GIT_AUTHOR_NAME='Dries'

python3
```

```python
import tempfile

import git


def main():
    with tempfile.TemporaryDirectory() as folder:
        repo = git.Repo.init(folder, expand_vars=False)
        repo.index.commit("Init test repo with empty commit.")
        repo.head.reference = repo.create_head("master")


if __name__ == "__main__":
    main()
```

Thanks,
Dries

Byron:
Thanks for providing what’s needed to reproduce this issue, even though I didn’t thanks to writing this on an iPad. Something that would be nice to add to the issue is the stack trace that follows from executing the example.

Something I wonder is if you already verified that not copying the environment in https://github.com/gitpython-developers/GitPython/blob/master/git/cmd.py#L699 will fix the issue. If I recall correctly, child processes always inherit from their parent process environment, and the copy is meant to allow idempotent changes to the env dict.

dries007:
Just for your info: [if you use the env kwarg in Popen, the parent's env is not cloned](https://docs.python.org/3/library/subprocess.html#subprocess.run), but you're right, it doesn't actually fix the issue.

[...]

PR #965, opened 2019-12-08
Added parsing for '@1400000000 +0000' date format as used by git commit hooks

dries007:
The git pre-commit hook is run with the `GIT_AUTHOR_DATE` environment variable set.
The format used (on my machine anyway) is `@1574863356 +0100`, which is not handles correctly by [`parse_date`](https://github.com/gitpython-developers/GitPython/blob/master/git/objects/util.py#L150). 

This PR fixes that crash, but it doesn't quite fully resolve issue #963. This issue also contains steps to reproduce and a stacktrace.


Byron:
Apologies for the late response! 

PR #968, opened 2019-12-13
Fix Repo.__repr__ when subclassed

yaccz:


Byron:
Thanks for your contribution! It looks like travis failed, could you have a look?
Usually I would link the exact line of failure here, but it seems I can't access travis-ci right now.
My guess is that `git_dir` can't always be used safely.

yaccz:
@Byron fixed.

yaccz:
I do not know why AppVeyor is failing. It does not seem related to my changes.

Byron:
Thanks a lot, and sorry for the late reply. Unfortunately the release train just left the station, which means this fix will land in 3.0.8.

PR #974, opened 2020-01-22
Remove A from Dockerfile

Matt-Gleich:
Removed A from Dockerfile that I added accidentally. THIS WILL BREAK THE BUILD


Byron:
Good one! Reminds me of reviewing more carefully 😅

PR #988, opened 2020-02-24
Fix param format of Repo.commit

tkdchen:
Signed-off-by: Chenxiong Qi <qcxhome@gmail.com>

This fixes the format:

`
The Commit object for the specified revision :param rev: revision specifier, see git-rev-parse for viable options. :return: git.Commit
`


Byron:

Issue #993, opened 2020-03-03
GitPython 3.0.3 - 3.1.0 unable to get diffs from bare repo


romuaa:
GitPython 3.0.2 works properly.
GitPython 3.0.3 is unable to read the bare repo.
GitPython 3.0.4 and 3.0.5 are not able to get diffs
GitPython 3.0.6 is unable to read the bare repo.
GitPython 3.0.7 - 3.0.9, 3.1.0 are not able to get diffs

Test script and script outputs attached.
Test script is a Python 3 script named to diff_fail.script.txt because .py not supported as an upload.

[diff_fail.script.txt](https://github.com/gitpython-developers/GitPython/files/4280662/diff_fail.script.txt)

[gitpython_3.0.2.output.txt](https://github.com/gitpython-developers/GitPython/files/4280651/gitpython_3.0.2.output.txt)

[gitpython_3.1.0.output.txt](https://github.com/gitpython-developers/GitPython/files/4280654/gitpython_3.1.0.output.txt)

Byron:
Thanks for posting, and for providing the means to reproduce the issue.
It surprises me that GitPython seems to have regressed.

Which OS are you using, and which Git version? Sometimes this matters a lot.
Thanks

dongs0104:
👍 
I tested it on python:3.7.3-stretch docker, 3.1.0 version gitpython, git version is 2.11.0

romuaa:
Sorry for late reply. For me it occurs on macOS Mojave 10.14.6 with git version 2.25.0
Haven't tested it on our Linux servers since I caught it while developing.

Byron:
Thanks for the update. The question I can't really answer to myself is if the behaviour of GitPython 3.0.2 is desirable, as it seems to ignore the dangling ref or doesn't run into that issue for other reasons. What does git itself do?

Since the issue seems to come from the diff implementation, it must be something that changed there in more recent versions of GitPython. Maybe that can be tracked down to lead to an improved quality of life, especially under the assumption that git itself won't fail like that.

Byron:
Having just seen issue 999, it looks like we got the answer: Git itself will fail to diff if there is a dangling HEAD reference. Thus the behaviour of GitPython can be considered adequate, assuming git always knows what's best :D.

Thus I recommend closing this issue, what do you think, @romuaa ?

romuaa:
[...]

PR #999, opened 2020-03-18
#993 fix: wrong refs 'HEAD' exception

dongs0104:
When git HEAD refs is broken, it raise BadName exception.
so submodules logic which is using a rev_parse function to HEAD get Exception, so handle BadName to return empty list 

https://github.com/gitpython-developers/GitPython/commit/b303cb0c5995bf9c74db34a8082cdf5258c250fe
https://github.com/gitpython-developers/GitPython/blob/b303cb0c5995bf9c74db34a8082cdf5258c250fe/git/diff.py#L281-L287

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -20,7 +20,8 @@ from git.config import (
 from git.exc import (
     InvalidGitRepositoryError,
     NoSuchPathError,
-    RepositoryDirtyError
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for your contribution :)! It's much appreciated.

CC: #993 , where diff is broken due to a very similar issue.

PR #1005, opened 2020-04-08
Remove forced verbosity when fetching from a remote

ngcortes:
When fetching repos with many remotes, the verbose flag (ie. v=True) in the body of the git.remote.Remote.fetch method produces a lot of output that isn't necessarily needed.

Adding a "verbose" kwarg to the method signature would make verbosity optional.


Byron:
Thanks a lot! A small change for a big improvement - much appreciated :).

PR #1007, opened 2020-04-21
make iterator python3 compatible

grubberr:


git/cmd.py
```diff
@@ -498,7 +498,7 @@ def readlines(self, size=-1):
         def __iter__(self):
             return self
 
-        def next(self):
+        def __next__(self):
```
Byron, line 501: Would this be a breaking change if people used __next__ directly?

I am wondering if the old method name should be retained for safety, like so:

```python
        def next(self):
              return self.__next__()
```

Byron:

PR #1009, opened 2020-05-02
Submodule depth

Liambeguin:


Liambeguin:
Hi, 

I managed to test it locally but I'm having trouble with the test case...
this is how I tested it:

```sh
$ virtualenv -p python3 venv
$ source venv/bin/activate
$ pip install . 
$ rm -rf foo && python ./test.py 
```

`test.py` file:
```python
import git

repo = git.Repo.init('foo')

sm_name = 'test'
sm_depth = 1

sm = repo.create_submodule(sm_name, sm_name, url='https://github.com/gitpython-developers/GitPython.git', depth=sm_depth)
print(len(list(sm.module().iter_commits())), sm_depth)
assert len(list(sm.module().iter_commits())) == sm_depth
```

Byron:
Thanks a lot, especially for making it so easy to reproduce it is indeed working.
Since I also couldn't figure out why the test ignored the passed depth parameter, even though it did show up in the invoked git command, I marked this test as permanently skipped to document the issue. Maybe someone will find the answer one day :).

Liambeguin:
Great! Thanks for reviewing :)

PR #1028, opened 2020-07-10
move test and avoid packaging them

### LINKED ISSUE #1027: Don't install unittest code by default
When doing `python3 setup.py install` on a local checkout of the code, files from `git/test` will be installed into the standard site-packages directory.
This mean that theoretically even the requirements from `test-requirements.txt` are mandatory for the base package and secondly it isn't necessary as the chances that an user is actually using the test code are pretty low.
So overall it doesn't make much sense to me.

Is there any way to avoid having the test files installed via setuptools?

priv-kweihmann:
Move test files out of git module and into the root dir.
This enables it to ignore them while setuptools packaging and is resulting in a way smaller release package.

Closes #1027


Byron:
I am absolutely amazed and am super thankful for this PR and the effort that undoubtedly went into it.
It will have a massive impact to everyone, given that GitPython is downloaded a lot!

Will cut a new release right away!

Issue #1030, opened 2020-07-12
ModuleNotFoundError: No module named 'git.objects'


HappyEinara:
The release just now might have broken something.

Steps to reproduce:
- Fresh venv, python3.8
- pip install --upgrade gitpython
- python
then:
```
>>> import git
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vagrant/.virtualenvs/tmpgit/lib/python3.8/site-packages/git/__init__.py", line 41, in <module>
    from git.objects import *               # @NoMove @IgnorePep8
ModuleNotFoundError: No module named 'git.objects'
```

michalb-dsg:
I got the same error

Byron:
Definitely related to the changes coming from #1028 , could @priv-kweihmann have a look?
The tar archive definitely has the 'git/objects' directory, so I am a bit surprised it doesn't get installed.

As immediate action I am rolling the change back.

p0tr3c:
same here :-(

Byron:
v3.1.5 was released with the PR in question reverted. My apologies for the inconvenience!

I hope there is a way to test the usability of releases prior to making them, somehow, it's not the first time a release breaks even though it looks alright. Maybe something like this can be contributed. Personally I do consider setup.py and friends black magic that I cannot touch or tame, and must be relying on contributions to improve the situation. Thanks for your understanding and support.

priv-kweihmann:
> Definitely related to the changes coming from #1028 , could @priv-kweihmann have a look?
> The tar archive definitely has the 'git/objects' directory, so I am a bit surprised it doesn't get installed.
> 
> As immediate action I am rolling the change back.
I will - as this could take some time, it's really better to rollback for now. Will try to include some tests so that this won't happen again

Byron:
I am closing this issue as it was fixed with v3.1.5, which was released about 30 minutes after v3.1.4.
With the PR above we have brought back the reduced package size in v3.1.6, which is likely to be working as we are now testing for git being importable after having been installed (from source).

PR #1031, opened 2020-07-12
[RFC/WIP] move tests and avoid packaging them

priv-kweihmann:
This is the followup to #1028.

Move test files out of git module and into the root dir.
This enables it to ignore them while setuptools packaging and is resulting in a way smaller release package.

additionally a new test is added, which tests package installation, to avoid ModuleNotFound errors (e.g. #1030) in future

Please review and test locally if possible - that's why I marked it as RFC/WIP for now


Byron:
Thanks a lot for giving it another go, and for 'the test' which will probably save many people a lot of time in future by catching bad releases before they happen.

I will cut a new release, fingers crossed!

PS: This PR is probably worth ~15GB in saved transfer volume per month if [these stats](https://pypistats.org/packages/gitpython) can be trusted - amazing!

Issue #1032, opened 2020-07-13
Tutorial examples in documentation no longer displaying


kiancross:
https://github.com/gitpython-developers/GitPython/pull/1031 has broken the documentation examples at: https://gitpython.readthedocs.io/en/stable/tutorial.html

PR #1033, opened 2020-07-13
Fixed broken paths in tutorial documentation

kiancross:
PR for https://github.com/gitpython-developers/GitPython/issues/1032
Fix for bug introduced in https://github.com/gitpython-developers/GitPython/pull/1031.

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -10,14 +10,14 @@ GitPython Tutorial
 
 GitPython provides object model access to your git repository. This tutorial is composed of multiple sections, most of which explains a real-life usecase.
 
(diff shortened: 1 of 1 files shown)


kiancross:
Warnings should probably be treated as errors when building the documentation (including a non-existent file is a warning not an error). This would then have been caught by the GitHub action which builds the documentation. However, the other superfluous warnings being produced by Sphinx would need to be sorted first.

Byron:
Thanks a lot for helping to fix this one!
So many places where releases can go wrong, and thus far GitPython was just pretty lucky.

I agree, it would be great to have sphinx fail instead of ignoring warnings - and I wonder if one could keep only the warnings that are known to be errors, instead of having to fix all the other warnings, in case that get's us to having at least some test earlier.

On the other hand, maybe now we are stable once again so we are good for the years to come.

It seems like [latest](https://gitpython.readthedocs.io/en/latest/tutorial.html) is now back, but another release is required to fix stable.
Why not :D.

PR #1035, opened 2020-07-13
Treat warnings as errors when building documentation

kiancross:
This fixes all the current warnings when building the documentation and updates the `Makefile` to treat warnings as errors.

In reference to: https://github.com/gitpython-developers/GitPython/pull/1033

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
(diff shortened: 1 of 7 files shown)


Byron:
You are on fire! Thanks so much!

With this fix there will be no silent breakage of the documentation anymore, and of course, it has gotten a little better as many of these warnings seem like actual errors or mistakes.

PR #1038, opened 2020-07-17
Ensure only fully matching symrefs are deleted

westphahl:
Deleting a symbolic ref with e.g. the name 'refs/remotes/origin/mas'
would also delete 'refs/remotes/origin/master' if the ref had to be
deleted from the pack file.

In order to fix this the full ref is now checked for a match.

### DIFF
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -445,12 +445,14 @@ class SymbolicReference(object):
                     made_change = False
                     dropped_last_line = False
                     for line in reader:
+                        line = line.decode(defenc)
+                        _, _, line_ref = line.partition(' ')
+                        line_ref = line_ref.strip()
                         # keep line if it is a comment or if the ref to delete is not
                         # in the line
                         # If we deleted the last line and this one is a tag-reference object,
                         # we drop it as well
-                        line = line.decode(defenc)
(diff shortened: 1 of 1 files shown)


westphahl:
Any pointers on how to test this would be appreciated. The problem is that I need a repo with the remote still in the pack file to exercise the changed code path.

Byron:
My apologies for the late reply, I do try to prioritize PRs and mistook it as an issue.

That said, [the tests](https://github.com/gitpython-developers/GitPython/blob/master/test/test_refs.py#L30:L30) seems to succeed despite the changes, which might mean that the behaviour described here is not actually intended.
Maybe it would be possible to add a test which tries to reproduce the undesirable behaviour, which in turn is fixed with this patch.

Please let me know if you have any issues or lack the time to dive into it, as my disposition here is to merge anyway.

westphahl:
I don't know a way to create a git repo that has the remote still in the pack file, as that would be needed to reproduce the issue. We are using a patched version of GitPython for some time now and don't see the described behavior anymore.

In case you also don't know a way to setup a git repo with the remote in the pack file I'd be glad if you could merge it as is.

Byron:
Sometimes a little nudge is all it takes - let's merge and hope for the best.
Thanks again for taking this on, it's much appreciated!

Issue #1019, opened 2020-06-08
Different behavior between repo.remote().url and repo.remote().urls when dealing with UNC paths


HudsonMC16:
Not sure if this is specifically realted to UNC paths or not, but here is what I'm observing. 

Clone a repo from a networked server to your local machine. Then:

```
>>> from Pathlib import Path
>>> path_str = local_repo.remote('origin').url
>>> print(path_str)
'\\\\\\\\network\\\\path\\\\to\\\\remote\\\\repo'
>>> Path(path_str).exists()
False
>>> Path(path_str).resolve()
WindowsPath('C:/network/path/to/remote/repo')
```
whereas:
```
>>> from pathlib import Path
>>> path_str = next(local_repo.remote('origin').urls)
>>> print(path_str)
'\\\\network\\path\\to\\remote\\repo'
>>> Path(path_str).exists()
True
>>> Path(path_str).resolve()
WindowsPath('//network/path/to/remote/repo')
```
Thanks!

Byron:
Thanks a lot for posting this here!

I had a quick look and believe the reason is the following: When [asking for `urls`](https://github.com/gitpython-developers/GitPython/blob/24cd6dafc0008f155271f9462ae6ba6f0c0127a4/git/remote.py#L528-L534) git will be invoked to ask for its view of the world.

When [asking for `url`](https://github.com/gitpython-developers/GitPython/blob/24cd6dafc0008f155271f9462ae6ba6f0c0127a4/git/objects/submodule/base.py#L1119-L1121) it will use the potentially manipulated path that was used to instantiate the `Repo` instance.

The issue seems to be stemming from GitPython incorrectly transforming the UNC path, potentially treating it like some sort of Linux path.

HudsonMC16:
So seems it would make sense for each function to access the url in the same way. Do we have a preference? Should the `.url` code be fixed to quit mangling the paths, or be rewritten to invoke git, similar to `.urls`?

Byron:
I think it does make sense that their output is consistent, with `urls` being correct in this case.

Since calling implementing `url` like `next(urls)` is not the same both in terms of performance nor (probably) the result itself, I would clearly prefer `url` to be fixed to handle UNC paths correctly on windows. This might have the positive side-effect of fixing additional bugs that that stem from incorrect path handling.

Byron:
[...]

PR #1044, opened 2020-07-27
fix for extra back slashes for windows UNC paths

HudsonMC16:
This is to address UNC paths breaking, as discussed in issue #1019

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -419,6 +419,8 @@ class GitConfigParser(with_metaclass(MetaParserBuilder, cp.RawConfigParser, obje
                         optval = string_decode(optval[1:])
                     # end handle multi-line
                     # preserves multiple values for duplicate optnames
+                    # fix for extra back slashes so windows UNC paths are not mangled
+                    optval = optval.replace('\\\\', '\\')
                     cursect.add(optname, optval)
                 else:
                     # check if it's an option with no value - it's just ignored by git
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,4 +42,5 @@ Contributors are:
 -Harmon <harmon.public _at_ gmail.com>
 -Liam Beguin <liambeguin _at_ gmail.com>
 -Ram Rachum <ram _at_ rachum.com>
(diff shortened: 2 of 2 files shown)


git/config.py
```diff
@@ -419,6 +419,8 @@ def string_decode(v):
                         optval = string_decode(optval[1:])
                     # end handle multi-line
                     # preserves multiple values for duplicate optnames
+                    # fix for extra back slashes so windows UNC paths are not mangled
+                    optval = optval.replace('\\\\', '\\')
```
Byron, line 423: Have you checked what happens if that file is written back to disk?
Maybe this change could be constrained to files that are open for reading only (if this is even known).

Otherwise I fear that writing such a mangled file back, for example because some other value changed, will cause it to become invalid for some users.

What is your thoughts on this?

HudsonMC16: It looks like it might be platform dependent. Config entries get written [here](https://github.com/gitpython-developers/GitPython/blob/30387f16920f69544fcc7db40dfae554bcd7d1cc/git/config.py#L516) and the `defenc` encoding is determined [here](https://github.com/gitpython-developers/GitPython/blob/30387f16920f69544fcc7db40dfae554bcd7d1cc/git/compat.py#L24). I'll test it on my win10 machine, and if it works, we might wrap that `replace` call in an [`if is_win:`](https://github.com/gitpython-developers/GitPython/blob/30387f16920f69544fcc7db40dfae554bcd7d1cc/git/compat.py#L21) statement as a precaution.

Byron:
@HudsonMC16 I reworded the contribution guidelines to make clear adding oneself to the AUTHORS file is not a requirement at all.

NotoriousRebel:
Any updates?

HudsonMC16:
I'm sorry, I completely lost track of this. I'll run some tests after new year's.

Byron:
Thanks for the work put into this PR from everyone involved. I hope it can be resubmitted once there is some time to see it through. Thanks for your understanding.

Issue #1053, opened 2020-09-02
Config reader does not support includeIf


Svenito:
If the `.gitconfig` file includes an `includeIf` directive, the config parser doesn't respect it.

`~/.gitconfig` file contents:

```
[includeIf "gitdir:~/work/git/"]
        path = ~/.gitconfig.work

[user]
        email = my@email.com
        name = Me
        useconfigonly = true
```

`~/.gitconfig.work` contents

```
[user]
        email: work@email.com
        name = Work
```

Change to the `~/work/git/project` directory and run `git config -l` and notice that the name is set to `Work`. Running the following in a Python REPL

```
import os
from git import Repo
repo = Repo(os.getcwd())
repo.repo.config_reader().get_value("user", "name")
'Me'
```
And if the user hasn't got a default user setting in the `~/gitconfig` file, it will throw `ConfigParser.NoSectionError: No section: 'user'` which is expected.

Config parser should support the `includeIf` directive in the config file.

Byron:
Thanks for the summary!

As the implementation does support include directives already, at least in theory that should provide the necessary infrastructure to add support for conditional includes.

Here is the [a link to the git documentation](https://git-scm.com/docs/git-config/2.15.4#_conditional_includes) in case anyone wants to give it a try.

buddly27:
@Byron I just submitted a PR: #1054 

Let me know if I missed anything!

PR #1054, opened 2020-09-03
Read conditional include

buddly27:
Closing #1053 

Read conditional includes as defined in:
https://git-scm.com/docs/git-config#_conditional_includes

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,4 +43,5 @@ Contributors are:
 -Liam Beguin <liambeguin _at_ gmail.com>
 -Ram Rachum <ram _at_ rachum.com>
 -Alba Mendez <me _at_ alba.sh>
+-Jeremy Retailleau <jeremy.retailleau _at_ gmail.com>
 Portions derived from other open source works and are clearly marked.
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,5 +43,4 @@ Contributors are:
 -Liam Beguin <liambeguin _at_ gmail.com>
 -Ram Rachum <ram _at_ rachum.com>
 -Alba Mendez <me _at_ alba.sh>
--Jeremy Retailleau <jeremy.retailleau _at_ gmail.com>
 Portions derived from other open source works and are clearly marked.
--- a/git/config.py
+++ b/git/config.py
@@ -464,35 +464,37 @@ class GitConfigParser(with_metaclass(MetaParserBuilder, cp.RawConfigParser, obje
                 paths += self.items(section)
 
             match = CONDITIONAL_INCLUDE_REGEXP.search(section)
-            if match is not None and self._repo is not None:
-                keyword = match.group(1)
-                value = match.group(2).strip()
+            if match is None or self._repo is None:
+                continue
 
-                if keyword in ["gitdir", "gitdir/i"]:
-                    value = osp.expanduser(value)
+            keyword = match.group(1)
+            value = match.group(2).strip()
 
-                    if not any(value.startswith(s) for s in ["./", "/"]):
-                        value = "**/" + value
-                    if value.endswith("/"):
-                        value += "**"
+            if keyword in ["gitdir", "gitdir/i"]:
+                value = osp.expanduser(value)
 
-                    # Ensure that glob is always case insensitive if required.
-                    if keyword.endswith("/i"):
-                        value = re.sub(
-                            r"[a-zA-Z]",
-                            lambda m: "[{}{}]".format(
-                                m.group().lower(),
-                                m.group().upper()
-                            ),
-                            value
-                        )
+                if not any(value.startswith(s) for s in ["./", "/"]):
+                    value = "**/" + value
+                if value.endswith("/"):
+                    value += "**"
 
-                    if fnmatch.fnmatchcase(self._repo.git_dir, value):
-                        paths += self.items(section)
+                # Ensure that glob is always case insensitive if required.
+                if keyword.endswith("/i"):
+                    value = re.sub(
+                        r"[a-zA-Z]",
+                        lambda m: "[{}{}]".format(
+                            m.group().lower(),
+                            m.group().upper()
+                        ),
+                        value
+                    )
 
-                elif keyword == "onbranch":
-                    if fnmatch.fnmatchcase(self._repo.active_branch.name, value):
-                        paths += self.items(section)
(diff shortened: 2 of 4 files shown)


test/test_config.py
```diff
@@ -238,6 +240,111 @@ def check_test_value(cr, value):
...
             check_test_value(cr, tv)
 
+    @with_rw_directory
+    def test_conditional_includes_from_git_dir(self, rw_dir):
+        # Initiate repository path
+        git_dir = osp.join(rw_dir, "target1", "repo1")
+        os.makedirs(git_dir)
+
+        # Initiate mocked repository
+        repo = mock.Mock(git_dir=git_dir)
```
Byron, line 250: So nice to see a mock for once. It's my first time in Python.

test/test_config.py
```diff
@@ -238,6 +240,111 @@ def check_test_value(cr, value):
...
+
+        # Initiate config files.
+        path1 = osp.join(rw_dir, "config1")
+        path2 = osp.join(rw_dir, "config2")
+        template = "[includeIf \"{}:{}\"]\n    path={}\n"
+
+        with open(path1, "w") as stream:
+            stream.write(template.format("gitdir", git_dir, path2))
+
+        # Ensure that config is ignored if no repo is set.
```
Byron, line 260: It's great to have blocks with a description explaining in plain language what's asserted on below.

AUTHORS
```diff
@@ -43,4 +43,5 @@ Contributors are:
 -Liam Beguin <liambeguin _at_ gmail.com>
 -Ram Rachum <ram _at_ rachum.com>
 -Alba Mendez <me _at_ alba.sh>
+-Jeremy Retailleau <jeremy.retailleau _at_ gmail.com>
```
Byron, line 46: Having an AUTHORS file was inherited from when I took over the project a decade ago. The contribution guidelines shouldn't force anyone to add themselves there, please feel free to remove yourself if it doesn't feel right.

The contribution guidelines have been updated to reflect that stance - I wasn't aware it actually demanded that, sorry for that.

git/config.py
```diff
@@ -443,7 +452,49 @@ def string_decode(v):
...
+        """Return all paths that must be included to configuration.
+        """
+        paths = []
+
+        for section in self.sections():
+            if section == "include":
+                paths += self.items(section)
+
+            match = CONDITIONAL_INCLUDE_REGEXP.search(section)
+            if match is not None and self._repo is not None:
```
Byron, line 467: I think it would be better to 'continue' if one of the prerequisites isn't met, reducing the indentation level by one.
Along the lines of this:
```python
if match is None or self._repo is None:
    continue
```

git/config.py
```diff
@@ -443,7 +452,49 @@ def string_decode(v):
...
+                                m.group().upper()
+                            ),
+                            value
+                        )
+
+                    if fnmatch.fnmatchcase(self._repo.git_dir, value):
+                        paths += self.items(section)
+
+                elif keyword == "onbranch":
+                    if fnmatch.fnmatchcase(self._repo.active_branch.name, value):
```
Byron, line 494: What happens if there is no active branch? This could happen when `HEAD` is detached, i.e. points to a commit directly.
I would assume that `active_branch` is then `None` or throws, and this should be handled here. It's probably easy to test for once it's clear what actually happens then.

git/config.py
```diff
@@ -443,7 +452,49 @@ def string_decode(v):
...
+                        value = "**/" + value
+                    if value.endswith("/"):
+                        value += "**"
+
+                    # Ensure that glob is always case insensitive if required.
+                    if keyword.endswith("/i"):
+                        value = re.sub(
+                            r"[a-zA-Z]",
+                            lambda m: "[{}{}]".format(
+                                m.group().lower(),
```
Byron, line 484: Just as a remark: I really don't know how lower casing works in a unicode world, and if any of this is safe to do on paths. It's incredibly hard to get right, especially in Python.

Byron (changes requested):
Thanks a lot for tackling this issue. It's a fantastic display on how tests can be written, and I hope future PRs will have more of that.

I found two spots to look at or discuss, along with a lot of praise in the other comments.

Please let me know what you think.

AUTHORS
buddly27: Thanks for the clarification. I'm gonna remove my name then as git history should be enough to find out who did what :)

git/config.py
buddly27: True, I'm gonna change that

git/config.py
buddly27: True, it seems that a `TypeError` exception is raise when `HEAD` is detached
https://github.com/gitpython-developers/GitPython/blob/master/git/refs/symbolic.py#L272

```console
> repo = git.Repo("/somewhere")
> repo.active_branch
...
TypeError: HEAD is a detached symbolic reference as it points to 'ae451cd89c1b727ba632066c8b8a80ab8d22a0f8'
```

I'll had a `try.. except..` to just ignore this `includeIf` when branch cannot be retrieved.

git/config.py
buddly27: I think it is shouldn't be an issue as Python should be abstracting the complexity for us
https://docs.python.org/3.8/library/stdtypes.html#str.lower
https://docs.python.org/3.8/library/stdtypes.html#str.upper

Did you ever run into any issue with these two methods? Maybe in earlier Python 3 versions?

git/config.py
buddly27: Added

git/config.py
buddly27: Changed

test/test_config.py
buddly27: I still mostly code in a Python 2.7 environment so I didn't know that mock was added to the unittest module, very nice addition :)
https://docs.python.org/3/library/unittest.mock-examples.html

git/config.py
Byron: Maybe at the time this code runs, it's already in the clear as it already is using a `str` which should have a known encoding.
The whole encoding story in GitPython is so thoroughly broken also due to Python making it so easy to get wrong that I simply don't trust any of it, which probably is a overly generic 😅.

Byron:
Fantastic, thanks a lot. I am going to create a new release soon!

PR #1063, opened 2020-09-28
Find paths ignored in .gitignore

saroad2:
Using [git check-ignore](https://git-scm.com/docs/git-check-ignore) in order to find whether paths are ignored or not in .gitignore

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -697,6 +697,19 @@ class Repo(object):
         finalize_process(proc)
         return untracked_files
 
+    def get_ignored(self, *paths):
+        """Checks if paths are ignored via .gitignore
+        Doing so using the "git check-ignore" method.
+
+        :param paths: List of paths to check whether they are ignored or not
+        :return: sublist of ignored paths
+        """
+        try:
+            proc = self.git.check_ignore(*paths)
+        except GitCommandError:
+            return []
+        return proc.replace("\\\\", "\\").replace('"', "").split("\n")
+
     @property
     def active_branch(self):
         """The name of the currently active branch.
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -702,7 +702,7 @@ class Repo(object):
         Doing so using the "git check-ignore" method.
 
         :param paths: List of paths to check whether they are ignored or not
-        :return: sublist of those paths which are ignored
+        :return: subset of those paths which are ignored
         """
         try:
             proc = self.git.check_ignore(*paths)
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -697,12 +697,12 @@ class Repo(object):
         finalize_process(proc)
         return untracked_files
 
-    def get_ignored(self, *paths):
+    def ignored(self, *paths):
         """Checks if paths are ignored via .gitignore
         Doing so using the "git check-ignore" method.
 
         :param paths: List of paths to check whether they are ignored or not
-        :return: sublist of ignored paths
+        :return: sublist of those paths which are ignored
         """
         try:
             proc = self.git.check_ignore(*paths)


git/repo/base.py
```diff
@@ -697,6 +697,19 @@ def _get_untracked_files(self, *args, **kwargs):
...
+        Doing so using the "git check-ignore" method.
+
+        :param paths: List of paths to check whether they are ignored or not
+        :return: sublist of ignored paths
+        """
+        try:
+            proc = self.git.check_ignore(*paths)
+        except GitCommandError:
+            return []
+        return proc.replace("\\\\", "\\").replace('"', "").split("\n")
```
Byron, line 711: I believe this is not sufficient to undo the [c-style quoting](https://github.com/git/git/blob/9bc233ae1cf19a49e51842c7959d80a675dbd1c0/builtin/check-ignore.c#L44:L44) that git applies to the output, even though superficially it may seem to be the case.
In case you think it's worth improving this, here [is the code to unquote in C](https://github.com/git/git/blob/9bc233ae1cf19a49e51842c7959d80a675dbd1c0/quote.c#L394:L394) and [here it is in Rust](https://github.com/Byron/gitoxide/blob/e355b4ad133075152312816816af5ce72cf79cff/git-odb/src/alternate/unquote.rs#L22).
If not, just let me know and we go ahead with the merge, fixing it when the first bug report comes in.

Byron, line 711: I believe this is not sufficient to undo the [c-style quoting](https://github.com/git/git/blob/9bc233ae1cf19a49e51842c7959d80a675dbd1c0/builtin/check-ignore.c#L44:L44) that git applies to the output, even though superficially it may seem to be the case.
In case you think it's worth improving this, here [is the code to unquote in C](https://github.com/git/git/blob/9bc233ae1cf19a49e51842c7959d80a675dbd1c0/quote.c#L394:L394) and [here it is in Rust](https://github.com/Byron/gitoxide/blob/e355b4ad133075152312816816af5ce72cf79cff/git-odb/src/alternate/unquote.rs#L22).
If not, just let me know and we go ahead with the merge, fixing it when the first bug report comes in.

git/repo/base.py
```diff
@@ -697,6 +697,19 @@ def _get_untracked_files(self, *args, **kwargs):
         finalize_process(proc)
         return untracked_files
 
+    def get_ignored(self, *paths):
```
Byron, line 700: This method should be `ignored`, as there is no `get_` prefix for the other getters either.

git/repo/base.py
```diff
@@ -697,6 +697,19 @@ def _get_untracked_files(self, *args, **kwargs):
...
+        """Checks if paths are ignored via .gitignore
+        Doing so using the "git check-ignore" method.
+
+        :param paths: List of paths to check whether they are ignored or not
+        :return: sublist of ignored paths
```
Byron, line 705: What's the difference between `sublist` and `list`?

Byron (changes requested):
Thanks a lot for you contribution! 

The merge is just two minor changes away, with the option for a bigger change for correctness. The latter won't block a merge though in case there is no time for that.

git/repo/base.py
saroad2: I would love to make this a more bulletproof conversion, but unfortunately, I don't know how to read C and Rust code.
If you want, feel free to edit this PR and make the needed changes.

git/repo/base.py
saroad2: What I meant is that I'm returning a `list` object which is a subset of the given paths which are ignored. Updated the documentation in order to make it clearer.

git/repo/base.py
```diff
@@ -697,6 +697,19 @@ def _get_untracked_files(self, *args, **kwargs):
         finalize_process(proc)
         return untracked_files
 
+    def ignored(self, *paths):
+        """Checks if paths are ignored via .gitignore
+        Doing so using the "git check-ignore" method.
+
+        :param paths: List of paths to check whether they are ignored or not
+        :return: sublist of those paths which are ignored
```
Byron, line 705: I think sublist really wants to be ‘subset’, then it would make sense to me.

Byron (changes requested):
Thanks for the improvements! I think that leaves just one tiny thing before it can be merged.
Thanks for bearing with me.

Byron:
Great, thanks again!

saroad2:
Thanks a lot for the merge!

When are you planning on releasing a new version to PyPi? I would like to use this new ability in one of my projects.

Byron:
Version 3.1.9 was just published to PyPi. Cheers.

Issue #1066, opened 2020-10-04
Conventions for sys.path are not followed


xverges:
From https://docs.python.org/3/library/sys.html#sys.path

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter. If the script directory is not available (e.g. if the interpreter is invoked interactively or if the script is read from standard input), path[0] is the empty string, which directs Python to search modules in the current directory first.

Instead, `sys.path[0]` points to `gitdb`

Byron:
Thanks for letting me know and provide the fix right away, it's much appreciated.
Since the PR was merged, I think it's safe to close this issue.

PR #1067, opened 2020-10-04
Do not break convention when updating sys.path

xverges:
Refers to #1066


xverges:
Created using the wrong branch. Sorry

PR #1068, opened 2020-10-04
Do not break sys.path convention when updating it

xverges:
Refers to #1066


Byron:
Thanks a lot! Great work!i

PR #1073, opened 2020-10-23
Fix default actor name handling

athos-ribeiro:
In c96476b, the new default_name nested function does not contain a
retun statement. This leads to an issue when the environment variables
are not present, where the actor name would not be set.

Signed-off-by: Athos Ribeiro <athos@redhat.com>

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -591,7 +591,7 @@ class Actor(object):
             return user_id
 
         def default_name():
-            default_email().split('@')[0]
+            return default_email().split('@')[0]
 
         for attr, evar, cvar, default in (('name', env_name, cls.conf_name, default_name),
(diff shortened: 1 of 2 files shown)


git/util.py
```diff
@@ -591,7 +591,7 @@ def default_email():
             return user_id
 
         def default_name():
-            default_email().split('@')[0]
+            return default_email().split('@')[0]
```
Byron, line 594: It’s funny this slipped by me as well since I only write Rust these says, in which this is exactly what we would have expected.
Maybe I have just been a sloppy reviewer though 😅

athos-ribeiro: Thanks for catching this one early, @harupy!

Byron:
v3.1.11 was just released.

Issue #1081, opened 2020-11-09
UnicodeDecodeError in diff filename


ishepard:
Hey @Byron,
just another UnicodeDecodeError here :) this time in the filename of a diff object. Originally opened [here](https://github.com/ishepard/pydriller/issues/144).

[This](https://github.com/adobe/brackets/commit/8b3ae041041dfeecd059c2b19c72e76223e501d3) is the culprit.
You can repro with this code:
```
from git import Repo
r = Repo("/tmp/brackets")
commit = r.commit("8b3ae041041dfeecd059c2b19c72e76223e501d3")

diff = commit.parents[0].diff(commit, create_patch=True)
```
The error:
```
  File ".../tmp.py", line 13, in <module>
    diff = commit.parents[0].diff(commit, create_patch=True)
  File ".../venv/lib/python3.8/site-packages/git/diff.py", line 145, in diff
    index = diff_method(self.repo, proc)
  File ".../venv/lib/python3.8/site-packages/git/diff.py", line 455, in _index_from_patch_format
    index.append(Diff(repo,
  File ".../venv/lib/python3.8/site-packages/git/diff.py", line 282, in __init__
    if submodule.path == a_rawpath.decode("utf-8"):
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 54: invalid continuation byte
```

The error is on [this line](https://github.com/gitpython-developers/GitPython/blob/master/git/diff.py#L278). We don't decode correctly the filename.

Now, we have many fixes..which one do you prefer? 
I saw on the same file you use many times the 'replace' (like [here](https://github.com/gitpython-developers/GitPython/blob/master/git/diff.py#L371)). I can open a PR with this change if you think it's a good idea.

Byron:
I see, one more encoding issue indeed 😅.

A PR with a similar fix would be appreciated as at least it would be consistent. However, if there is a better way (without breaking backwards compatibility), that should certainly be considered as well. To my mind the right way to deal with this is to not actually assume any encoding but to work with bytes only. [`gitoxide`](https://github.com/Byron/gitoxide) is getting that right as you can imagine :D.

ishepard:
Indeed, working with bytes instead of unicodes would be the best..especially now that we dropped py2 support. Also, git returns bytes, I think..so instead of decoding it, we just give back bytes to the users.

[...]

PR #1082, opened 2020-11-10
Fix UnicodeDecodeError in diff filename

ishepard:
Fix for #1081.

If the filename of a diff object contained non UTF-8 chars, GitPython was raising a UnicodeDecodeError.

This PR fixes this by using the "replace" flag when decoding.
Furthermore, the variable `defenc` is used, that is `sys.getfilesystemencoding()`.


Byron:
Thanks a lot for the fix, it's much appreciated!

PR #1085, opened 2020-11-18
WIP: Added ability to define git environment in submodule add/update methods

isolovey:
Define git environment in submodule add/update methods

Basically just propagating the `env` dictionary to the `clone_from` invocation inside submodule add/update methods. Does that make sense to merge upstream?


Byron:
Fantastic work, what a nice addition!

PR #1102, opened 2021-01-05
Add '-z' on top of '--raw' to avoid path name mangling

### LINKED ISSUE #1099: Diff::a_path is wrong for unicode characters
The problem is slightly mentioned in #865.

Let's consider the following example:
```shell
git init foo
cd ./foo
echo a > a
git add a
git commit -a -m 'Add a'
echo "foo" > špatně.txt
git add špatně.txt
git commit -a -m Add
```

```python3
In [4]: Repo('.').commit().diff('HEAD~')[0].a_path
Out[4]: '"\\305\\241patn\\304\\233.txt"'
```

Unless one uses `git config --global core.quotepath false`, the file name is quoted and I don't see a simple solution how to decode it to unicode? One can use `-z` option that can resolve the problem:

```
git diff HEAD~ --raw -z | cat
:000000 100644 0000000 257cc56 Ašpatně.txt
```

marxin:
It's follow-up of #1101.

Fixes #1099

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -108,6 +108,7 @@ class Diffable(object):
             args.append("-p")
         else:
             args.append("--raw")
+            args.append("-z")
 
         # in any way, assure we don't see colored output,
         # fixes https://github.com/gitpython-developers/GitPython/issues/172
@@ -483,7 +484,7 @@ class Diff(object):
             if not line.startswith(":"):
                 return
 
-            meta, _, path = line[1:].partition('\t')
+            meta, _, path = line[1:].partition('\x00')
             old_mode, new_mode, a_blob_id, b_blob_id, _change_type = meta.split(None, 4)
             # Change type can be R100
             # R: status letter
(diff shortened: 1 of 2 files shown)


git/diff.py
```diff
@@ -479,55 +479,55 @@ def _index_from_raw_format(cls, repo, proc):
...
-            # END add/remove handling
-
-            diff = Diff(repo, a_path, b_path, a_blob_id, b_blob_id, old_mode, new_mode,
-                        new_file, deleted_file, copied_file, rename_from, rename_to,
-                        '', change_type, score)
-            index.append(diff)
+        def handle_diff_line(lines):
+            lines = lines.decode(defenc)
+
+            for line in lines.split(':')[1:]:
```
Byron, line 485: That's interesting! Wouldn't that mean that with typical windows paths one would get line splits where there are none?

Byron: Possible answer: Maybe there are no 'windows' paths as all paths are relative to the git repository 🤦‍♂️.

marxin: All right.

PR #1105, opened 2021-01-09
Fix package name in intro docs to reflect actual pypi name

Tossrock:
The package uses PascalCase on pypi. Having the wrong name in the docs is confusing.


Byron:
Thanks. I hope installation also works in lowercase though.

PR #1110, opened 2021-01-16
Make git.cmd.Git.CatFileContentStream iterable

Hex052:
Add __next__ method to git.cmd.Git.CatFileContentStream, so it can actually be used as an iterable

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -497,6 +497,9 @@ class Git(LazyMixin):
         # skipcq: PYL-E0301
         def __iter__(self):
             return self
+        
+        def __next__(self):
+            return self.next()
 
         def next(self):
             line = self.readline()


Byron:
How does an iterator indicate 'end of iteration'? If I remember correctly, it needs to throw a special exception which would break the API of 'next()'.
If iteration should be implemented, that could of course be done but should not break existing APIs.
Maybe I am missing something though.

Hex052:
To end iteration, an iterator needs to raise the exception [`StopIteration`](https://docs.python.org/3/library/exceptions.html#StopIteration), which is currently done in the `next()` method. To avoid breaking anything that might possibly use the existing `next()` method, I just added a new function that called the existing one. The existing function will only raise `StopIteration` at the very end (not on blank lines) since `readline()` does not strip any trailing newlines.

I added `__next__` because [python requires both `__iter__` and `__next__`](https://docs.python.org/3/library/stdtypes.html?highlight=__next__#iterator-types) in the iterator.

Hopefully that clears it up?

Byron:
Thanks a lot for filling in the blanks :) and for making this contribution!

PR #1117, opened 2021-01-28
fix universal_newlines TypeError

x-santiaga-x:


Byron:
Thanks a lot!

I made a mistake and accidentally amended to commit bffe076 which prevents GitHub to see that the commit is already in master. Thus this PR is merged despite not looking like it.

PR #1124, opened 2021-02-16
add replace method to git.Commit

### LINKED ISSUE #1123: RFE: Give Commit objects a replace method that returns new Commit with modified attributes
The `datetime` objects produced by Python's `datetime` module are immutable, much like `Commit` objects. The `datetime` module provides a convenient `replace` function that returns a new object based on an old with one or more attributes replaced.  E.g, I can write:

```
>>> d = datetime.datetime.now()
>>> d.replace(hour=0, minute=0, second=0)
datetime.datetime(2021, 2, 15, 0, 0, 0, 318021)
```

I would like to see the same api on `Commit` objects, so that one could write:

```
>>> newcommit = oldcommit.replace(message='This is a test', author='Cookie Monster')
```

larsks:
This adds a replace method to git.Commit. The replace method returns a
copy of the Commit object with attributes replaced from keyword
arguments. For example:

```
>>> old = repo.head.commit
>>> new = old.replace(message='This is a test')
```

closes #1123

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -136,6 +136,41 @@ class Commit(base.Object, Iterable, Diffable, Traversable, Serializable):
     def _get_intermediate_items(cls, commit):
         return commit.parents
 
+    @classmethod
+    def _calculate_sha_(cls, repo, commit):
+        '''Calculate the sha of a commit.
+
+        :param repo: Repo object the commit should be part of
+        :param commit: Commit object for which to generate the sha
+        '''
+
+        stream = BytesIO()
+        commit._serialize(stream)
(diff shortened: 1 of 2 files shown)


Byron:
Thanks so much, absolutely perfect work (at least as far as I can tell :)).

I must admit that initially I was skeptical when reading the corresponding issue, but now that I see the implementation I couldn't find a reason not to have this capability.

If you would like the next release to contain a dedicated and deserved entry in the changelog (`change.rst`), please feel free to submit another PR. The next release is not scheduled but from experience these happen every few weeks.

PR #1126, opened 2021-02-17
add a changelog entry for #1124

larsks:
Thanks for the kind words on #1124! This PR adds a changelog entry, although I wasn't quite sure what you wanted that to look like for a future release. I noticed there were two entries for 3.1.12, and following the links it looks like the latest should have been 3.1.13 so I fixed that while I was there.

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -2,7 +2,13 @@
 Changelog
 =========
 
-3.1.12
+3.1.??
+======
+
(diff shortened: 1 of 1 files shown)


Byron:
That's pretty much what I had in mind, and thanks again for the fly-by fix :D! When looking at these kinds of mistakes I think I probably shouldn't be steering this ship ;).

PR #1131, opened 2021-02-24
Dropping python 3.4

Yobmod:
#1095
I've removed python 3.4 compat everywhere I could find it.
Except in the Dockerfile (still has py2.7 listed) and submodules.

All tests on Linux (ubuntu 20.04) still passed.
On windows 10, I get: 409 tests run, FAILED (failures=1, errors=6, skipped=14). But that is the same as before I made any changes.

Also added ".mypy_cache/" ".pytest_cache/" to the .gitignore


Byron:
Please note that I am waiting with the merge until #1132 is clarified.

PR #1184, opened 2021-02-26
Add python 3.9

### LINKED ISSUE #1132: Adding python 3.9, release prior to dropping 3.4.
#1131 

Hi,
When running the tests after dropping 3.4, I also tested against 3.9 on Linux. All tests passed, so I was thinking to add simply add 3.9. But will probably want final 3.4 release cutting before either are merged?

Dropping 3.4 allows the use of function type hints and async/await coroutines.

Note: According to https://pypistats.org/packages/gitpython python 3.4 downloads are at 0.04% and 3.5 is still at 3.38%

Yobmod:
Added python 3.9
fix #1132


Byron:
Thanks a lot! Ready to move towards dropping py 3.4 and adding the first types.

PR #1193, opened 2021-03-11
 Restore order of operators before executing the git command only for < py3.6

bertwesarg:
Since Python 3.6 kwargs are stable ordered again and Python 3.5 reached
its end-of-life too, so we can revert 89ade7bfff534ae799d7dd693b206931d5ed3d4f
again. Thus make it able to pass ordered options to Git commands again.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -19,6 +19,7 @@ import sys
 import threading
 from collections import OrderedDict
 from textwrap import dedent
+import warnings
 
 from git.compat import (
     defenc,
@@ -902,8 +903,14 @@ class Git(LazyMixin):
 
     def transform_kwargs(self, split_single_char_options=True, **kwargs):
         """Transforms Python style kwargs into git command line options."""
+        # Python 3.6 preserves the order of kwargs and thus has a stable
+        # order. For older versions sort the kwargs by the key to get a stable
+        # order.
+        if sys.version_info[:2] < (3, 6):
+            kwargs = OrderedDict(sorted(kwargs.items(), key=lambda x: x[0]))
+            warnings.warn("Python 3.5 support is deprecated. It does not preserve the order\n" +
+                          "for key-word arguments. Thus they will be sorted!!")
         args = []
-        kwargs = OrderedDict(sorted(kwargs.items(), key=lambda x: x[0]))
         for k, v in kwargs.items():
             if isinstance(v, (list, tuple)):
                 for value in v:


.travis.yml
```diff
@@ -38,7 +36,6 @@ script:
   - ulimit -n
   - coverage run --omit="test/*" -m unittest --buffer
   - coverage report
-  - if [ "$TRAVIS_PYTHON_VERSION" == '3.5' ]; then cd doc && make html; fi
```
Byron, line 41: I think this should check for `'3.6'` instead to not loose documentation validation.

Byron (changes requested):

Does that also mean that kwargs are now ordered by default? Or does it mean that the order won't change between calls even though the order is not as specified?
In the latter case, the result might still be considered unstable compared to ordering them.

What do you think?

bertwesarg:
> What do you think?

The PEP speaks about _preserving order_, this should also mean, that the order wont change _between calls_. I cannot reproduce the example `git.remote.pull()`, given in the original commit. But I can reproduce the _random_ order of kwargs between two executions of Python 3.5 with this code:

```python
def pkwargs(**kwargs):
    print(kwargs)
pkwargs(**{'one': 1, 'two': 2, 'three': 3, 'four': 4})
```

This gives me one of these possible outputs:

```
{'three': 3, 'four': 4, 'one': 1, 'two': 2}
{'two': 2, 'one': 1, 'three': 3, 'four': 4}
```

For Python 3.6 it is stable and preserves the order:

```
{'one': 1, 'two': 2, 'three': 3, 'four': 4}
```

bertwesarg:
The PEP link is also in the commit message: https://www.python.org/dev/peps/pep-0468/

Byron:
Thanks a lot! Now this PR would be ready for merge even though I feel a little uncomfortable to remove support half a year after py3.5 reached its end of life given that it took 2 years to remove 3.4 support to be able to support inline types which has incredibly high value.

My suggestion is to wait with the merge till Sept. 5, 2020 giving folks another half a year to stop using python 3.5 or face the potential for breakage.

If you would like a merge right now one could add a python version check to keep the sort code for py3.5 maybe along a depreciation warning to let people know the date of removal.

Thanks for sharing your thoughts.

bertwesarg:
I don't think, shifting the date further back will have any effect. Most of us (the users) will drop py3.5, ony if we notice, that its not supported by the package we are using. But in the end, this is your call.

I also do not have a problem, to make this conditionally on the python version in use, so stay tuned for an update, without the py3.5 drop.

git/cmd.py
```diff
@@ -902,8 +902,12 @@ def transform_kwarg(self, name, value, split_single_char_options):
 
     def transform_kwargs(self, split_single_char_options=True, **kwargs):
         """Transforms Python style kwargs into git command line options."""
+        # only since 3.6 Python preserves the order of kwargs and thus has a stable
+        # order. For older versions sort the kwargs by the key to get a stable
+        # order.
+        if sys.version_info[:2] < (3, 6):
```
Byron, line 908: In order to make people aware, could you add a deprecation warning mentioning the date until this will work?

Byron:
Thanks a lot for your help!

I will set a reminder for myself to remove python 3.5 support on September 5th. It would be super helpful to have a PR waiting which does that - the code was present here but seems to have been overwritten.

PR #1198, opened 2021-03-12
Replace password in URI by stars if present to avoid leaking secrets in logs

mickours:
Currently, if you use a git URI containing a password like `https://username:pass@git.example.com/project` and an exception occurs during a git clone the password is printed in the logs. It also happen when the logging is at debug level.

To avoid secrets to leak in the logs, this patch replace the password by a fix amount of stars (`******`).

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -9,6 +9,7 @@ import logging
 import os
 import re
 import warnings
+from urllib.parse import urlsplit, urlunsplit
 
 from git.cmd import (
     Git,
@@ -971,10 +972,15 @@ class Repo(object):
             (stdout, stderr) = proc.communicate()
             cmdline = getattr(proc, 'args', '')
             uri = cmdline[-2]
-            if "://" in uri and "@" in uri:
-                cred = uri.split("://")[1].split("@")[0].split(":")
-                if len(cred) == 2:
-                    cmdline[-2] = uri.replace(cred[1], "******")
+            try:
+                url = urlsplit(uri)
+                # Remove password from the URL if present
+                if url.password:
+                    edited_url = url._replace(
+                        netloc=url.netloc.replace(url.password, "****"))
+                    cmdline[-2] = urlunsplit(edited_url)
+            except ValueError:
+                log.debug("Unable to parse the URL %s", url)
             log.debug("Cmd(%s)'s unused stdout: %s", cmdline, stdout)
             finalize_process(proc, stderr=stderr)
 
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -9,7 +9,6 @@ import logging
 import os
 import re
 import warnings
-from urllib.parse import urlsplit, urlunsplit
 
 from git.cmd import (
     Git,
@@ -27,7 +26,7 @@ from git.index import IndexFile
 from git.objects import Submodule, RootModule, Commit
 from git.refs import HEAD, Head, Reference, TagReference
 from git.remote import Remote, add_progress, to_progress_instance
-from git.util import Actor, finalize_process, decygpath, hex_to_bin, expand_path
+from git.util import Actor, finalize_process, decygpath, hex_to_bin, expand_path, remove_password_if_present
 import os.path as osp
 
 from .fun import rev_parse, is_git_dir, find_submodule_git_dir, touch, find_worktree_git_dir
@@ -971,16 +970,8 @@ class Repo(object):
         else:
             (stdout, stderr) = proc.communicate()
             cmdline = getattr(proc, 'args', '')
-            uri = cmdline[-2]
-            try:
-                url = urlsplit(uri)
-                # Remove password from the URL if present
-                if url.password:
-                    edited_url = url._replace(
-                        netloc=url.netloc.replace(url.password, "****"))
-                    cmdline[-2] = urlunsplit(edited_url)
-            except ValueError:
-                log.debug("Unable to parse the URL %s", url)
+            cmdline = remove_password_if_present(cmdline)
+
             log.debug("Cmd(%s)'s unused stdout: %s", cmdline, stdout)
             finalize_process(proc, stderr=stderr)
 
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -969,7 +969,13 @@ class Repo(object):
             handle_process_output(proc, None, progress.new_message_handler(), finalize_process, decode_streams=False)
         else:
             (stdout, stderr) = proc.communicate()
-            log.debug("Cmd(%s)'s unused stdout: %s", getattr(proc, 'args', ''), stdout)
+            cmdline = getattr(proc, 'args', '')
+            uri = cmdline[-2]
+            if "://" in uri and "@" in uri:
+                cred = uri.split("://")[1].split("@")[0].split(":")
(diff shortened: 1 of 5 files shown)


Byron:
Thanks a lot for the initiative, it's much appreciated.

My early feedback is that the URL parsing is probably prone to failures even though it should work for most URLs. To be save, one would probably catch parsing exceptions.
Maybe it would be easier to use an existing URL parser and/or serializer if one exists in the python standard library.

mickours:
Thanks for the feedback! I've change the implementation to use urllib and handle parsing errors.

I'm still unable to make the test works on the CI despite the fact that it works locally. It seems that the failing command raise an assert which is not in my test. The problem is that  I need the exception to be raise in order to check if the password is inside. Any clue on that?

Byron:
Thanks for the update!

The failing test mentions the entire command-line, here it is:

```
 /usr/bin/git clone -v --progress ***fakerepo.example.com/testrepo /tmp/test_leaking_password_in_clone_logsam5xqh6q'
```

It looks like a replacement has happened even though I would expect an additional `*`, the password has been removed but it still seems to see the password in the error message. Since `finalize_proc` is only handed the process `proc`, this really only works if the `args` list is editable. The latter might be the case in some python versions, but not in others.

Something you could try is to reassign `proc.args` entirely. If that works, I have a few more requests just to be very very safe 😅.

git/repo/base.py
```diff
@@ -969,7 +970,18 @@ def _clone(cls, git, url, path, odb_default_type, progress, multi_options=None,
             handle_process_output(proc, None, progress.new_message_handler(), finalize_process, decode_streams=False)
         else:
             (stdout, stderr) = proc.communicate()
-            log.debug("Cmd(%s)'s unused stdout: %s", getattr(proc, 'args', ''), stdout)
+            cmdline = getattr(proc, 'args', '')
+            uri = cmdline[-2]
```
Byron, line 974: I think there is at least two arguments here, so `cmdline[-2]` will never fail. However, it would be great to only run the url parsing code if `cmdline[1]` is actually `clone`.

Byron: In a second step, one could consider other `git` commands as well or maybe even search arguments for URLs with password, allowing to fetches and pulls as well along with less common invocations.

Byron (changes requested):
Thanks a lot for bearing with me, and good luck with the CI issue. The test succeeds for me as well when running it locally.

git/repo/base.py
mickours: You're right this is a more general problem. I think I've move the code to utils and use it also in cmd.execute.

mickours: Also I'm parsing any elem of the command line just to be sur

mickours:
I track down every occurrence of the command line in the logs and it seems to be OK now. With this patch it will on every commands try to redact password out of the command line before logging it.

mickours:
I just realize that we do not have a succesfull clone test and when I tried it manually it's not working because of the in-place replacement. I've to rework this.

mickours:
Ok, I've change the in-place password replacement with a copy implementation which avoid side effects on the command line arguments. Also, I've added a successful git clone with a password  from an empty project.

git/util.py
```diff
@@ -338,6 +339,34 @@ def expand_path(p, expand_vars=True):
...
+    This should be used for every log line that print a command line.
+    """
+    new_cmdline = []
+    for index, to_parse in enumerate(cmdline):
+        new_cmdline.append(to_parse)
+        try:
+            url = urlsplit(to_parse)
+            # Remove password from the URL if present
+            if url.password is None:
+                raise ValueError()
```
Byron, line 359: I think it would be nicer to `continue` here.

git/util.py
```diff
@@ -338,6 +339,34 @@ def expand_path(p, expand_vars=True):
...
+
+            edited_url = url._replace(
+                netloc=url.netloc.replace(url.password, "*****"))
+            new_cmdline[index] = urlunsplit(edited_url)
+        except ValueError:
+            # This is not a valid URL
+            pass
```
Byron, line 366: Here it would be nicer to say `continue` too.

test/test_repo.py
```diff
@@ -238,6 +238,21 @@ def test_clone_from_with_path_contains_unicode(self):
...
+        try:
+            Repo.clone_from(
+                url="https://fakeuser:{}@fakerepo.example.com/testrepo".format(
+                    password),
+                to_path=rw_dir)
+        except GitCommandError as err:
+            assert password not in str(err), "The error message '%s' should not contain the password" % err
+        # Working example from a blank private project
+        Repo.clone_from(
+            url="https://gitlab+deploy-token-392045:mLWhVus7bjLsy8xj8q2V@gitlab.com/mercierm/test_git_python",
```
Byron, line 253: This might be a danger as it breaks the self-isolation even more (after all, GitPythons test rely on its own repository already), but I think it's OK to go with it and fix it when it breaks.
Maybe it never does.

Byron (changes requested):
Thanks a lot, this looks so much better already, great work! It's particularly helpful to see more tests, as now there is one in particular for redacting passwords in command-lines. I bet this helped during development.

Please excuse the use of the word 'nicer' in the line comments below, I simply lacked a more scientific explanation so I guess it comes down to 'taste'. This is not to be understood as judgement towards your code, it's just as correct and since it's Python you could make a point towards the current version being idiomatic and I would go with it.

Let's get this one to the finishing line :) 🏇

test/test_repo.py
mickours: I agree! I simply did not find another way to test the working use case. Maybe we can use a deploy token with the GitPython repository on Github but I think it is not available for public repo. Maybe you can create a private repo inside the `gitpython-developers` account in order to keep the responsibility in the same place?

git/util.py
mickours: done

git/util.py
mickours: done

mickours:
> Thanks a lot, this looks so much better already, great work! It's particularly helpful to see more tests, as now there is one in particular for redacting passwords in command-lines. I bet this helped during development.
You're right! :)

> Please excuse the use of the word 'nicer' in the line comments below, I simply lacked a more scientific explanation so I guess it comes down to 'taste'. This is not to be understood as judgement towards your code, it's just as correct and since it's Python you could make a point towards the current version being idiomatic and I would go with it.
No problem, I know that sometimes it's a matter of beauty ;)

> Let's get this one to the finishing line :) horse_racing
Thanks a lot for all you feedback!! :checkered_flag:

test/test_repo.py
Byron: I thought about that too and will do that when it breaks. It's certainly the lazy way of doing things and it's basically a mine that is primed to blow sometime in the future, but… it's going to be okay ;).

While writing this I thought to myself that I don't want to be that lazy and took a look at the github ways of doing this. It turns out it only supports repository deploy keys, which indeed are full blown ssh keys that I don't want to deal with right now. This makes your account one of the pillars of the happiness of GitPython's CI, something not everyone can say about themselves :D.

Byron (approved):
And it's done :)! Thanks again, I think it was worth it, too!

PR #1201, opened 2021-03-13
DNM before 2021-09-05: Remove support for Python 3.5

bertwesarg:


Byron:
Thanks so much! I will merge this one in half a year from now 😁.

ssbarnea:
Why waiting so long for removal of py35? One extra year after py35 already hit EOL seems too much for me.

PS. I say this only because I know the py36 opens the door of static typing, which highly appreciated on any library.

Byron:
Thanks for sharing the value provided by dropping python 3.5 earlier.

Some types have already been added and for that python 3.4 support was dropped I believe, I assume python 3.6 provides even better ways of inline typing?

In any case, I would be happy to accept a PR which drops python 3.5 support while adding more types in a way that require python 3.6 and make good use of the value added by dropping support earlier.

Yobmod:
Just to add (prompted by  #1270):
I have been adding comments where typing is impacted by 3.5 support. Searching for "3.5 dropped" gives 9 comments so far.

Byron:
@Yobmod Does that mean this PR needs additional work? Otherwise I would be merging it early, yank the current version on pypi once more and drop python 3.5 support with the next.
Thank you

Yobmod:
I dont think any reason to delay. 

Once this is merged I'll submit a separate PR with the typing improvements it makes possible. 
They won't change much but will cleanup some cast() and type ignores.

bristea:
Delaying removal of py35 support would do more damage than good to project mainly due to extra maintenance involved.

Byron:
@Yobmod and @bertwesarg Interestingly now that this branch is merged it fails on CI with failures I can't really fix without spending more time than I have. Maybe it's easy for you to fix due to your experience and I am looking forward to a PR.

In the mean time, I will undo the merge to get main fixed right away. 820d3cc9ceda3e5690d627677883b7f9d349b326 would be the one to revert and take a look at. 
Thanks so much!

Byron:
I reverted the revert…effectively re-adding the removal of python 3.5 as main fails generally now, independently of this commit.
It appears that CI is pulling a more recent version of mypi which now fails where the previous version didn't. Breaking changes seem to be the norm here and we might see this kind of failure more often in the future.

PR #1217, opened 2021-04-21
Remove windows special handling

Andor233:
Remove windows special handling when creating Remote

Python2.5 is no longer supported, and it will cause a problem when init Remote with a no remote git repo.


Byron:
Good catch, thanks for your contribution!

Issue #1223, opened 2021-04-22
[tracking] Unbreaking v3.1.15 (yanked)


Byron:
With the release of v3.1.15 a few breaking bugs were discovered. This caused yanking the version ~14h later.
Here we collect the kind of breakage and track their fixes to allow for a new release.

Until this issue is resolved no new releases can be made.

### Issues

* [x] #1221 `GitCommand missing stderr` - fixed with #1224
* [x] #1220 `GitCommand errors are stringifying bytes in 3.1.15` - fixed with #1222 
* [x]  #1219 `Python 3.5 to 3.8 failing with AttributeError: module 'os' has no attribute 'PathLike'`
  * Even though it cropped up after the release CI doesn't yet pick it up. Typing should be evaluated to catch this.
  
### Causes

My reviews aren't good enough and heavily rely heavily on the test-suite to pass. The latter, however, is far from complete, and the lack of static types makes it hard for tooling to expose issues at compile time.

### Solutions

Please chime in if you have ideas on how to prevent breaking releases in future.

CC: @muggenhor  @jmcgill298

muggenhor:
I'm not sure how exactly to (quickly) address the lack of completeness of the test-suite. But I've got some thoughts on both the test suite and the process of how these bugs managed to slip in:
* your quick response to yank this release is really appreciated and significantly helps to reduce the impact of this in practice for my use cases
* the original PRs was one big massive commit that combined different kinds of refactor work, thus making it next to impossible for any reviewer to review. (FYI: I define a refactoring as a single kind of change to the structure of a program that should not affect behavior at all.) As a result some behavioral changes slipped through. In PRs I'm a reviewer on I often require authors to split up there changes into a "single logical change" per commit. E.g. adding the type annotations should probably not have been combined with moving the locations of function calls. (This might not have prevented the bug from slipping through, but it would have made my bisection for #1221 immediately lead to the fix)

About the test suite:
* #1221 this was caught relatively quickly when I re-ran the test suite of [hopic], so there may be some inspirational test cases in there
* Having a TDD policy, of requiring new tests, for feature/bugfix PRs may help increase test coverage over time
* This may just be me and a reflection of what I work with more often, but I find [pytest] much easier and more convenient to write tests in than [pyunit]. So it took me more time to fit the test case of #1225 in pyunit than to write the basic test itself.

[...]

Issue #1221, opened 2021-04-21
GitCommand missing stderr


muggenhor:
With GitPython 3.1.14 `GitCommandError`, for failed clones would pass on the captured stderr output into its `stderr` attribute. With 3.1.15 this contains the empty string instead.

This was caught by this unit test: https://github.com/tomtom-international/hopic/blob/9a35520388f8109ccff6a89407bc2429ed0d0557/hopic/test/test_checkout.py#L84-L103

A reduced test that reduces this to _only_ testing GitPython (instead of the full integration with hopic) looks like the code below. That exhibits the exact same problem:

```python
import git
import pytest

def test_checkout_in_non_empty_dir(tmp_path):
    orig_repo = tmp_path / "orig"
    with git.Repo.init(orig_repo, expand_vars=False) as repo:
        repo.index.commit(message='Initial commit', **_commitargs)

    non_empty_dir = tmp_path / 'non-empty-clone'
    non_empty_dir.mkdir(parents=True)
    garbage_file = non_empty_dir / 'not-empty'
    garbage_file.write_text('Garbage!')

    # Verify that the clone fails complaining about the target directory not being empty/non-existent
    with pytest.raises(git.GitCommandError, match=r'(?is).*\bfatal:\s+destination\s+path\b.*\bexists\b.*\bnot\b.*\bempty\s+directory\b'):
        git.Repo.clone_from(orig_repo, non_empty_dir)
    assert garbage_file.exists()
```

With 3.1.14 this passes. With the `pytest.raises` expectation removed this output is displayed:
```console
___________________________________________________________________ test_checkout_in_non_empty_dir ____________________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-vanschig/pytest-334/test_checkout_in_non_empty_dir0')

    def test_checkout_in_non_empty_dir(tmp_path):
        orig_repo = tmp_path / "orig"
        with git.Repo.init(orig_repo, expand_vars=False) as repo:
            repo.index.commit(message='Initial commit', **_commitargs)

        non_empty_dir = tmp_path / 'non-empty-clone'
        non_empty_dir.mkdir(parents=True)
        garbage_file = non_empty_dir / 'not-empty'
        garbage_file.write_text('Garbage!')

        # Verify that the clone fails complaining about the target directory not being empty/non-existent
>       git.Repo.clone_from(orig_repo, non_empty_dir)

hopic/test/test_checkout.py:95:
[...]

PR #1225, opened 2021-04-22
test(clone): verify stderr for a failing clone into a non-empty dir

muggenhor:
This adds a test case that would have caught #1221. It thus addresses (partially) the concern of "how to prevent breaking things in the future" expressed in #1223. It additionally verifies that the fix in #1224 actually works.

Edit: this _should_ introduce a single _failing_ test case that should be flipped to passing by #1224.


Byron:
Thanks a lot for your help!

It's good to see this one fail - thus without #1224 there is an issue.
The commit was pushed into said PR which should remain green.

muggenhor:
merged along with #1224

Issue #1219, opened 2021-04-21
Python 3.5 failing with AttributeError: module 'os' has no attribute 'PathLike'


c4xuxo:
This morning our builds started to fail as soon we moved to 3.1.15.

I seems that 3.1.15 makes use of os.PathLike that was introduced in python 3.6.

From README.md seems that Python 3.5 is still supported.
```
REQUIREMENTS
GitPython needs the git executable to be installed on the system and available in your PATH for most operations. If it is not in your PATH, you can help GitPython find it by setting the GIT_PYTHON_GIT_EXECUTABLE=<path/to/git> environment variable.

Git (1.7.x or newer)
Python >= 3.5
```

```
class os.PathLike
An abstract base class for objects representing a file system path, e.g. pathlib.PurePath.

New in version 3.6.
```

Byron:
Thanks for reporting.

That's interesting as I would have thought that our CI can pick this up.
Can it be that PathLike was backported and a more recent version of python 3.5 supports it?

In any case, since python 3.5 is EOL and [this PR](https://github.com/gitpython-developers/GitPython/pull/1201) exists I would probably prefer to cancel python 3.5 support rather than undoing the use of PathLike.

In order to get this fixed, I could yank the offending version. However, first I would love to know why our CI seems to be fine with it. 

Thanks for your help

c4xuxo:
We are using default in Azure Pipelines for python3, that seems to be 3.5 for us.

We just patched to use python 3.6, so we are no longer impacted by 3.5.

Byron:
Good to hear.

Since there is no further information on how to reproduce this issue, as well as no error messages for others to find, I think it's ok to close this issue until further information is presented to help understanding what's going on.

Thanks for your understanding.

muggenhor:
FYI: I think this issue needs to be re-opened. It's still a problem with Python 3.6 too, just a different exception. (See below for details).

> That's interesting as I would have thought that our CI can pick this up.
> Can it be that PathLike was backported and a more recent version of python 3.5 supports it?

Doesn't appear to be the case:
```
$ docker run -it --rm python:3.5-slim
Python 3.5.10 (default, Sep 10 2020, 18:40:42) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from os import PathLike
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'PathLike'
>>> 
```

[...]

PR #1226, opened 2021-04-23
ci: check types with mypy

muggenhor:
This is intended to (partially) address the CI concerns expressed in #1219 and #1223 and will result in _partial_ type checking of the type annotations by using mypy. Keep in mind though that mypy is performing _best-effort static_ analysis in a _dynamically typed_ language so it can only partially check for correctness.

Some other tool(s) will be needed to have more complete type checking at runtime. E.g. [typeguard].

In order to get mypy running successfully this also makes a few (as little as possible) changes to the type annotations. The mypy setting to require annotations on everything is commented out because there's way too much functions lacking annotations right now.

[typeguard]: https://pypi.org/project/typeguard/

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -47,6 +47,11 @@ jobs:
         pip install flake8
         # stop the build if there are Python syntax errors or undefined names
         flake8 --ignore=W293,E265,E266,W503,W504,E731 --count  --show-source --statistics
+    - name: Check types with mypy
+      run: |
+        set -x
+        pip install tox
+        tox -e type
     - name: Test with nose
       run: |
         set -x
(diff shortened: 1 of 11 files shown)


Byron:
Thanks a lot for taking this on! Your effort is much appreciated, and also much needed to get GitPython back on track.

It appears that thanks to never actually running the type checker, the previous commits to add types weren't validated and seemingly aren't working (or are only working with python 3.9).

I have no idea how much additional work it would be to get mypy to pass, but if it's a dead-end it would certainly be alright to give up on types altogether and revert the commits that added them.

It's all up to you.

muggenhor:
Combined with #1227 this should pass now.

muggenhor:
Rebased (to force test-rerun) after #1227 got merged.

Unfortunately it looks like the 3.9 build failed due to what looks like an intermittent failure. And I cannot retrigger a build without a force push....

Byron:
Thanks a million!

PR #1227, opened 2021-04-23
fix(types): get the os.PathLike type as correctly as possible

### LINKED ISSUE #1219: Python 3.5 failing with AttributeError: module 'os' has no attribute 'PathLike'

muggenhor:
This should make our internal PathLike type compatible with Python < 3.6 and < 3.9.

Fixes #1219

### DIFF
--- a/git/types.py
+++ b/git/types.py
@@ -1,6 +1,20 @@
-import os  # @UnusedImport  ## not really unused, is in type string
+# -*- coding: utf-8 -*-
+# This module is part of GitPython and is released under
+# the BSD License: http://www.opensource.org/licenses/bsd-license.php
+
+import os
+import sys
 from typing import Union, Any
 
 
 TBD = Any
-PathLike = Union[str, 'os.PathLike[str]']
+
+if sys.version_info[:2] < (3, 6):
+    # os.PathLike (PEP-519) only got introduced with Python 3.6
+    PathLike = str
+elif sys.version_info[:2] < (3, 9):
+    # Python >= 3.6, < 3.9
+    PathLike = Union[str, os.PathLike]
+elif sys.version_info[:2] >= (3, 9):
+    # os.PathLike only becomes subscriptable from Python 3.9 onwards
+    PathLike = Union[str, os.PathLike[str]]


muggenhor:
Switched away from `try-except` to explicit version checks because that seems to be the recommended style python/mypy#698

Byron:
Thanks, very elegant!

I find it interesting that all PathLike types seem to use `str` in some way even though AFAIK these are to be decoded with some encoding from what the OS provides. As this encoding is unspecified or at least very dependent on knowledge about the OS and its settings, using a path based on `str` can fail with some sort of decoding error.

muggenhor:
FYI: this is the [type annotation definition of `os.PathLike`](https://github.com/python/typeshed/blob/27facc7ff94df4bbdf01dc57791ee013a220a807/stdlib/builtins.pyi#L962-L968). So what `os.PathLike[str]` means is: it has to be a type with an `__fscache__(self) -> str` method. So that's the method where any conversions should be performed.

muggenhor:
> I find it interesting that all PathLike types seem to use `str` in some way even though AFAIK these are to be decoded with some encoding from what the OS provides. As this encoding is unspecified or at least very dependent on knowledge about the OS and its settings, using a path based on `str` can fail with some sort of decoding error.

Hmm. Does that mean you think it should accept `bytes` in some locations too?

Byron:
> Hmm. Does that mean you think it should accept bytes in some locations too?

I would refrain from doing anything that reduces convenience or being idiomatic or that is breaking just to gain correctness.
But I think that as it stands, GitPython will work fine until it doesn't because one tries to handle a path with characters that don't decode with the pre-configured default encoding. Right now this is UTF-8, certainly a decent choice. But it won't always work either.

To me the whole encoding topic in GitPython is a lost cause as correctness could (to my mind) only be achieved if the API doesn't use strings (with encoding) to handle them. 

And by the looks of it, it's idiomatic to do that, too:

```python
# py 3.8
>>> type(os.listdir('.')[0])
<class 'str'>
```

PR #1234, opened 2021-05-12
Add types to git/config.py

Yobmod:
Adding types to config.py and cmd.py
Functions annotated all listed in commit titles.

I had to use a few casts in this one, due to the use of hassattr to distinguish strings from file objects. 
If anyone knows a better way, let me know. (Its an open issue on mypy, and Guido recomends rewriting and doing type checks - a job for the future!) https://github.com/python/mypy/issues/1424#issuecomment-223795437

Some minor logic changes, all for type checks (e.g. because str is a subtype of Sequence, have to check for it before other sequences) or to try to separate strings from bytes.

All check pass and mypy happy. I also ran some basic scripts against it, and no errors.

Might be worth cutting a release after this PR, in case an errors are hiding?
(No rush, I only submitted this one so fast as it was waiting in my fork already)

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -40,6 +40,18 @@ from .util import (
     stream_copy,
 )
 
+# typing ---------------------------------------------------------------------------
+
+from typing import TYPE_CHECKING
+
+from git.types import TBD
+
+if TYPE_CHECKING:
+   pass
+
+
+# ---------------------------------------------------------------------------------
+
 execute_kwargs = {'istream', 'with_extended_output',
                   'with_exceptions', 'as_process', 'stdout_as_string',
                   'output_stream', 'with_stdout', 'kill_after_timeout',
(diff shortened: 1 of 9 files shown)


Byron:
Thanks for the additional explanation - it definitely helps. My workflow is to checkout the PR and try the types locally, possibly changing some types to see what mypy thinks about that just to realize that the types presented here are the one (or only) way to make it happy.

In the very first file fo the diff `safe_decode` also can be called with None which is strange to see, but currently required due to call sites not properly checking what they try to decode. Very interesting, and a lot of potential for future improvements.

Maybe I can also direct your attention to #1235 in case there is a chance it might be related to typing.

In any case, I will cut a release momentarily.

_And last but not least: Thank you for all the work you put into GitPython types!_ They should help making the codebase more maintainable.

PR #1244, opened 2021-05-15
Added types to Index submodule

Yobmod:
Added types to Index submodule,
plus change some types in othe files to make them agree

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -3,6 +3,7 @@
 #
 # This module is part of GitPython and is released under
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
+from git.refs.reference import Reference
 import glob
 from io import BytesIO
 import os
@@ -65,7 +66,8 @@ from .util import (
 
 # typing -----------------------------------------------------------------------------
 
-from typing import Any, Callable, Dict, IO, Iterator, List, Sequence, TYPE_CHECKING, Tuple, Union
+from typing import (Any, Callable, Dict, IO, Iterable, Iterator, List,
+                    Sequence, TYPE_CHECKING, Tuple, Union)
 
 from git.types import PathLike, TBD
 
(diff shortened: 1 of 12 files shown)


d066zkie (approved):

d066zkie (approved):

d066zkie (approved):

Byron:
@Yobmod Thanks a lot! Do you know how it's possible that somebody I didn't assign can approve changes and do so repeatedly? Is that a GitHub bug? Should I be concerned?

Yobmod:
Oh, don't know how that is possible (or who the user is).

It's never happened until I used the draft feature. Maybe something to do with that?

git/index/base.py
```diff
@@ -446,20 +473,21 @@ def iter_blobs(self, predicate=lambda t: True):
...
         # END for each entry
 
-    def unmerged_blobs(self):
+    def unmerged_blobs(self) -> Dict[PathLike, List[Tuple[StageType, Blob]]]:
         """
         :return:
             Iterator yielding dict(path : list( tuple( stage, Blob, ...))), being
             a dictionary associating a path in the index with a list containing
             sorted stage/blob pairs
+            ##### Does it return iterator? or just the Dict?
```
Byron, line 482: It really does look like it returns a dict, and the type system probably agrees.
In these cases it should be fine to fix the documentation, which rightfully is confused about types in a codebase that thus far didn't have any statically declared.

Byron (approved):
Thanks a lot for yet another fantastic addition of types. 

All changes to the logic where based on the type-system recognising that some fields or variables might be None, so I wouldn't expect any issues arising from this.

Byron:
Please let me know if this is ready for merging, the PR appears to be in draft mode still.

Yobmod:
Now ready for review

git/index/base.py
Yobmod: ok, i changed the docstring then

Byron:
This looks good to me, thanks a lot!

Issue #1233, opened 2021-05-09
Progression not functional once a single non-parsable line was encountered


mxrch:
To run the script properly, just create a creds.py file with a token variable, and a .tmp folder
```python

import creds

from git import Repo, RemoteProgress
from pathlib import Path

import uuid


username = "gitpython-developers"
repo_name = "GitPython"
uuid = uuid.uuid4().hex[:20]

# ---------------

class Progress(RemoteProgress):
    def update(self, op_code, cur_count, max_count=None, message=''):
        print(f"{op_code} {cur_count} {max_count} {message}")

repo_url = f'https://{creds.token}:x-oauth-basic@github.com/{username}/{repo_name}'
repo_path = Path('.tmp') / Path(uuid)
print(f'Cloning the commits history of {repo_name}...')
err = None
try:
    Repo.clone_from("https://github.com/gitpython-developers/GitPython", repo_path, progress=Progress(), multi_options=["--filter=tree:0"])
    #Repo.clone_from(repo_url, repo_path, progress=Progress(), multi_options=["--filter=tree:0"])
except Exception as error:
    err = error
    print(err)
```

If you comment the 1st Repo.clone_from and uncomment the 2nd, the progress will not work anymore, even though the progress is showing up using the git binary.

Byron:
Thanks a lot for putting in the time to make this issue reproducible. With the latest GitPython from `main` and tested with my own token, the following code produced progress messages just like the one above.

```python
import creds

from git import Repo, RemoteProgress
from pathlib import Path

import uuid


username = "gitpython-developers"
repo_name = "GitPython"
uuid = uuid.uuid4().hex[:20]

# ---------------

class Progress(RemoteProgress):
    def update(self, op_code, cur_count, max_count=None, message=''):
        print(f"{op_code} {cur_count} {max_count} {message}")

repo_url = f'https://{creds.token}:x-oauth-basic@github.com/{username}/{repo_name}'
repo_path = Path('.tmp') / Path(uuid)
print(f'Cloning the commits history of {repo_name}...')
err = None
try:
    #Repo.clone_from("https://github.com/gitpython-developers/GitPython", repo_path, progress=Progress(), multi_options=["--filter=tree:0"])
    Repo.clone_from(repo_url, repo_path, progress=Progress(), multi_options=["--filter=tree:0"])
except Exception as error:
    err = error
    print(err)
```

[...]

PR #1248, opened 2021-05-21
fixed case where progress was no longer shown if a single error occured

mxrch:
it fixes the issue #1233


Byron:
Thanks a lot! I am optimistic that this change will only reduce the amounts of surprises people get when trying to display progress.

PR #1264, opened 2021-06-03
Issue 1261

### LINKED ISSUE #1261: repo.tag() only accepts 'ref/tags/tagname' instead of just 'tagname'
The param comment for `tag()` in `base.py` mentions `0.1.5` and `tags/0.1.5` as valid inputs, however the call to `TagReference` calls the inherited constructor (`Reference` in reference.py) with the default `check_path=True`, which returns a `ValueError`, if the tag/path given doesnt start with `self._common_path_default + '/'`.

Either the param comment needs to be adjusted or the tag parameter cleaned up so that it only returns the tag value to `Reference.__init__`.

pycharm64_UlSdzGazov
pycharm64_nnw8DMULi2

bytefluxio:
Fixes #1261

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,4 +43,5 @@ Contributors are:
 -Liam Beguin <liambeguin _at_ gmail.com>
 -Ram Rachum <ram _at_ rachum.com>
 -Alba Mendez <me _at_ alba.sh>
+-Robert Westman <robert _at_ byteflux.io>
 Portions derived from other open source works and are clearly marked.
--- a/git/refs/tag.py
+++ b/git/refs/tag.py
@@ -18,7 +18,8 @@ class TagReference(Reference):
         print(tagref.tag.message)"""
 
     __slots__ = ()
-    _common_path_default = "refs/tags"
+    _common_default = "tags"
+    _common_path_default = Reference._common_path_default + "/" + _common_default
 
     @property
     def commit(self):
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -406,7 +406,7 @@ class Repo(object):
         return TagReference(self, full_path)
 
     @staticmethod
-    def _to_full_tag_path(path: PathLike):
+    def _to_full_tag_path(path):
         if path.startswith(TagReference._common_path_default + '/'):
             return path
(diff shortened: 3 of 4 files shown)


bytefluxio:
I initially wanted to add [parameterized](https://github.com/wolever/parameterized), but there was an issue somewhere, which is why I decided to just collect the errors

Byron:
Thanks a lot!

With `tox -e type` you should be able to pacify the type checker. It looks like it's not sure `startswith` actually is  available.

bytefluxio:
I come from a strongly typed language, so this caught me unawares.

since reference.py doesn't do a PathLike check, I removed mine, too.

reference.py:
```
    def __init__(self, repo, path, check_path=True):
        """Initialize this instance
        :param repo: Our parent repository

        :param path:
            Path relative to the .git/ directory pointing to the ref in question, i.e.
            refs/heads/master
        :param check_path: if False, you can provide any path. Otherwise the path must start with the
            default path prefix of this type."""
        if check_path and not path.startswith(self._common_path_default + '/'):
            raise ValueError("Cannot instantiate %r from path %s" % (self.__class__.__name__, path))
        super(Reference, self).__init__(repo, path)
```

bytefluxio:
Weird, that my tests ran through without issues on Python 3.9.5 ¯\_(ツ)_/¯

Didn't use tux though, since I'm new to python, I had no idea about it.

Byron:
> I come from a strongly typed language, so this caught me unawares.

I stopped trying to understand or follow or write python many years ago, too, and I am blissfully unaware of how the typing really works. Removing types is quite alright here I think and more knowledgable individuals can add them later if they see benefits.

Anyway, with a tiny adjustment it worked as expected. Thanks a lot for your contribution - thanks to you GitPython finally does as it says on the packaging (in this instance) 10 years or so later 😅.

bytefluxio:
Thanks for the test fix.

I fixed that with "Fixes test to not throw false negative results" 057514e85bc99754e08d45385bf316920963adf9

But threw it away when I "Reverts auto format introduced with 2dbc2be" 79e24f78fa35136216130a10d163c91f9a6d4970

Shame on me

Issue #1270, opened 2021-06-14
Typing-related Exception in Python 3.5


luator:
With the latest version from PyPi I'm getting an exception in Python 3.5 which seems to be related to type hints

### Steps to reproduce
```
python3 -m venv venv_gitpython
. venv_gitpython/bin/activate
pip install gitpython
python3 -c "import git"
```

### Expected behaviour
Import succeeds without error

### Actual behaviour
Import fails with
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/felixwidmaier/venv_gitpython/lib/python3.5/site-packages/git/__init__.py", line 42, in <module>
    from git.config import GitConfigParser  # @NoMove @IgnorePep8
  File "/home/felixwidmaier/venv_gitpython/lib/python3.5/site-packages/git/config.py", line 238, in <module>
    class GitConfigParser(with_metaclass(MetaParserBuilder, cp.RawConfigParser, object)):  # type: ignore ## mypy does not understand dynamic class creation # noqa: E501
  File "/home/felixwidmaier/venv_gitpython/lib/python3.5/site-packages/git/config.py", line 279, in GitConfigParser
    repo: Union['Repo', None] = None) -> None:
  File "/usr/lib/python3.5/typing.py", line 552, in __getitem__
    dict(self.__dict__), parameters, _root=True)
  File "/usr/lib/python3.5/typing.py", line 512, in __new__
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 512, in <genexpr>
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/home/felixwidmaier/venv_gitpython/lib/python3.5/site-packages/typing_extensions.py", line 635, in __subclasscheck__
    raise TypeError("Literal cannot be used with issubclass().")
TypeError: Literal cannot be used with issubclass().
```

Tested with Python 3.5.2.  The error does not occur with 3.6.9.

I saw that there are already plans to drop support for 3.5 soon, so unless this is an easy-to-fix issue, it's probably not worth spending time on it.  However, in this case I think it would be good to officially drop 3.5-support now to avoid confusion.

luator:
Note for others still stuck with Python 3.5 for whatever reason:  3.1.14 is the latest version that still seems to work.

Byron:
[...]

PR #1271, opened 2021-06-17
Add initial types to object, and fix CI

Yobmod:
Adds initial types to object submodule.

Fixed CI issue that arose due to mypy/typeshed updating: it didn't like IterableList overriding list.get_item() argument types, which were changed in typeshed). Added imports for SupportsIndex with python version checks (added in py3.8).

Also removed a python 3.5 from setup.py that got missed for #1270.


Byron:
Thanks a million, it's great to see CI back online and of course be able to make the 3.1.18 release :).

PR #1279, opened 2021-06-25
Finish typing object, improve verious other types.

Yobmod:
This one finished adding types to tree.py so objects folder is done.

Also:

-Add Typedict import and used to improve some Dict types.
-Add Typeguard (Pep647) and start to replace cast() and type: ignores with runtime checks and type narrowing.
-Change typing-extentions dependancy to <py3.10, for Typeguard
-Make IterableList generic and fix all uses to be IterableList[Iterable-type], e.g. IterableList[Head]
-Rename Iterable to IterableObj and changed throughout, due to naming conflic with typing.Iterable. I left git.util.Iterable in place for now and added a deprecationwarning in case users are using it themselves for some reason.

Ready to merge


Byron:
The work you are doing here is monumental and I am impressed by the consistency with which you chip off little pieces one by one until the mountain is gone.

I also hope that when it's all done you will be open to become a maintainer to keep GitPython operational, I have a feeling that I won't be able to do it in case something breaks when new versions of something break backwards compatibility.

PR #1280, opened 2021-06-30
Added clone multi_options to Submodule

Lakhtenkov-iv:
The purpose of the change is to add multiple_options for clone in Submodule simmilar as it is done in the Repo.clone().


Byron:
Thanks a lot, it's great to see some tests for this feature, too!

PR #1282, opened 2021-06-30
Start adding types to Submodule, add py.typed to manifest

Yobmod:
Started to add types to Submodule

Removed include_package_data() from setup and added py.typed to MANIFEST.INI instead (this is what mypy package does, but not what the mypy instructions say!)

Improve types for .traverse() with TypeVar and overloads, create TraversableIterableObj class to reduce code duplication from overloads.

Add some more checks to test_commit.test_traverse()


setup.py
```diff
@@ -95,7 +95,7 @@ def build_py_modules(basedir, excludes=[]):
     license="BSD",
     url="https://github.com/gitpython-developers/GitPython",
     packages=find_packages(exclude=("test.*")),
-    package_data={'git': ['**/*.pyi', 'py.typed']},
+    # package_data={'git': ['**/*.pyi', 'py.typed']},
```
Byron, line 98: Is this something that can be removed?

Byron (changes requested):
That's great work, thanks a lot!

It's good to see that people are already starting to use it, too.

Can't wait to merge.

git/objects/submodule/base.py
```diff
@@ -363,8 +382,8 @@ def add(cls, repo, name, path, url=None, branch=None, no_checkout=False, depth=N
 
         # assure we never put backslashes into the url, as some operating systems
         # like it ...
-        if url is not None:
-            url = to_native_path_linux(url)
+        # if url is not None:
+        #     url = to_native_path_linux(url)   to_native_path_linux does nothing??
```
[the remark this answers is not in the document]
Yobmod: Whoops, I've uncommented it!

setup.py
Yobmod: As far as I can tell, we should use package_data OR use_package_data.
Use_package_data=True looks at MANIFEST.INI for files and ignores files listed in package_data.
(like here: https://github.com/encode/httpx/issues/193#issue-477224461)

So as we already have use_package_data=True, i've moved the py.typed entry to MANIFEST.INI.
I'll try a fresh install from github once merged and see if it then picks up the py.typed.

Byron: Ok, I see that this is to be removed once it's clear that the MANIFEST.INI  will work as expected also in conjunction with 'py.typed'.

Byron:
Thanks a million for all your work!
It's clear you are committed to getting this done and I would be glad if you would want to become a maintainer. Now that types have been added and it feels like the project would be in serious trouble without your continued support in keeping maintaining types as python evolves along with the type checkers.
Thanks a lot for your consideration.

PR #1285, opened 2021-07-09
Finish initial typing of Index and Submodule

Yobmod:
* Finished adding initial types to Index and Submodule modules
* Gone back to try and improve types in various places, especially git/diff.py objects.tree and Traversable
* Renamed 'diff' in various places (I think none user facing) to distingish the submodule from the function and the variables.
* Added some TypeGuards (is_config_level()) and Protocols (Has_Repo, Has_id_attribute)
* Removed 'out_append = out.append' in a couple of places, just because they prevent getting info on hover in IDEs
* Made IterableList ___getattr___ return type same as the list was initialised with, , which is not strictly correct but better than Any.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -474,8 +474,8 @@ class Submodule(IndexObject, TraversableIterableObj):
                 sm._branch_path = br.path
 
         # we deliberately assume that our head matches our index !
-        if mrepo is not None:
-            sm.binsha = mrepo.head.commit.binsha
+        mrepo = cast('Repo', mrepo)
+        sm.binsha = mrepo.head.commit.binsha
         index.add([sm], write=True)
 
         return sm
@@ -652,7 +652,7 @@ class Submodule(IndexObject, TraversableIterableObj):
                 may_reset = True
                 if mrepo.head.commit.binsha != self.NULL_BIN_SHA:
                     base_commit = mrepo.merge_base(mrepo.head.commit, hexsha)
-                    if len(base_commit) == 0 or (base_commit[0] is not None and base_commit[0].hexsha == hexsha):
+                    if len(base_commit) == 0 or base_commit[0].hexsha == hexsha:  # type: ignore
                         if force:
                             msg = "Will force checkout or reset on local branch that is possibly in the future of"
                             msg += "the commit it will be checked out to, effectively 'forgetting' new commits"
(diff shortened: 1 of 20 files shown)


Yobmod:
I had a question about Traversable.list_traverse().
Currently, it transparently passes any args to self.traverse(), which means it can return a tuple of each node when as_edge=True.
But an IterableList of tuples isn't useful as has no _id_attribute_ to select them with, and each tuple will be half redundant info.

Should I hard-code list_traverse to be as_edge=False? Or I could change it to return a normal List of tuples.

Byron:
> I had a question about Traversable.list_traverse().
> Currently, it transparently passes any args to self.traverse(), which means it can return a tuple of each node when as_edge=True.
> But an IterableList of tuples isn't useful as has no _id_attribute_ to select them with, and each tuple will be half redundant info.
> 
> Should I hard-code list_traverse to be as_edge=False? Or I could change it to return a normal List of tuples.

I think if it’s not a breaking change you should do whatever seems right. If it must be a breaking change a major version bump could be done, but I would love to avoid it.

This API suffers from being overly abstract, it worked in a few cases but was kept even when it was out of place :/.

Yobmod:
ok, ive typed it as though as_edge=False with a type: ignore for now.
So if people override the default they'll just get wrong type hints.
I'll look at adding overloads in the future, but it would need a lot of them to really be correct.

Otherwise this is ready to merge. Just Refs to go!

Re: adding me as a a maintainers. Thats fine by me. 
But I don't really know much about Git, thats why I use Gitpython in the first place, lol

git/objects/submodule/base.py
```diff
@@ -1065,13 +1079,14 @@ def rename(self, new_name):
             destination_module_abspath = self._module_abspath(self.repo, self.path, new_name)
             source_dir = mod.git_dir
             # Let's be sure the submodule name is not so obviously tied to a directory
-            if destination_module_abspath.startswith(mod.git_dir):
+            if str(destination_module_abspath).startswith(str(mod.git_dir)):
```
Byron, line 1082: I understand that the type checker wanted add these conversions here, but is there a way to do without them?
It's just that `git_dir` is a directory and its value might not be decodable by the encoding enforced by `str`.

It would be great if you could give it another look and if there is an alternative, you could push that to `main` directly.

Byron (approved):
This is tremendous work, thanks again! And I admit I could only skim through all changes and hope for some pattern matching to kick in.

The changes related to setting `as_edge` to false seem to be done for real, but also nothing seems to break in the test suite so I am hopeful it truly doesn't affect anything.

I am looking forward to seeing this work finished, apparently there is just the _refs_ part to go.

Byron:
> But I don't really know much about Git, thats why I use Gitpython in the first place, lol

But you know _everything_ about the types added here :), while having interest to keep the types working for yourself. As I would probably be overwhelmed when trying to fix issues arising from it in future, you are now the major force in keeping this project maintained.

PR #1289, opened 2021-07-12
Tidying up, and replace nose with pytest

### LINKED ISSUE #1287: Nose unsupported, doesnt work with Python 3.10
3 months until 3.10 is released (Oct 2021), so i've just tried run the test suite on 3.10.0beta4.

Didn't get very far, as Nose does not work on 3.10. The issue was reported during 3.9alpha release, so not going to be fixed or released: https://github.com/nose-devs/nose/issues/1099.

Finally time to move to pytest? Then can use all the plugins! Seems is as easy to move to pytest as Nose2.
If anyone wants to help with that..... ('ve never used Nose before).
https://github.com/pytest-dev/nose2pytest might help?

Yobmod:
Fix #1287
Some tidying up for config files.
All I wanted was a linter that could report on imports only needed for typing! 
Anyway...

**setup.py**
Change distutils build_py to setuptools wrapper. Deprecated since py3.8, I thought it caused a problem with 3.10beta. Now works py3.6-3.10.

**Tox.ini**
Ignore E704 (mutiple expression per line) so can make @overloads neater
Add [pytest] config section

**requirements**
Created requirements-dev.txt with pytest and plugins
Edit: Changed nose to pytest in test-requirements.txt

**Pythonpackage.yaml**
Add pip install wheel and setuptools, newer packages were having to fallback to EZ_setup without them.
Invoke mypy directly, no need for tox. 
Add py3.10.0-Beta.3 to test matrix....
....which needed update to actions/setup-python@v2
Add pytest step - runs after Documentation and set to always pass, but shows errors in info
Edit: Swapped nose for pytest

**Results** 
Pytest results: 378 passed, 42 skipped, 3 warnings
Nose results: 420 tests, 42 skipped (= 378)

So we might just be able to switch directly to Pytest to run the tests.
But test code still needs to be gone through to remove any nose dependancy for py3.10. 
Edit: if there is any - all looks like unittest to me?
Edit2: So ive removed nose everywhere, and all same tests pass or fail or are skipped!


Byron:
Fantastic work once again! I remember that now being able to use pytest was a hurdle for some contributors, I am glad that's removed now.

Maybe you could also have a look at the [respective section in the docs](https://github.com/gitpython-developers/GitPython#running-tests) and see if a command-line without `tox` could be suggested there as well.

There also is a [Makefile](https://github.com/gitpython-developers/GitPython/blob/main/Makefile) that could be loaded up with useful targets for typical local development, to help running tests and linters for example.

PR #1295, opened 2021-07-19
Add types to refs

Yobmod:
Adding types to refs module, and backtracking to make sure they agree with previous types.

I also changed the hard-coding of Traversable.list_traverse(on_edge= True) that I put in last time to be a check for it in kwargs. At the moment all it does is raise an additional deprecation warning if anyone tried to use on_edge= True - I don't think anyone would be, but better than breaking it without warning.
Also made Traversable, Serilizable and IterableObj into Protocols (typings version of abstract base classes) - shouldnt have any effect on users unless they are doing something wierd.

Added flake8-bugbear and flake8-comprehensions and fixed the (minor) complaints they had.

Removed some checks for python 3.5 in tests.


Byron:
Tremendous work, thank you!

PR #1298, opened 2021-07-24
Revert use of Typeguard and therefore typing-extensions==3.10.0.0

Yobmod:
* Downgrade typing-extensions back to 3.7.4.3 and remove Typeguard.
Typing-extensions 3.10.0.0 keeps giving problems even after the requirements.txt updated. Its been out for 3 months, but installing from Testpypi says it doesnt exist (have to install it manually first), and lots of libraries are still fixed to 3.7.4.x (including tensorflow and VScode's restructured text linter?!)
I've commented out all the function defs that use Typeguard, so they can be put back once the world updates. In the meantime, i replaced them all with 'cast' to keep mypy happy.

* Added more complex typing to BaseIndexEntry, SectionConstraint, GitConfigParser, _OMD.

* Rmved with_metaclass shim, not needed for python 3.

* Already includes the changes from previous PR that fixed the py.typed issue in setup.py.

#1095

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,7 +20,7 @@ ignore = E265,E266,E731,E704,
          W293, W504,
          ANN0 ANN1 ANN2,
          TC002,
-         # TC0, TC1, TC2
+         TC0, TC1, TC2
          # B,
          A,
          D,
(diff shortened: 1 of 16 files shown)


Byron:
Thank you!

I will create a new release on your mark.

PR #1299, opened 2021-07-27
Added support of spaces for clone multi_options

Lakhtenkov-iv:


Byron:
Thanks a lot, it's great to see it has a test, too.

Lakhtenkov-iv:
Thanks @Byron for review.
Do you have ETA when you are going to release `3.1.20` that would fix yanked `3.1.19`?

Byron:
The next release happens once @Yobmod gives it a go, and I would expect that to happen during the weekend at the latest.

Yobmod:
I've checked the newest version install's on windows, and can create a repo
and push/pull.
so ready for release.

I'm thinking maybe release it as a .rc (or .dev or .beta), and see if any
errors reported for a week or 2? (so it won't automatically install on
update for downstream)
But up to you @byron

Byron:
Thanks a million! As long as people are not showing up with pitchforks I suppose it's OK to release 3.1.20 and get feedback fast, possibly yanking the release if necessary. Those who can't have that will have pinned their version by now.

Since you have done so much work testing and found quite a few issues on the way I am also confident that there won't be any obvious flaws or side-effects this time.

v20 is out, let's hold our breath, I will keep the finger on the yank button, too :D.

PR #1302, opened 2021-07-30
Fix typing of Head.create_head

dimbleby:
After creating a `Head` via `Head.create_head()`, mypy complains if I try to do head-like things with it:

```
error: "SymbolicReference" has no attribute "checkout"  
error: "SymbolicReference" has no attribute "set_tracking_branch"
```


Yobmod:
Hi, thanks for the fix!

PR #1301, opened 2021-07-28
Hopefully fixes Orderedict error

Yobmod:
Hopefully fixes Orderedict error for py<3.7.2 and py>3.10 
Talk about an unstable api!

I added type: ignores to pass mypy for 3.6, and removed the version guard for py<3.10 (typing.OrderedDict gets deprecated in favor or collections.Ordereddict again, but it will still work until 3.12 at least, so can worry about it once 3.6 dropped. Or I guess once 3.7.1 dropped really.

Also added some types that got missed in Symbolic.py

#1300 #1095

Testing on GitHubActions appears to be broken? The last commits all stuck ("Can't find any online and idle self-hosted or hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-latest'")

PR #1303, opened 2021-07-31
Update ref types

Yobmod:
Split #1301 to find why CI not working.

-Improve types of refs in symbolic.py, head.py, tag.py, repo/base.py


Yobmod:
CI still broken

PR #1309, opened 2021-08-02
Add flake8-typing-imports

kasium:
The idea to include this plugin comes from #1095 . In short, flake8-typing-imports checks if a import from `typing` is possible in a certain python version.

I ran the plugin in your code base, and it's **NOT** compatible with <= 3.6.2. You use NamingTuple with defaults and NoReturn which were added in later patches of 3.6. To let in still run I set the minimum version to 3.7.0.

This means, that either we should fix the issues or gitpython needs to drop the support for this version which is anyway deprecated.
This plugin will hopefully prevent that any new issues are added

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -31,3 +31,5 @@ exclude = .tox,.venv,build,dist,doc,git/ext/,test
 
 rst-roles =  # for flake8-RST-docstrings
     attr,class,func,meth,mod,obj,ref,term,var  # used by sphinx
+
+min-python-version = 3.7.0
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,6 +4,7 @@ mypy
 flake8
 flake8-bugbear
 flake8-comprehensions
+flake8-typing-imports
 
 virtualenv
 


kasium:
Output for 3.6.0

```
git/types.py:7:1: TYP001 guard import by `if TYPE_CHECKING:`: NoReturn (not in 3.6.0, 3.6.1)
from typing import (Callable, Dict, NoReturn, Sequence, Tuple, Union, Any, Iterator,       # noqa: F401
^
git/index/base.py:70:1: TYP001 guard import by `if TYPE_CHECKING:`: NoReturn (not in 3.6.0, 3.6.1)
from typing import (Any, BinaryIO, Callable, Dict, IO, Iterable, Iterator, List, NoReturn,
^
git/index/typ.py:69:5: TYP005 NamedTuple does not support defaults in 3.6.0
    ctime_bytes: bytes = pack(">LL", 0, 0)
    ^
git/index/typ.py:70:5: TYP005 NamedTuple does not support defaults in 3.6.0
    mtime_bytes: bytes = pack(">LL", 0, 0)
    ^
git/index/typ.py:71:5: TYP005 NamedTuple does not support defaults in 3.6.0
    dev: int = 0
    ^
git/index/typ.py:72:5: TYP005 NamedTuple does not support defaults in 3.6.0
    inode: int = 0
    ^
git/index/typ.py:73:5: TYP005 NamedTuple does not support defaults in 3.6.0
    uid: int = 0
    ^
git/index/typ.py:74:5: TYP005 NamedTuple does not support defaults in 3.6.0
    gid: int = 0
    ^
git/index/typ.py:75:5: TYP005 NamedTuple does not support defaults in 3.6.0
    size: int = 0
    ^
git/objects/util.py:25:1: TYP001 guard import by `if TYPE_CHECKING:`: Deque (not in 3.6.0)
from typing import (Any, Callable, Deque, Iterator, NamedTuple, overload, Sequence,
^
git/refs/remote.py:12:1: TYP001 guard import by `if TYPE_CHECKING:`: NoReturn (not in 3.6.0, 3.6.1)
from typing import Any, Iterator, NoReturn, Union, TYPE_CHECKING
^
4     TYP001 guard import by `if TYPE_CHECKING:`: NoReturn (not in 3.6.0, 3.6.1)
7     TYP005 NamedTuple does not support defaults in 3.6.0
11
```

Yobmod:
I've set the checks to run :)
Seems useful already!

So we have to decide: Drop 3.6 now, or i'll pull out NamedTuple defaults (there are only a couple of uses) for now.
NoReturn is available in typing_extensions, so i'll fix that.

I was planning to wait until 3.10 released to drop 3.6.
But I see it had it's final release this June and  i have noticed big packages already dropping it (numpy and pandas and therefore most of scientific packages).

PR #1311, opened 2021-08-02
Drop 3.6, increase type strictness

Yobmod:
- Drop 3.6, per discussion.

- Add some mypy configs to increase strictness

- Type fixes for stricter config.

- Replace some Any types with the runtime results from tests

- Try out `__future__.annotations` and `|` union syntax

Ready for next release I think, except needs to be tested against 3.7.0 or 3.7.1 on linux
(I'm on Windows atm, so tests are dodgy).


doc/source/intro.rst
```diff
@@ -13,15 +13,17 @@ The object database implementation is optimized for handling large quantities of
 Requirements
 ============
 
-* `Python`_ >= 3.6
+* `Python`_ >= 3.7
 * `Git`_ 1.7.0 or newer
     It should also work with older versions, but it may be that some operations
     involving remotes will not work as expected.
 * `GitDB`_ - a pure python git database implementation
+* `typing_extensions`_ >= 3.10.0
```
kasium, line 21: According to requirements.txt this should be "typing-extensions>=3.7.4.3;python_version<"3.10""

setup.py
```diff
@@ -114,9 +115,9 @@ def build_py_modules(basedir, excludes=()):
         "Operating System :: MacOS :: MacOS X",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9"
+        "Programming Language :: Python :: 3.10"
```
kasium, line 121: You can also add "Typing :: Typed"

Yobmod: Good idea, added it, and fixed intro

PR #1314, opened 2021-08-03
Do not call get_user_id if it is not needed

### LINKED ISSUE #356: Windows : ImportError: No module named pwd on util.py
When i try to build gitpython on Windows Server i have this error message.

```powershell
File "C:\project\.tox\back\lib\site-packages\git\objects\commit.py", line 318, in create_from_tree
    committer = committer or Actor.committer(cr)
  File "C:\project\.tox\back\lib\site-packages\git\util.py", line 416, in committer
    return cls._main_actor(cls.env_committer_name, cls.env_committer_email, config_reader)
  File "C:\project\.tox\back\lib\site-packages\git\util.py", line 386, in _main_actor
    default_email = get_user_id()
  File "C:\project\.tox\back\lib\site-packages\git\util.py", line 152, in get_user_id
    return "%s@%s" % (getpass.getuser(), platform.node())
  File "c:\python27\Lib\getpass.py", line 157, in getuser
    import pwd
ImportError: No module named pwd
```

After search, the only solution is try [this hack](https://t37.net/on-getting-the-running-process-user-under-python-and-supervisord.html) on [utiy.py](https://github.com/gitpython-developers/GitPython/blob/master/git/util.py#L14) file.

Then use this :  

```python
try:
    import pwd
except ImportError:
    import winpwd as pwd
import os

def get_user_id():
    """:return: string identifying the currently active system user as name@node"""
    return "%s@%s" % (pwd.getpwuid(os.getuid()).pw_name, platform.node())
```

instead of 

```python
import getpass

def get_user_id():
    """:return: string identifying the currently active system user as name@node"""
    return "%s@%s" % (getpass.getuser(), platform.node())
```

What do you think about that ?

eric-wieser:
On systems without any environment variables (such as sandboxed CI runs) and no `pwd` module (such as windows), GitPython crashes.

This is just a logic error - there is no reason to fallback to `pwd` until the config lookup has actually failed. The implementation of `ConfigWriter.get_value` is to catch `Exception`, so this patch doesn't attempt to work out a more specific exception than what was used before.

This fixes #356

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -217,8 +217,23 @@ class TestUtils(TestBase):
             self.assertIsInstance(Actor.author(cr), Actor)
         # END assure config reader is handled
 
+    @with_rw_repo('HEAD')
     @mock.patch("getpass.getuser")
-    def test_actor_get_uid_laziness_not_called(self, mock_get_uid):
+    def test_actor_get_uid_laziness_not_called(self, rwrepo, mock_get_uid):
+        with rwrepo.config_writer() as cw:
+            cw.set_value("user", "name", "John Config Doe")
+            cw.set_value("user", "email", "jcdoe@example.com")
+        
+        cr = rwrepo.config_reader()
+        committer = Actor.committer(cr)
+        author = Actor.author(cr)
+        
+        self.assertEqual(committer.name, 'John Config Doe')
+        self.assertEqual(committer.email, 'jcdoe@example.com')
+        self.assertEqual(author.name, 'John Config Doe')
+        self.assertEqual(author.email, 'jcdoe@example.com')
+        self.assertFalse(mock_get_uid.called)
+            
         env = {
             "GIT_AUTHOR_NAME": "John Doe",
             "GIT_AUTHOR_EMAIL": "jdoe@example.com",
@@ -226,7 +241,7 @@ class TestUtils(TestBase):
             "GIT_COMMITTER_EMAIL": "jane@example.com",
         }
         os.environ.update(env)
-        for cr in (None, self.rorepo.config_reader()):
+        for cr in (None, rwrepo.config_reader()):
             committer = Actor.committer(cr)
             author = Actor.author(cr)
             self.assertEqual(committer.name, 'Jane Doe')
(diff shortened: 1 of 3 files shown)


eric-wieser:
It looks like a maintainer might have to approve the workflow runs each time I try to fix the mypy troubles...

Yobmod:
I've added an overload to config.py  get_value() in main that fixes a mypy issue

eric-wieser:
> And I merged the other PR, so hopefully the checks on this one will auto run

No you didn't, you closed it!

test/test_util.py
```diff
@@ -238,16 +238,16 @@ def test_actor_get_uid_laziness_not_called(self, mock_get_uid):
...
-            committer = Actor.committer(cr)
-            author = Actor.author(cr)
-            if cr is None:  # otherwise, use value from config_reader
-                self.assertEqual(committer.name, 'user')
-                self.assertTrue(committer.email.startswith('user@'))
-                self.assertEqual(author.name, 'user')
-                self.assertTrue(committer.email.startswith('user@'))
+        committer = Actor.committer(None)
+        author = Actor.author(None)
+        # We can't test with `self.rorepo.config_reader()` here, as the uuid laziness
```
eric-wieser, line 243: I assume there's some way to get a writeable repo that we can override the config in, but I'm not familiar enough with your fixtures to want to go to the effort of working that out.

Byron: Thanks a lot for helping with the test. Getting a repository with write permissions [can be done like this](https://github.com/gitpython-developers/GitPython/blob/main/test/test_base.py#L104:L107).

eric-wieser: Which test fails?

Yobmod: Test_remote.py, assertion at line 333 failed.

Also came with warnings about not being able to create temporary folder at git://localhost:19418/daemon_repo-test_base-5tjbtmwj.

But fixed by setting git.email --global

eric-wieser: Oh, I thought you meant my new test failed!

Byron (approved):
Thank a lot, this looks good to me.

It seems the difficulty stems more from dealing with tests and CI than the actual code change, and I don't fully follow. Hence I leave it to @Yobmod to merge in case there is nothing I missed.

PR #1315, opened 2021-08-03
Fix trailing whitespace and incorrect overload

eric-wieser:
CI fails on master due to extra whitespace. The annotation is also incorrect, as if "1" is read from the config it will be converted to an integer, not a string.


Yobmod:
Thanks

eric-wieser:
Did you mean to close and not merge?

eric-wieser:
@Yobmod ^

Yobmod:
Whoops! 
I shouldn't try to work on trains!

Merging....

PR #1317, opened 2021-08-08
Fix type in setup.py, improve types in cmd.py

Yobmod:


Yobmod:
Closing, as somehow it is coming from my master branch instead of main

Issue #916, opened 2019-09-09
Git().custom_environment(GIT_SSH_COMMAND=git_ssh_cmd) issue


CaptainJL:
Hey all,

I am having an issue with using the custom_environment parameter.
This is on a Windows 10 machine.

Basically this will not work, returning the error (see under code)
```
import git
from git import Repo
from git import Git
git_ssh_cmd = 'ssh -i gpt_key'
with Git().custom_environment(GIT_SSH_COMMAND=git_ssh_cmd):
     Repo.clone_from('git@bitbucket.org:<private git>.git', 'python_test_cloning')
```

> git@bitbucket.org: Permission denied (publickey).
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.

However setting the env variable in the Repo.clone_from command works fine:
`Repo.clone_from('git@bitbucket.org:<private git>.git', 'python_test_cloning', env={"GIT_SSH_COMMAND": 'ssh -i gpt_key'})`

I need to be able to use the ssh key for operation other than cloning, so I need to be able to set up the local environment purely via the python interface. This also fails for reference (when I did sucessfully clone).
```
repo = git.Repo("python_test_cloning")
with repo.git.custom_environment(GIT_SSH_COMMAND='ssh -i gpt_key'):
    for remote in repo.remotes:
        remote.fetch()
```
with error:

> git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
>   cmdline: git fetch -v origin
>   stderr: 'fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.'

Any ideas of why this is not working and how to fix it?
Remember that this is Windows 10.

Thanks,
CaptainJL

Byron:
Thanks for the detailed information! Something I would do to verify the custom environment information is actually passed by [enabling debug output in the logs](https://gitpython.readthedocs.io/en/stable/tutorial.html#git-command-debugging-and-customization).

_Something you could do beforehand though_ is to use a fully qualified, absolute path to the key file, such as in `with repo.git.custom_environment(GIT_SSH_COMMAND='ssh -i /path/to/gpt_key'):...` .

Please let me know if this worked, or when there are interesting logs to look at. Thanks

CaptainJL:
Hey @Byron,
I had done the full string multiple times and it did exactly the same.

I have attached the debug-level log from the command:
```
[...]

Issue #866, opened 2019-04-29
can't clone use ssh key


haofrank:
**hi~ I can't use my id_rsa key to clone my repo.
this is my directory structure：**
├── code
├── git_test.py
├── id_rsa
├── id_rsa.pub
└── ssh_executable.sh

> git_test.py
``` python
import os
import git
from git import Repo
from git import Git

git_url = "git@xxx/xxx.git"
git_branch = "master"
out_dir = "./code"

# One
git_ssh_identity_file = os.path.expanduser('./id_rsa')
git_ssh_cmd = 'ssh -i %s' % git_ssh_identity_file

with Git().custom_environment(GIT_SSH_COMMAND=git_ssh_cmd):
  repo = Repo.clone_from(git_url, out_dir, branch="master")
  print("Success")

# Two
ssh_executable = os.path.join("./", 'ssh_executable.sh')

with Git().custom_environment(GIT_SSH=ssh_executable):
  repo = Repo.clone_from(git_url, out_dir, branch="master")
  print("Success")
```
>ssh_executable.sh
```shell
#!/bin/sh
ID_RSA="./id_rsa"
exec /usr/bin/ssh -o StrictHostKeyChecking=no -i $ID_RSA "$@"
```

 **I tried both methods and could not succeed clone repo....** 
 **if I move the `id_res` `id_res.pub` to `~/.ssh`. it will success. hoping to get your help~**
```
Traceback (most recent call last):
  File "git_test.py", line 15, in <module>
    repo = Repo.clone_from(git_url, out_dir, branch="master")
  File "/Users/frank/miniconda3/lib/python3.7/site-packages/git/repo/base.py", line 988, in clone_from
    return cls._clone(git, url, to_path, GitCmdObjectDB, progress, **kwargs)
  File "/Users/frank/miniconda3/lib/python3.7/site-packages/git/repo/base.py", line 939, in _clone
    finalize_process(proc, stderr=stderr)
  File "/Users/frank/miniconda3/lib/python3.7/site-packages/git/util.py", line 333, in finalize_process
    proc.wait(**kwargs)
  File "/Users/frank/miniconda3/lib/python3.7/site-packages/git/cmd.py", line 415, in wait
    raise GitCommandError(self.args, status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git clone --branch=master -v git@my.address.git ./code
  stderr: 'Cloning into './code'...
[...]

Issue #725, opened 2018-02-14
Use custom ssh key cannot clone


Too-Naive:
#690 
After solve the fetch/pull error, I found that I still can not clone from remote.
```python
os.environ['GIT_SSH_COMMAND'] = 'sudo ssh -v -i /root/key'
with Git().custom_environment(GIT_SSH_COMMAND=os.environ['GIT_SSH_COMMAND']):
	repo = Repo.clone_from('git@github.com:somebody/someproject','test')
```
output:
```
DEBUG:git.cmd:Popen(['git', 'clone', '-v', 'git@github.com:somebody/someproject', 'test'], cwd=/root/test/wel, universal_newlines=True, shell=None)
DEBUG:git.repo.base:Cmd(['git', 'clone', '-v', 'git@github.com:somebody/someproject', 'test'])'s unused stdout: Initialized empty Git repository in /root/test/.git/

DEBUG:git.cmd:AutoInterrupt wait stderr: 'Permission denied (publickey).\nfatal: The remote end hung up unexpectedly\n'
Traceback (most recent call last):
  File "test.py", line 7, in <module>
    repo = Repo.clone_from('git@github.com:somebody/someproject','test')
  File "/usr/local/lib/python2.7/site-packages/git/repo/base.py", line 982, in clone_from
    return cls._clone(git, url, to_path, GitCmdObjectDB, progress, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/git/repo/base.py", line 933, in _clone
    finalize_process(proc, stderr=stderr)
  File "/usr/local/lib/python2.7/site-packages/git/util.py", line 336, in finalize_process
    proc.wait(**kwargs)
  File "/usr/local/lib/python2.7/site-packages/git/cmd.py", line 418, in wait
    raise GitCommandError(self.args, status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v git@github.com:somebody/someproject test
  stderr: 'Permission denied (publickey).
fatal: The remote end hung up unexpectedly
```

Byron:
I would think creating a temporary instance of the `Git` command helper does not affect the `clone_from` call at all.

Instead you probably want to pass the environment directly when cloning [as described in the docs](http://gitpython.readthedocs.io/en/stable/reference.html?highlight=clone_from#git.repo.base.Repo.clone_from).

I am closing this one assuming that will work for you. But please feel free to comment if that's not the case and it might be reopened.

Issue #400, opened 2016-03-24
Cloning with Specific Private SSH Key


Neil511:
I was wondering if there was an option or configuration in GitPython that allows you to specify what ssh key you would like to use for certain clones? 

Example: I have a submodule that I would like to clone but it requires special access that I would like my script to employ. I need to allow access to anyone running the script and would like to avoid forcing users to setup an ssh config to do this.

Is there perhaps another way of tackling this problem that I'm unaware of?

Byron:
Hi @Neil511, 

indeed `git` itself supports this via [`GIT_SSH`](https://www.google.de/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=gitpython+GIT_SSH) and [`SSH_COMMAND`](https://www.google.de/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=gitpython+SSH_COMMAND), and I have linked numerous resources that should help to use it.

The problem mentioned here is also detailed, to some extend, in the [gitpython tutorial](http://gitpython.readthedocs.org/en/stable/tutorial.html), just look for `id_rsa` in there.

As I believe this question is answered, I am closing the issue - please feel free to comment on it in case you need it reopened.

theherk:
@Byron: As far as I can tell, this cannot be done. I have worked around it in the meantime with:

    os.environ['GIT_SSH_COMMAND'] = ssh_cmd

However, I cannot use `custom_environment` or `update_environment` before instantiating the repo.

I have tried this:

    ssh_cmd = 'ssh -i /home/user/.ssh/other_rsa'
    repo = Repo()
    repo.git.update_environment(GIT_SSH_COMMAND=ssh_cmd)
    repo = Repo.clone_from(repo_url, repo_dir)

NOGO - looks for wrong keys

Also:

    ssh_cmd = 'ssh -i /home/user/.ssh/other_rsa'
    repo = Repo()
    with repo.git.custom_environment(GIT_SSH_COMMAND=ssh_cmd)
        repo = Repo.clone_from(repo_url, repo_dir)

NOGO - looks for wrong keys

Also:

    ssh_cmd = 'ssh -i /home/user/.ssh/other_rsa'
    repo = Repo.clone_from(repo_url, repo_dir, env={'GIT_SSH_COMMAND': ssh_cmd})

NOGO - looks for wrong keys

[...]

PR #1322, opened 2021-08-17
Use the Git class type definition within Repo classmethods

mmulich:
Allow the GitCommandWrapperType definition to be used within the Repo
classmethods. This change follows the intended purpose as stated in
the code, "Subclasses may easily bring in their own custom types by
placing a constructor or type here."

The usecase that prompted this change has to do with
`GIT_SSH_COMMAND`. The goal is to setup a custom `Git` class with
knowledge of the value, something like as follows

```python
from git import Git as BaseGit, Repo as BaseRepo

class Git(BaseGit):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # For example, assign the SSH command using the current flask
        # app's configured setting.
        self.update_environment(GIT_SSH_COMMAND=current_app.config['GIT_SSH_COMMAND'])

class Repo(BaseRepo):
    GitCommandWrapperType = _Git
```

With this change, the above example will allow the developer to use
`Repo.clone_from(...)` with the intended outcome. Otherwise the
developer will have two differing result when using `Repo(...)` vs
`Repo.clone_from(...)`.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -1042,7 +1042,7 @@ class Repo(object):
             os.makedirs(path, 0o755)
 
         # git command automatically chdir into the directory
-        git = Git(path)
+        git = cls.GitCommandWrapperType(path)
         git.init(**kwargs)
         return cls(path, odbt=odbt)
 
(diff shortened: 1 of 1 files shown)


mmulich:
I believe this could help address issues around confusing usage of an SSH key when using `Repo.clone_from`. For example, #400 #725 #866 and #916 to name a few. Or at least this may be a step towards it. The use of `Repo.GitCommandWrapperType` as stated in my example would need to be documented, but first things first... fix the implementation.

Thanks for having simple contribution instructions.

Byron:
Thanks a lot for the fix for this long-standing oversight, it's much appreciated!

And thanks for the hint, I looked at the contribution guidelines and found it outdated, that's fixed now to hopefully further improve the experience.

PR #1325, opened 2021-08-23
Add encoding to utf-8 for fetch_info_lines, for fetch_head_info;

f100024:
Fix encoding error:
```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 39230-39231: character maps to <undefined>
```


Byron:
Thanks a lot for the fix.

In theory the encoding of `fetch_info_lines` is undefined, so it's strange that converting them into a `str()` works as this requires a known encoding. Especially for printing I think it's sufficient to encode with replacement instead of failure in case of encoding errors.

PR #1346, opened 2021-09-20
Fix typing issues with delete_head and Remote.add

rra:
delete_head and Head.delete historically accept either Head objects
or a str name of a head.  Adjust the typing to match.

Using assignment to make add an alias for create unfortunately
confuses mypy, since it loses track of the fact that it's a
classmethod and starts treating it like a staticmethod.  Replace
with a stub wrapper instead.

### DIFF
--- a/git/refs/head.py
+++ b/git/refs/head.py
@@ -129,7 +129,7 @@ class Head(Reference):
     k_config_remote_ref = "merge"           # branch to merge from remote
 
     @classmethod
-    def delete(cls, repo: 'Repo', *heads: 'Head', force: bool = False, **kwargs: Any) -> None:
+    def delete(cls, repo: 'Repo', *heads: 'Union[Head, str]', force: bool = False, **kwargs: Any) -> None:
         """Delete the given heads
 
         :param force:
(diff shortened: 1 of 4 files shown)


rra:
Ah, hm, I missed the impact on `Remote.delete`. Trying to think about the best way to handle that.

rra:
Amended with a possible fix. Alternately, perhaps the intent was not to allow strs for `Head.delete` and they only worked by accident (although in that case I wasn't sure how to create the necessary `Head` object to pass in).

rra:
FYI, for background, this fixes issues that I saw in https://github.com/lsst-sqre/neophile after updating GitPython, namely:
```
src/neophile/repository.py:58: error: Argument 1 to "delete_head" of "Repo" has incompatible type "str"; expected "Head"  [arg-type]
src/neophile/repository.py:76: error: Argument 1 to "delete_head" of "Repo" has incompatible type "str"; expected "Head"  [arg-type]
src/neophile/pr.py:312: error: Too few arguments  [call-arg]
src/neophile/pr.py:313: error: Argument 1 has incompatible type "Repo"; expected "Type[Remote]"  [arg-type]
src/neophile/pr.py:313: error: Argument 2 has incompatible type "str"; expected "Repo"  [arg-type]
```
The last three messages are from calling `Remote.add` with the code as it existed prior to this patch.

Byron (approved):
Thanks a lot for this contribution, it's much appreciated. Let's allow @Yobmod to also take a look in case I am missing something.

Yobmod (approved):

PR #1350, opened 2021-09-29
Add trailer property

Ket3r:
Git provides ways to interact trailer lines in the commit messages that look similar to RFC 822 e-mail headers ([git interpret-trailers](https://git-scm.com/docs/git-interpret-trailers) or [git log --format="%(trailers)"](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emtrailersoptionsem)). The new `trailers` property of the commit object parses its message and returns this trailer information in a dictionary. This is e.g. usefull when writing a Changelog generator [https://docs.gitlab.com/ee/development/changelog.html](https://docs.gitlab.com/ee/development/changelog.html)

Example commit message:
```
Adds feature xyz

Better explanation of this feature

Changelog: Added
Co-Author: Jane Doe
```

To access the trailer information you can now use
```python
commit.trailers["Changelog"] # Returns: Added
commit.trailers["Co-Author"] # Returns: Jane Doe
```


test/test_commit.py
```diff
@@ -429,3 +430,48 @@ def test_datetimes(self):
...
+            assert commit.trailers[KEY_2] == VALUE_2
+
+        # Check that trailer stays empty for multiple msg combinations
+        msgs = []
+        msgs.append(f"Subject\n")
+        msgs.append(f"Subject\n\nBody with some\nText\n")
+        msgs.append(f"Subject\n\nBody with\nText\n\nContinuation but\n doesn't contain colon\n")
+        msgs.append(f"Subject\n\nBody with\nText\n\nContinuation but\n only contains one :\n")
+        msgs.append(f"Subject\n\nBody with\nText\n\nKey: Value\nLine without colon\n")
+        msgs.append(f"Subject\n\nBody with\nText\n\nLine without colon\nKey: Value\n")
```
Byron, line 463: What about this one?

```suggestion
        msgs.append(f"Subject\n\nBody with\nText\n\nLine without colon\nhere comes the sun: yeah right\n")
```

I think it's valid for tokens to contain whitespace, maybe adding this to the test would help setting this as an example. It's a bit confusing to allow that though as it might make some last paragraphs that merely contain a colon a valid trailer.

Byron (commented):

There is only a tiny question about what to do with whitespace in keys. If it's allows, it's worth mentioning it in the docs. Otherwise there could be a test that shows those won't be considered keys, in order to reduce ambiguity.

What do you think about this?

git/objects/commit.py
```diff
@@ -315,6 +316,53 @@ def stats(self) -> Stats:
...
+                "key1": "value1",
+                "key2": "value 2 with inner spaces"
+            }
+
+        :return: Dictionary containing whitespace stripped trailer information
+
+        """
+        d = {}
+        cmd = ['git', 'interpret-trailers', '--parse']
+        proc: Git.AutoInterrupt = self.repo.git.execute(cmd, as_process=True, istream=PIPE)  # type: ignore
```
Ket3r, line 356: Can someone help with this?
I don't know how to get mypy to accept `subprocess.PIPE` as argument

Byron: Maybe @Yobmod could help here.

Yobmod: It looks like the isstream type annotation is too narrow in cmd.Git.execute().
It should be int | IO | None if we accept any args that Popen does.

But execute() needs a lot more annotations anyway to clear up the remaining str | bytes problems from py2.
So you can update the type annotation, ork leave the type: ignore is ok for now.
Once updated, mypy should then warn that the type: ignore is no longer needed.

Byron:
As I have disabled mypi to not fail CI anymore, I think it's safe to merge this feature and deal with mypi related typing improvements another time.

Thanks everyone for their involvement and help, and sorry for the wait.

nickbroon:
Should trailers be a list instead of a dictionary?

Consider for example this linux kernel commit: https://gitlab.com/linux-kernel/stable/-/commit/1382999aa0548a171a272ca817f6c38e797c458c which has following trailers that include multiple `cc` and `signed-off-by` trailers:

```
Link: https://lore.kernel.org/lkml/7cbe96cf-e0b5-ba63-d1b4-f63d2e826efa@suse.cz/


Cc: stable@vger.kernel.org # 6.1+
Reported-by: default avatar[Vlastimil Babka](mailto:vbabka@suse.cz) <[vbabka@suse.cz](mailto:vbabka@suse.cz)>
Suggested-by: default avatar[Linus Torvalds](mailto:torvalds@linux-foundation.org) <[torvalds@linux-foundation.org](mailto:torvalds@linux-foundation.org)>
Acked-by: default avatar[Luigi Semenzato](mailto:semenzato@chromium.org) <[semenzato@chromium.org](mailto:semenzato@chromium.org)>
Cc: Peter Huewe <peterhuewe@gmx.de>
Cc: Jarkko Sakkinen <jarkko@kernel.org>
Cc: James Bottomley <James.Bottomley@HansenPartnership.com>
Cc: Johannes Altmanninger <aclopte@gmail.com>
Signed-off-by: Jason Donenfeld's avatar[Jason A. Donenfeld](https://gitlab.com/zx2c4) <[Jason@zx2c4.com](mailto:Jason@zx2c4.com)>
Signed-off-by: default avatar[Linus Torvalds](mailto:torvalds@linux-foundation.org) <[torvalds@linux-foundation.org](mailto:torvalds@linux-foundation.org)>
```

And `git interpret-trailers --parse ...` will return all of these.

Byron:
I agree - sorry for not catching this. It's better to return it as flat list and leave any kind of indexing to the user.
Maybe a follow-up PR could be created that adds another method, like `trailer_list()` (or better, if you have ideas) which returns such list. The `trailers()` could then use that to build a dict along with a big warning that it only shows the last seen value with a certain prefix.

nickbroon:
> I agree - sorry for not catching this. It's better to return it as flat list and leave any kind of indexing to the user. Maybe a follow-up PR could be created that adds another method, like `trailer_list()` (or better, if you have ideas) which returns such list. The `trailers()` could then use that to build a dict along with a big warning that it only shows the last seen value with a certain prefix.

Assuming you don't want to make non-backwards compatible change to the API with the fix to this bug then adding a new `trailer_list()` API seems appropriate. Adding as documentation/comment warning on trailer() should definitely be done, and perhaps marking it as deprecated for removal in future revision should be considered as I don't with its current behaviour it's particularly useful.

PR #1351, opened 2021-09-29
Fix broken test requirements

Ket3r:
The ddt package changed the function signature in version 1.4.3 from
idata(iterable) to idata(iterable, index_len). Hopefully this was just a
mistake and the new argument will be optional in future versions (see issue datadriventests/ddt#97)


Byron:
Thanks a million!

Issue #1349, opened 2021-09-23
mypy is failing on `from git import Blob`


pbeza:
Why `mypy` is failing when running this code:
```bash
$ cat /tmp/test.py 
from git import Blob

$ mypy /tmp/test.py
/tmp/test.py:1: error: Module "git" has no attribute "Blob"
Found 1 error in 1 file (checked 1 source file)

$ python3 -c "import git; print(git.__version__)"
3.1.23

$ mypy --version
mypy 0.910

$ python3 --version
Python 3.9.7
```
?

I can see that there is a conditional import in [`git/objects/base.py`](https://github.com/gitpython-developers/GitPython/blob/05825dcaaf6bf3de7d7e2bc595ad6dcb145f11e8/git/objects/base.py#L22):
```python
if TYPE_CHECKING:
    from git.repo import Repo
    from gitdb.base import OStream
    from .tree import Tree
    from .blob import Blob
    from .submodule.base import Submodule
    from git.refs.reference import 
```
and it seems that `TYPE_CHECKING` is set to `False` (I guess so based on the above error). Do you know why is that? How to workaround the problem?

Byron:
Thanks for letting us know. If `Blob` was previously available and now it is behind a conditional, this may be an issue that was recently introduced.

However, it looks like it should rather be imported from `git.objects`, and maybe that was also the case [historically](https://gitpython.readthedocs.io/en/3.0.7/search.html?q=Blob&check_keywords=yes&area=default).

Yobmod:
I'll have a look at this in the morning.

But initial thoughts are that the import from .blob is a relative import in the objects module, so that should work. (And mypy is working for me)

pbeza:
> What is test.py? Your own code? Can I see it?

It's a one-liner – see my post above:
```bash
$ cat /tmp/test.py 
from git import Blob
```

> You might just be able to change a mypy setting to get it to pass. E.g allow_reexport = True and strict = False (see the mypy settings in our pyproject.toml.) if using strict, the order of settings also makes a difference).
> 
> If changing the import or mypy settings fixes it, let me know.

I will check it and let you know.

tucked:
### SSCCE
``` sh
python3.7 -m venv venv
venv/bin/pip install GitPython==3.1.24 mypy==0.910
vim sscce.py
```
``` py
import git
git.Repo
```
``` sh
venv/bin/python sscce.py  # runs fine
venv/bin/mypy --strict sscce.py
```
[...]

PR #1352, opened 2021-09-30
Replace wildcard imports with concrete imports

trym-b:
All `from <module> import *` has now been replaced by
`from <module> import X, Y, ...`.

Contributes to #1349

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -5,7 +5,7 @@
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
 # flake8: noqa
 #@PydevCodeAnalysisIgnore
-from git.exc import *                       # @NoMove @IgnorePep8
+from git.exc import GitError, GitCommandError, GitCommandNotFound, UnmergedEntriesError, CheckoutError, InvalidGitRepositoryError, NoSuchPathError, BadName                       # @NoMove @IgnorePep8
 import inspect
 import os
 import sys
@@ -39,16 +39,16 @@ _init_externals()
 #{ Imports
 
 try:
-    from git.config import GitConfigParser  # @NoMove @IgnorePep8
-    from git.objects import *               # @NoMove @IgnorePep8
-    from git.refs import *                  # @NoMove @IgnorePep8
-    from git.diff import *                  # @NoMove @IgnorePep8
-    from git.db import *                    # @NoMove @IgnorePep8
-    from git.cmd import Git                 # @NoMove @IgnorePep8
-    from git.repo import Repo               # @NoMove @IgnorePep8
-    from git.remote import *                # @NoMove @IgnorePep8
-    from git.index import *                 # @NoMove @IgnorePep8
-    from git.util import (                  # @NoMove @IgnorePep8
+    from git.config import GitConfigParser                                                         # @NoMove @IgnorePep8
+    from git.objects import Blob, Commit, Object, Submodule, Tree                                  # @NoMove @IgnorePep8
+    from git.refs import Head, Reference, RefLog, RemoteReference, SymbolicReference, TagReference # @NoMove @IgnorePep8
+    from git.diff import Diff, DiffIndex, NULL_TREE                                                # @NoMove @IgnorePep8
+    from git.db import GitCmdObjectDB, GitDB                                                       # @NoMove @IgnorePep8
+    from git.cmd import Git                                                                        # @NoMove @IgnorePep8
+    from git.repo import Repo                                                                      # @NoMove @IgnorePep8
+    from git.remote import FetchInfo, PushInfo, Remote, RemoteProgress                             # @NoMove @IgnorePep8
+    from git.index import BlobFilter, IndexEntry, IndexFile                                        # @NoMove @IgnorePep8
+    from git.util import (                                                                         # @NoMove @IgnorePep8
         LockFile,
         BlockingLockFile,
         Stats,
--- a/git/exc.py
+++ b/git/exc.py
@@ -5,8 +5,7 @@
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
 """ Module containing all exceptions thrown throughout the git package, """
 
-from gitdb.exc import BadName  # NOQA @UnusedWildImport skipcq: PYL-W0401, PYL-W0614
-from gitdb.exc import *     # NOQA @UnusedWildImport skipcq: PYL-W0401, PYL-W0614
+from gitdb.exc import BadName, BadObject  # NOQA @UnusedWildImport skipcq: PYL-W0401, PYL-W0614
 from git.compat import safe_decode
 
 # typing ----------------------------------------------------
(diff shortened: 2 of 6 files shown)


Byron:
That's such a great improvement, thanks a lot!

trym-b:
I might have accidentally introduced a regression in this PR. The reason for this might be because of the messy way imports were handled before (functionality was exposed through nested wildcard imports) and lack of testing.

In the commit before this PR (5e73cabd041f45337b270d5e78674d88448929e6) the result of
```python
import git
dir(git)
```
was
```python
old = ['Actor', 'AmbiguousObjectName', 'BadName', 'BadObject', 'BadObjectType', 'BaseIndexEntry', 'Blob', 'BlobFilter', 'BlockingLockFile', 'CacheError', 'CheckoutError', 'CommandError', 'Commit', 'Diff', 'DiffIndex', 'Diffable', 'FetchInfo', 'GIT_OK', 'Git', 'GitCmdObjectDB', 'GitCommandError', 'GitCommandNotFound', 'GitConfigParser', 'GitDB', 'GitError', 'HEAD', 'Head', 'HookExecutionError', 'IndexEntry', 'IndexFile', 'IndexObject', 'InvalidDBRoot', 'InvalidGitRepositoryError', 'List', 'LockFile', 'NULL_TREE', 'NoSuchPathError', 'ODBError', 'Object', 'Optional', 'ParseError', 'PathLike', 'PushInfo', 'RefLog', 'RefLogEntry', 'Reference', 'Remote', 'RemoteProgress', 'RemoteReference', 'Repo', 'RepositoryDirtyError', 'RootModule', 'RootUpdateProgress', 'Sequence', 'Stats', 'Submodule', 'SymbolicReference', 'TYPE_CHECKING', 'Tag', 'TagObject', 'TagReference', 'Tree', 'TreeModifier', 'Tuple', 'Union', 'UnmergedEntriesError', 'UnsupportedOperation', 'UpdateProgress', 'WorkTreeRepositoryUnsupported', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_init_externals', 'base', 'cmd', 'compat', 'config', 'db', 'diff', 'exc', 'fun', 'head', 'index', 'inspect', 'log', 'objects', 'os', 'osp', 'reference', 'refresh', 'refs', 'remote', 'repo', 'rmtree', 'safe_decode', 'symbolic', 'sys', 'tag', 'to_hex_sha', 'typ', 'types', 'util']
```
while this PR's commit (53d94b8091b36847bb9e495c76bb5a3ec2a2fdb5) had this output:
```python
new = ['Actor', 'BadName', 'Blob', 'BlobFilter', 'BlockingLockFile', 'CheckoutError', 'Commit', 'Diff', 'DiffIndex', 'FetchInfo', 'GIT_OK', 'Git', 'GitCmdObjectDB', 'GitCommandError', 'GitCommandNotFound', 'GitConfigParser', 'GitDB', 'GitError', 'Head', 'IndexEntry', 'IndexFile', 'InvalidGitRepositoryError', 'LockFile', 'NULL_TREE', 'NoSuchPathError', 'Object', 'Optional', 'PathLike', 'PushInfo', 'RefLog', 'Reference', 'Remote', 'RemoteProgress', 'RemoteReference', 'Repo', 'Stats', 'Submodule', 'SymbolicReference', 'TagReference', 'Tree', 'UnmergedEntriesError', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_init_externals', 'cmd', 'compat', 'config', 'db', 'diff', 'exc', 'index', 'inspect', 'objects', 'os', 'osp', 'refresh', 'refs', 'remote', 'repo', 'rmtree', 'sys', 'types', 'util']
```

The interesting part if the diff between the old and the new:
```python
list(set(old) - set(new))
['CommandError', 'Tuple', 'Diffable', 'HEAD', 'UnsupportedOperation', 'ODBError', 'HookExecutionError', 'List', 'CacheError', 'log', 'BadObject', 'base', 'head', 'reference', 'TreeModifier', 'AmbiguousObjectName', 'typ', 'RootModule', 'symbolic', 'UpdateProgress', 'IndexObject', 'tag', 'WorkTreeRepositoryUnsupported', 'RepositoryDirtyError', 'fun', 'TYPE_CHECKING', 'InvalidDBRoot', 'BaseIndexEntry', 'RefLogEntry', 'Union', 'BadObjectType', 'RootUpdateProgress', 'ParseError', 'TagObject', 'to_hex_sha', 'safe_decode', 'Tag', 'Sequence']
```

Some of the lost imports should probably have been tested and been imported on `import git`, like `TreeModifier`, while others like `List`, `Union` and `to_hex_sha` should perhaps not be exposed?

I can help out fixing this regression, but I would like to know what should be exposed when importing `git` and what should not. Sorry for any inconvenience this might have caused.

Byron:
Thanks a million for, somehow magically, returning to this PR before it becomes an issue. It's clearly something the review should have caught 😅.

> I can help out fixing this regression, but I would like to know what should be exposed when importing git and what should not. Sorry for any inconvenience this might have caused.

It would be great to see another PR for this for sure, otherwise the only fix I could make is to revert the previous PR. I'd say that the intend was to import only what's defined in the modules the `*` is importing from. By now this is most certainly a bug that much code relies on, so when fixing it I'd play it safe and re-introduce all imports, even the ones that very clearly seem unintended.

Doing so seems to be the only way to pacify mypi and avoid breakage.

trym-b:
> It would be great to see another PR for this for sure, otherwise the only fix I could make is to revert the previous PR. I'd say that the intend was to import only what's defined in the modules the `*` is importing from. By now this is most certainly a bug that much code relies on, so when fixing it I'd play it safe and re-introduce all imports, even the ones that very clearly seem unintended.

I agree, the best way forward now is to revert the PR, and also probably add a test that verifies that the result of `dir(git)` has not changed. After that such a test had been added there might be a chance to remove the wildcard imports.

I have also seen that two submodules (https://github.com/gitpython-developers/smmap and https://github.com/gitpython-developers/gitdb) used by Gitpython also have the issue of doing wildcard imports that might mess with `mypy`, so perhaps it would be best to start with those first.

I will revert this commit asap. Thanks for the feedback

Byron:
> I agree, the best way forward now is to revert the PR, and also probably add a test that verifies that the result of dir(git) has not changed. After that such a test had been added there might be a chance to remove the wildcard imports.

That's a very good idea! Loving it :).

> I have also seen that two submodules (https://github.com/gitpython-developers/smmap and https://github.com/gitpython-developers/gitdb) used by Gitpython also have the issue of doing wildcard imports that might mess with mypy, so perhaps it would be best to start with those first.

Another great catch! I'd love if these wouldn't cause additional issues or if there would be an easy way, but if not, PRs to them as well as new releases can be made for them, too.

PR #1360, opened 2021-10-12
Let remote.push return a PushInfoList

Sjord:
List-like, so that it's backward compatible. But it has a new method
raise_on_error, that throws an exception if anything failed to push.

Related to #621

This is more of a proposal for this design. Is this actually backward-compatible? Should we check the flags of all PushInfos in `raise_on_error`? Is this better than a new method `push_really_or_else_raise()` or `push(raise=True)`? What if a `PushInfo` is missing from the result because it totally failed?

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -116,6 +116,22 @@ def to_progress_instance(progress: Union[Callable[..., Any], RemoteProgress, Non
     return progress
 
 
+class PushInfoList(IterableList):
+    def __new__(cls) -> 'IterableList[IterableObj]':
+        return super(IterableList, cls).__new__(cls, 'push_infos')
+
+    def __init__(self) -> None:
+        super().__init__('push_infos')
+        self.exception = None
+
+    def raise_on_error(self):
+        """
+        Raise an exception if any ref failed to push.
+        """
+        if self.exception:
+            raise self.exception
+
+
 class PushInfo(IterableObj, object):
     """
     Carries information about the result of a push operation of a single head::
@@ -774,7 +790,7 @@ class Remote(LazyMixin, IterableObj):
 
     def _get_push_info(self, proc: 'Git.AutoInterrupt',
                        progress: Union[Callable[..., Any], RemoteProgress, None],
-                       kill_after_timeout: Union[None, float] = None) -> IterableList[PushInfo]:
+                       kill_after_timeout: Union[None, float] = None) -> PushInfoList:
         progress = to_progress_instance(progress)
 
         # read progress information from stderr
@@ -782,7 +798,7 @@ class Remote(LazyMixin, IterableObj):
         # read the lines manually as it will use carriage returns between the messages
         # to override the previous one. This is why we read the bytes manually
         progress_handler = progress.new_message_handler()
-        output: IterableList[PushInfo] = IterableList('push_infos')
+        output: PushInfoList = PushInfoList()
 
         def stdout_handler(line: str) -> None:
             try:
@@ -796,13 +812,14 @@ class Remote(LazyMixin, IterableObj):
         stderr_text = progress.error_lines and '\n'.join(progress.error_lines) or ''
         try:
             proc.wait(stderr=stderr_text)
-        except Exception:
+        except Exception as e:
             # This is different than fetch (which fails if there is any std_err
             # even if there is an output)
             if not output:
                 raise
             elif stderr_text:
                 log.warning("Error lines received while fetching: %s", stderr_text)
+                output.exception = e
 
         return output
 
--- a/git/remote.py
+++ b/git/remote.py
@@ -122,14 +122,14 @@ class PushInfoList(IterableList):
 
     def __init__(self) -> None:
         super().__init__('push_infos')
-        self.exception = None
+        self.error = None
 
-    def raise_on_error(self):
+    def raise_if_error(self):
         """
         Raise an exception if any ref failed to push.
         """
-        if self.exception:
-            raise self.exception
+        if self.error:
+            raise self.error
 
 
 class PushInfo(IterableObj, object):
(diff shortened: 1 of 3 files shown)


Byron:
> Well, the new PushInfoList extends IterableList and list. So isinstance(..., list) would still return true, and PushInfoList can still be iterated over to retrieve PushInfo items. It's true that the return type is different, but I thought that would have limited impact.

Thanks for the clarification. Could you post an example invocation for me to understand how one would enable exceptions using the return type? My intuition here is that a separate method with different semantics will be similarly opt-in, but inherently more straightforward to use.

Sjord:
The most straightforward usage is like this:

```
remote.push().raise_on_error()
```

The return value of `push()` is backwards compatible, making it possible to combine the ability to raise exceptions with checking flags.

```
push_info = remote.push()
if not push_info[0].flags & push_info.NO_MATCH:
    push_info.raise_on_error()
```

This adds more flexibility for custom error handling, compared to a new method.

A similar pattern is used in requests' [raise_for_status](https://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status), which raises an error when a HTTP request returns an error status code.

Byron:
Thanks for the clarification! I think my biggest stumbling block was the naming, specifically `…on_error()` which made me think it's a capability that is turned on to, then, raise if an error occurs. Of course it didn't compute that this would be set on a return value.

Maybe it will help most to rename it to `raise_for_error()` instead to go with the mainstream, even though my intuition would have been `raise_if_error()`.

Since the the respective field is called `exception`, I think I would be confused if the actual value is not stored in `result.error`. Maybe it's better to either not expose this by making the field private, or to rename either method or field for consistency. Since built-in error types are usually called `…Error`, like `IOError`, I suppose the field should be named similarly to match.

Once the naming is clear, I think there shouldn't be anything else holding this PR back.

Byron:
@Sjord Please feel free to mark this PR as ready when ready. If there are problems with types, [AT]Yobmod might be able to provide some clues.

git/remote.py
```diff
@@ -116,6 +116,23 @@ def to_progress_instance(progress: Union[Callable[..., Any], RemoteProgress, Non
     return progress
 
 
+class PushInfoList(IterableList):
```
Byron, line 119: The strange thing is that I would expect `IterableList[PushInfo]` to be the base class, or at least something in this type should indicate that `PushInfo`s are iterated over.

That way, `PushInfoList` would be fitting in place of `IterableList[PushInfo]`, which might even be a test to add to assure the type system stays compatible, too.

Something like…

```python

def f(l: IterableList[PushInfo]):
    pass

f(PushInfoList())
```

Byron (changes requested):
This looks great, thanks a lot.

I think once we can prove that the type system won't break by this change either it's definitely ready to be merged.

Thanks a lot

git/remote.py
Sjord: I agree, but I had trouble [getting this to work with mypy](https://github.com/gitpython-developers/GitPython/runs/4165677582?check_suite_focus=true).

The list with pushinfos was created like this:

```
output: IterableList[PushInfo] = IterableList('push_infos')
```

I am not sure what the `push_infos` argument is for, but I tried moving it into PushInfoList:

```
class PushInfoList(IterableList[PushInfo]):
    def __new__(cls) -> 'PushInfoList':
        return super().__new__(cls, 'push_infos')

    def __init__(self) -> None:
        super().__init__('push_infos')
        self.error: Optional[Exception] = None
```

However, mypy doesn't agree with this. The types don't match up in `__new__`. `super(IterableList[PushInfo], cls)` doesn't work, so now I have a call explicitly to `IterableList.__new__`. Is that OK?

```
class PushInfoList(IterableList[PushInfo]):
    def __new__(cls) -> 'PushInfoList':
        return cast(PushInfoList, IterableList.__new__(cls, 'push_infos'))
```

Byron: `"push_infos"` is supposed to be an `id_attribute`, which definitely isn't useful here as [it is used like this](https://github.com/gitpython-developers/GitPython/blob/main/git/util.py#L995:L1001).

Maybe @Yobmod can help us clarify if this `.__new__` calling convention to pacify mypi is the expected way, or if there are more idiomatic ways to do this.

From my point of view, it's certainly OK even though I can't tell if a future patch release of mypi is going to stop accepting it.

Byron (changes requested):
Thanks for the improvement to the types, much appreciated.

Is it possible to add a test for this type, too, to validate that the new type now passes for the old if used in a typed function like this?


```Python
def f(l: IterableList[PushInfo]):
    pass

f(PushInfoList())
```

Sjord:
> Is it possible to add a test for this type, too, to validate that the new type now passes for the old if used in a typed function like this?

This is not so easy. I tried the folllowing, but `isinstance` and `issubclass` don't work on generic types:

```
self.assertIsInstance(results, IterableList[PushInfo])
```

Another possibility would be to let mypy do the testing; mypy would raise an error on your example code if `PushInfoList` is no longer compatible with `IterableList[PushInfo]`. But the unittest code is not checked with mypy, so putting it there won't have any effect. I am also not comfortable with putting dead code in git/remote.py just for mypy testing sake.

Byron:
> Another possibility would be to let mypy do the testing; mypy would raise an error on your example code if PushInfoList is no longer compatible with IterableList[PushInfo]. But the unittest code is not checked with mypy, so putting it there won't have any effect. I am also not comfortable with putting dead code in git/remote.py just for mypy testing sake.

Thanks for having given it a shot. I wasn't even aware that the tests aren't yet checked with mypi and can imagine that trying to do that would result in a slew of errors that aren't worth fixing.

If you can tell me something like the code I suggested worked in your local tests, I think we are ready to merge, having tried our best to not break anything.

Thank you

Sjord:
> If you can tell me something like the code I suggested worked in your local tests

Yes, I temporarily put the example you supplied in remote.py, and mypy agrees that a `PushInfoList` instance is acceptable for a `IterableList[PushInfo]` parameter.

Byron (approved):
Thanks a lot for all the work that went into this and for hanging in there :).

PR #1361, opened 2021-10-13
Add support for Python 3.10

hugovk:
Python 3.10 was released last week:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

Sphinx has only just added support for Python 3.10 (https://github.com/sphinx-doc/sphinx/pull/9712), not yet released. The fix is part of their 4.3.0 milestone (https://github.com/sphinx-doc/sphinx/milestone/114): "Due by November 06, 2021".

So in the meantime, for Python 3.10 this installs the specific commit of Shpinx which works for 3.10. Once Sphinx 4.3.0 is out, this can be replaced with a normal pin.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,6 @@
-sphinx==4.1.2
+# TODO Temporary until Sphinx 4.3.0 released with Python 3.10 support:
+# https://github.com/sphinx-doc/sphinx/pull/9712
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for the contribution, looks good to me!

I hope to see you again for when sphinx is ready to make the move as well.

PR #1375, opened 2021-11-09
Warn when working tree is empty on non-bare repos

HerrMuellerluedenscheid:
When loading an empty repo by accident and iterating through commits lead to suprprising results. It should be warned while loading the non-bare repo that this does not appear to be a repo.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -219,6 +219,9 @@ class Repo(object):
         # in the first place
         if self._bare:
             self._working_tree_dir = None
+
+        if not self._bare and self._working_tree_dir is None:
+            log.warning('No tree found. Is this a git repository?')
         # END working dir handling
 
         self.working_dir: Optional[PathLike] = self._working_tree_dir or self.common_dir


HerrMuellerluedenscheid:
Can even be considered to raise an exception here.

Byron (changes requested):
I think it would be interesting to explore how GitPython reacts if the working tree dir is `None` even though the repo isn't bare. Because that seems like it's a bug as the working tree dir should be set even if it's empty.

In other words, non-bare repositories that are empty, i.e. a `.git/` directory in an otherwise empty directory, should not lead to surprises except for being empty.

Note that the main branch might not be born yet on empty repositories, which is something the caller always has to be aware of independently.

Independently of all that, it would definitely help this PR along if there would be more substance to the encountered issues in the form of 'expected - actual' comparisons.

Byron:
Since this PR has been waiting for changes for a while now I think it's fair to close it for now.
Thanks a lot for the contribution, and when there is time to respond to the change request we can certainly find a way to get it merged, too.

PR #1380, opened 2021-11-30
Revert the use of typing_extensions in py3.8+

mgorny:
The original change requiring py3.10 TypeGuard (and matching
typing_extensions) has been reverted, so revert the requirement
on typing_extensions as well.


Byron:
Thanks a lot!

PR #1387, opened 2021-12-23
Use NUL character to extract meta and path from git diff

NHanser:
Use NUL character instead of semicolon to extract meta and path. Avoid errors in during git diff when dealing with filenames containing semicolons

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -509,9 +509,9 @@ class Diff(object):
     def _handle_diff_line(lines_bytes: bytes, repo: 'Repo', index: DiffIndex) -> None:
         lines = lines_bytes.decode(defenc)
 
-        for line in lines.split(':')[1:]:
-            meta, _, path = line.partition('\x00')
-            path = path.rstrip('\x00')
+        it = iter(lines.split('\x00'))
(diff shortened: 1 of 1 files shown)


Byron:
It looks like CI is currently broken as it pulls in a more recent version of mypi, which now takes offense in additional portions of the code.

@Yobmod Is this something that could be fixed, maybe alongside pinning the mypi version so this won't sneak up again? Thanks for your advice.

Byron:
Mypi is now used for advisory only, and set to not fail CI anymore.

That way we can fix the type-system from time to time as the need arises, but won't allow it to prevent changes anymore.
I'd hope there are better ways to do this, like pinning the mypi version.

Byron:
I am sorry but had to revert this PR with 3ef81e182fcd3fca3f83216cf81d92d08c19cf5e as it did [cause breakage](https://github.com/gitpython-developers/GitPython/runs/4734342765?check_suite_focus=true#step:7:718) in main for reasons I didn't uncover yet. It's odd that CI didn't discover this earlier.

Maybe you could have another look and we can try again.

PR #1392, opened 2022-01-09
Use NUL character to extract meta and path from git diff

NHanser:
Use NUL character instead of semicolon to extract meta and path. Avoid errors in during git diff when dealing with filenames containing semicolons
Re-submit (https://github.com/gitpython-developers/GitPython/pull/1387#issuecomment-1007079621)

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -509,9 +509,9 @@ class Diff(object):
     def _handle_diff_line(lines_bytes: bytes, repo: 'Repo', index: DiffIndex) -> None:
         lines = lines_bytes.decode(defenc)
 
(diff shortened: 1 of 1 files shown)


NHanser:
Test not passed. Need to review why (only valid with raw? I'm using diff with this option if I remember (don't have code with me))

Byron:
Thanks for giving it another look. By the way, it's possible to turn on actions in your own fork which then should run them each time you push into a branch even, no PR to check CI is needed then.

I also don't mind to keep this PR open and let it re-run each time there is a change to see if that worked and in case it's not reproducible locally - I'd think it should be.

PR #1393, opened 2022-01-09
Fix doc string error in Objects.Commit

takuti:
Simple errors in the API documentation. Checked locally by `make html`


Byron:
Thanks for the fixes and taking the time to contribute them.

PR #1398, opened 2022-01-11
import unittest adds 0.250s to script launch time

wonder-mice:
This should not be imported at root level, since it adds a lot of initialization overhead without need.


Byron:
That's such a great catch, thanks a million.

This will definitely save a lot of cycles world-wide!

PR #1399, opened 2022-01-13
Use bash to open extensionless hooks on windows

### LINKED ISSUE #703: Hooks are not executed on windows correctly
Upon quick troubleshooting of #702 on appveyor, it seems that their (succesfull) execution was never tested on windows (last succesful appeveyor run is https://ci.appveyor.com/project/Byron/gitpython/build/1.0.161 if I got it right).

I could not figure out a legit way to make hooks runnable without us parsing the shebang etc.

### LINKED ISSUE #971: OSError: [WinError 193] %1 is not a valid Win32 application when commiting LFS-enabled repository on Windows
Hi. I found that GitPython 3.0.5 throw errors when committing to LFS-enabled repository.

Here is error log:

> Traceback (most recent call last):
>   File "C:\Users\XXXXXXXX\AppData\Local\Programs\Python\Python37\lib\site-packages\git\index\fun.py", line 85, in run_commit_hook
>     creationflags=PROC_CREATIONFLAGS,)
>   File "C:\Users\XXXXXXXX\AppData\Local\Programs\Python\Python37\lib\subprocess.py", line 775, in __init__
>     restore_signals, start_new_session)
>   File "C:\Users\XXXXXXXX\AppData\Local\Programs\Python\Python37\lib\subprocess.py", line 1178, in _execute_child
>     startupinfo)
> OSError: [WinError 193] %1 is not a valid Win32 application
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "c:/Users/XXXXXXXX/Downloads/GitPythonTest/test.py", line 7, in <module>
>     repo.index.commit('AA')
>   File "C:\Users\XXXXXXXX\AppData\Local\Programs\Python\Python37\lib\site-packages\git\index\base.py",\index\base.py", line 959, in commit
>     run_commit_hook('post-commit', self)
>   File "C:\Users\XXXXXXXX\AppData\Local\Programs\Python\Python37\lib\site-packages\git\index\fun.py", line 87, in run_commit_hook
>     raise HookExecutionError(hp, ex)
> git.exc.HookExecutionError: Hook('C:\Users\XXXXXXXX\Downloads\GitPythonTest\.git\hooks\post-commit') failed due to: OSError('[WinError 193] %1 is not not a valid Win32 application')
>   cmdline: C:\Users\XXXXXXXX\Downloads\GitPythonTest\.git\hooks\post-commit

This can be reproduced by following code:
```python
import git

repo = git.Repo('.')

repo.git.add(['test.png', '-A'])

repo.index.commit('AA')
```
Target repository is LFS-enabled and *.png are tracked.

idbrii:
Fix #971. Partly resolve #703.

If the hook doesn't have a file extension, then Windows won't know how
to run it and you'll get "[WinError 193] %1 is not a valid Win32
application". It's very likely that it's a shell script of some kind, so
use bash.exe (commonly installed via Windows Subsystem for Linux). We
don't want to run all hooks with bash because they could be .bat files.

test_index.py passes (two are skipped).

### DIFF
--- a/git/index/fun.py
+++ b/git/index/fun.py
@@ -3,6 +3,7 @@
 # NOTE: Autodoc hates it if this is a docstring
 
 from io import BytesIO
+from pathlib import Path
 import os
 from stat import (
     S_IFDIR,
@@ -21,6 +22,7 @@ from git.compat import (
     force_text,
     force_bytes,
     is_posix,
+    is_win,
     safe_decode,
 )
 from git.exc import (
@@ -76,6 +78,10 @@ def hook_path(name: str, git_dir: PathLike) -> str:
     return osp.join(git_dir, 'hooks', name)
 
 
+def _has_file_extension(path):
+    return osp.splitext(path)[1]
+
+
 def run_commit_hook(name: str, index: 'IndexFile', *args: str) -> None:
     """Run the commit hook of the given name. Silently ignores hooks that do not exist.
     :param name: name of hook, like 'pre-commit'
@@ -89,8 +95,15 @@ def run_commit_hook(name: str, index: 'IndexFile', *args: str) -> None:
     env = os.environ.copy()
     env['GIT_INDEX_FILE'] = safe_decode(str(index.path))
     env['GIT_EDITOR'] = ':'
+    cmd = [hp]
     try:
-        cmd = subprocess.Popen([hp] + list(args),
+        if is_win and not _has_file_extension(hp):
+            # Windows only uses extensions to determine how to open files
+            # (doesn't understand shebangs). Try using bash to run the hook.
+            relative_hp = Path(hp).relative_to(index.repo.working_dir).as_posix()
+            cmd = ["bash.exe", relative_hp]
+        
+        cmd = subprocess.Popen(cmd + list(args),
                                env=env,
                                stdout=subprocess.PIPE,
                                stderr=subprocess.PIPE,
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -13,6 +13,7 @@ from stat import (
 )
 import tempfile
 from unittest import skipIf
+import shutil
 
 from git import (
     IndexFile,
@@ -52,8 +53,9 @@ from git.cmd import Git
 
 HOOKS_SHEBANG = "#!/usr/bin/env sh\n"
 
+is_win_without_bash = is_win and not shutil.which('bash.exe')
 
-@skipIf(HIDE_WINDOWS_KNOWN_ERRORS, "TODO: fix hooks execution on Windows: #703")
+
 def _make_hook(git_dir, name, content, make_exec=True):
     """A helper to create a hook"""
     hp = hook_path(name, git_dir)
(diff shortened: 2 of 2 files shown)


Byron:
Thanks a lot for the contribution, much appreciated.

Doing this makes a lot of sense and will generally make life easier on windows. If there is any demand, one could imagine making the default interpreter/runner binary configurable, and using `bash.exe` for now seems like a decent default.

idbrii:
I wrote a test for a python hook to prove to myself it would fail:

```
diff --git a/test/test_index.py b/test/test_index.py
index 233a4c64..d6373250 100644
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -56,14 +56,14 @@ HOOKS_SHEBANG = "#!/usr/bin/env sh\n"
 is_win_without_bash = is_win and not shutil.which('bash.exe')
 
 
-def _make_hook(git_dir, name, content, make_exec=True):
+def _make_hook(git_dir, name, content, make_exec=True, shebang=HOOKS_SHEBANG):
     """A helper to create a hook"""
     hp = hook_path(name, git_dir)
     hpd = osp.dirname(hp)
     if not osp.isdir(hpd):
         os.mkdir(hpd)
     with open(hp, "wt") as fp:
-        fp.write(HOOKS_SHEBANG + content)
+        fp.write(shebang + content)
     if make_exec:
         os.chmod(hp, 0o744)
     return hp
@@ -912,6 +912,33 @@ class TestIndex(TestBase):
         new_commit = index.commit(commit_message)
         self.assertEqual(new_commit.message, "{} {}".format(commit_message, from_hook_message))
 
+    @with_rw_repo('HEAD', bare=True)
+    def test_commit_msg_hook_success_in_py(self, rw_repo):
+        index = rw_repo.index
+        hp = _make_hook(
+            index.repo.git_dir,
+            'commit-msg',
+            "import sys; print('stdout'); print('stderr', file=sys.stderr); sys.exit(1)",
+            shebang = "#!/usr/bin/env python3\n"
+        )
+        try:
+            index.commit("This should fail")
+        except HookExecutionError as err:
+            if is_win_without_bash:
+                self.assertIsInstance(err.status, OSError)
+                self.assertEqual(err.command, [hp])
+                self.assertEqual(err.stdout, '')
+                self.assertEqual(err.stderr, '')
+                assert str(err)
+            else:
+                self.assertEqual(err.status, 1)
+                self.assertEqual(err.command, [hp])
+                self.assertEqual(err.stdout, "\n  stdout: 'stdout\n'")
+                self.assertEqual(err.stderr, "\n  stderr: 'stderr\n'")
+                assert str(err)
+        else:
+            raise AssertionError("Should have cought a HookExecutionError")
+
     @with_rw_repo('HEAD', bare=True)
     def test_commit_msg_hook_fail(self, rw_repo):
         index = rw_repo.index
```

And I was thinking instead of calling bash.exe with the hook, we could call it with a call_from_unix script like this so bash will figure out the right program to use:
```
#! /bin/sh
$1
```

And use:

            cmd = ["bash.exe", "call_from_unix.sh", relative_hp]

I'm currently not sure if this is a good idea or terrible idea.

Byron:
Sorry, actually I forgot to merge yesterday and didn't mean to instill doubt.

I'd say this PR is already an improvement due to the added chance for success.

If there are any issues that the new suggestions with the 'proxy script' (for lack of a better name) would solve that can be implemented, too, but I'd wait. The current solution won't add another program invocation which might be unnecessary, but always costly.

PR #1400, opened 2022-01-21
set unassigned stdin to improve pyinstaller compatibility

smokephil:
To create a window application with pyinstaller, all suprocess input and output streams must be assigned and must not be None.
https://stackoverflow.com/a/51706087/7076612

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -12,7 +12,8 @@ import signal
 from subprocess import (
     call,
     Popen,
-    PIPE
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot. `None` probably means that it inherits stdin from its parent environment, which probably rarely is what anybody wants to rely on when using GitPython.

I hope there is no unintended side-effects, and there is only one way to find out :).

PR #1411, opened 2022-02-23
fix iter_commits comment, more in line with iter_items

houssam7737:
Pretty simple contribution: fix the comment description of the paths param which is currently wrong. 
Original description of what paths does confused me when I was reading through code, this should make it clearer and more in-line with the description in iter_items.


Byron:
It's indeed much clearer than before. Thanks for taking the time to improve it :)!

PR #1416, opened 2022-02-26
adding patch property to Commit class

dduraipandian:
- added patch property to Commit class
- added relevant test cases and tested it.

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -316,6 +316,18 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         return Stats._list_from_string(self.repo, text)
 
     @property
+    def patch(self) -> str:
+        """Get a git patch from changes between this commit and its first parent
+        or from all changes done if this is the very first commit.
+
+        :return: String"""
+        if not self.parents:
+            text = self.repo.git.diff(self.hexsha)
+        else:
+            text = self.repo.git.diff(self.parents[0].hexsha, self.hexsha, '--', numstat=False)
+        return text
+
+    @property
     def trailers(self) -> Dict:
         """Get the trailers of the message as dictionary
 
--- a/test/test_commit.py
+++ b/test/test_commit.py
@@ -149,6 +149,81 @@ class TestCommit(TestCommitSerialization):
         self.assertEqual(commit.committer_tz_offset, 14400, commit.committer_tz_offset)
         self.assertEqual(commit.message, "initial project\n")
 
+    def test_patch_with_parent(self):
+        expected_path_result = """diff --git a/doc/source/tutorial.rst b/doc/source/tutorial.rst
+index bc386e7c..fcbc18bf 100644
+--- a/doc/source/tutorial.rst
++++ b/doc/source/tutorial.rst
+@@ -66,7 +66,7 @@ Archive the repository contents to a tar file.
+ Advanced Repo Usage
+ ===================
+ 
+-And of course, there is much more you can do with this type, most of the following will be explained in greater detail in specific tutorials. Don't worry if you don't understand some of these examples right away, as they may require a thorough understanding of gits inner workings.
++And of course, there is much more you can do with this type, most of the following will be explained in greater detail in specific tutorials. Don't worry if you don't understand some of these examples right away, as they may require a thorough understanding of git's inner workings.
+ 
+ Query relevant repository paths ...
+ 
+@@ -363,7 +363,7 @@ Handling Remotes
+     :start-after: # [25-test_references_and_objects]
+     :end-before: # ![25-test_references_and_objects]
+ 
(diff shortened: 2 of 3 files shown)


test/test_commit.py
```diff
@@ -149,6 +149,81 @@ def check_entries(d):
         self.assertEqual(commit.committer_tz_offset, 14400, commit.committer_tz_offset)
         self.assertEqual(commit.message, "initial project\n")
 
+    def test_patch_with_parent(self):
```
Byron, line 152: Since this is a patch with parent, I think it should be easy to find a patch with is much smaller. If there truly is no way, the expected value could be read from a file.

test/test_commit.py
```diff
@@ -149,6 +149,81 @@ def check_entries(d):
...
+ GitCmdObjectDB
+ ==============
+-The git command database uses persistent git-cat-file instances to read repository information. These operate very fast under all conditions, but will consume additional memory for the process itself. When extracting large files, memory usage will be much higher than the one of the ``GitDB``::
++The git command database uses persistent git-cat-file instances to read repository information. These operate very fast under all conditions, but will consume additional memory for the process itself. When extracting large files, memory usage will be much higher than ``GitDB``::
+ 
+     repo = Repo("path/to/repo", odbt=GitCmdObjectDB)
+ """
+        commit = self.rorepo.commit('c0740570b31f0f0fe499bf4fc5abbf89feb1757d')
+        patch = commit.patch
+        self.assertEqual(patch[:200], expected_path_result[:200])
```
Byron, line 203: I don't understand why it would slice the result into lines like this. This also means it would possibly only test for equivalence on a subset of lines, missing some lines behind the 600 mark, for example.

test/test_commit.py
```diff
@@ -149,6 +149,81 @@ def check_entries(d):
...
++count= True
++statistics = True
++# E265 = comment blocks like @{ section, which it can't handle
++# E266 = too many leading '#' for block comment
++# E731 = do not assign a lambda expression, use a def
++# W293 = Blank line contains whitespace
++# W504"""
+        commit = self.rorepo.commit('33ebe7acec14b25c5f84f35a664803fcab2f7781')
+        patch = commit.patch
+        self.assertEqual(patch[:200], expected_path_result[:200])
```
Byron, line 225: As above, I don't think the slice syntax should be necessary here. If it is, a description on why that is would be necessary, it could be passed as third argument to `assertEqual` I believe.

git/objects/commit.py
```diff
@@ -315,6 +315,18 @@ def stats(self) -> Stats:
             text = self.repo.git.diff(self.parents[0].hexsha, self.hexsha, '--', numstat=True)
         return Stats._list_from_string(self.repo, text)
 
+    @property
+    def patch(self) -> str:
```
Byron, line 319: It probably would be worth describing that this method can fail if the processed bytes aren't in the encoding that python expects (as it returns `str` and not `bytes`). Hence this method can throw an exception and from my experience it's more than likely that this happens to some.

This also means that at some point in the future and if there is demand one could provide a variant of this method that returns bytes instead.

Byron (changes requested):
Thanks a lot for contributing, I think it's nearly there.

There are a few comments, please take a look.

For changes, please don't force-push but instead create new commits which makes reviewing changes much easier. Thank you.

test/test_commit.py
dduraipandian: Thanks for reviewing it. I will try to get commit with a smaller patch.

test/test_commit.py
dduraipandian: I can't test with whole patch text. I have to set maxDiff = None to disable limit and test. For test_with_parent, I will try to get a commit with smaller patch. For test_with_no_parent, which is first commit, I will save patch text to file and set maxDiff in setup and test it.

test/test_commit.py
dduraipandian: have to set maxDiff = None to disable limit and test. I will do the change and test it.

git/objects/commit.py
dduraipandian: Thanks for this note. I will check to see if i can return bytes than string and leave the choice to users to convert them to string. Is that okay?

Byron: That would be okay, yes. If it doesn't work with returning `bytes` right now then the possibility of decoding errors could be highlighted in the documentation specifically.

test/test_commit.py
Byron: Then I think it would be important to make clear how `maxDiff` works in the property's documentation so callers know how to deal with it. To me for instance it wasn't clear this exists.

Or said differently, if there is a technical limitation there is less of a need to work around it from my side, as long as it can be made clear in the docs how to handle it.

PR #1417, opened 2022-03-21
Fixed setting ref with non-ascii in path

ActiveChooN:
Have faced the issue when trying to set a reference to a branch with non-ASCII symbols in the path. [Snippet](<https://github.com/openvinotoolkit/cvat/blob/dd60b2d8e0693772bc637b99a4f6684df887de30/cvat/apps/dataset_repo/dataset_repo.py#L131>).  I guess, this fix would resolve it.


Byron:

PR #1419, opened 2022-04-01
Allow `Repo.create_head`'s `commit` arg to be a `SymbolicReference`

DMRobertson:
This propagates the type of `Head.create`'s `reference` argument. That `create` function is inherited from `SymbolicReference`.

Mypy doesn't pass on my machine, but as far as I can see it reports no new errors compared to the CI runs on the main branch.


Byron:
Thank you, that seems like a great quality of life improvement for those using mypi.

Issue #1377, opened 2021-11-12
git show result misses newline at end of diff


markhh80:
The result of git.Git.show (...) misses newlines ("\n") at the end of a file.

**Reproduction**

1) setup git repo / use existing
2) add empty textfile "foo.txt", commit file
3) modify file by inserting string "foo\r\n", commit changes
4) run git cmd "show [commit sha]:foo.txt"
5) ... via GitPython, example:
  
  ```
  # commits = ...
  repo.git.show(f"{commits[0].hexsha}:{testfile}")
  ```
6) ... from Git terminal, pipe through cat etc. to see control characters, example: 

  ```
  git show 74e70e7:foo.txt | cat -vet | less
  ```

**Observation**

The string returned from show via GitPython has a trailing "\r", missing the "\n". The result from terminal execution has a trailing "\r\n".

**Expected**

Both outputs have the correct trailing "\r\n".

**Addendum**

- run GitPython cmd with option "with_extended_output" -> same result
- run GitPython cmd with option "stdout_as_string" -> same result (but in binary format)
- crosscheck: run "git show" via subprocess.popen -> output has correct "\r\n"

  ```
  # commits = ...
  p = subprocess.Popen(["git", "show", f"{commits[0].hexsha}:{testfile}"], stdout=subprocess.PIPE)
  out, err = p.communicate()
  print(repr(out))
  ```

Byron:
Thanks a lot for the wonderful description of the issue and for spending the time that undoubtedly when into it.

It's likely that GitPython strips `git` output somewhere in its command execution machinery, and I wonder why it was added in the first place.

toku-sa-n:
Hello, I've also encountered this problem (see my [SO question](https://stackoverflow.com/questions/71672179/the-file-is-not-a-zip-file-error-for-the-output-of-git-show-by-gitpython) and the answer). I found the line where GitPython strips the last `\n`.
https://github.com/gitpython-developers/GitPython/blob/0b33576f8e7add5671f8927dff228e7f92eec076/git/cmd.py#L946-L948
The test below fails for the current main branch, and succeeds after removing `[:-1]`.
```python
    @with_rw_directory
    def test_git_show_last_newline(self, rw_dir):
        r = Repo.init(rw_dir)
        fp = osp.join(rw_dir, 'hello.txt')
        with open(fp, 'w') as fs:
            fs.write("hello\n")
        r.git.add(Git.polish_url(fp))
        r.git.commit(message="init")
[...]

PR #1423, opened 2022-04-06
feat(cmd): add the `strip_newline` flag

toku-sa-n:
This PR adds the `strip_newline` flag to the `Git.execute` method. When this flag is set to `True`, it will trim the trailing `\n`. The default value is `True` for backward compatibility. Setting it to `False` is helpful for, e.g., the `git show` output, especially with the binary file, as the missing `\n` may invalidate the file.

Looking at the first some bytes and checking if the output is binary or not is another option to prevent making an invalid binary file. Still, I chose to add a parameter because it is easy to implement, is fully backward compatible, and doesn't add many lines to the source code.

Fixes: #1377

By the way, do we need to handle stderr too?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,4 +45,5 @@ Contributors are:
 -Alba Mendez <me _at_ alba.sh>
 -Robert Westman <robert _at_ byteflux.io>
 -Hugo van Kemenade
+-Hiroki Tokunaga <tokusan441 _at_ gmail.com>
 Portions derived from other open source works and are clearly marked.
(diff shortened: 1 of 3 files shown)


Byron (commented):
Thanks a lot for the fix!

Regarding stdout/stderr: `strip_newline` might imply a newline is stripped universally, even though it applies to `stdout` only at the moment.

Maybe it should be called `strip_newline_in_stdout` to be specific while allowing to add a similar flag for `stderr` when there is a need?

PS: CI had issues but after restarting it the error now seems 'legit' (as it is upset about whitespace :/)

Byron (approved):
Thank you, looks great to me!

PR #1437, opened 2022-05-04
Strip usernames from URLs as well as passwords

### LINKED ISSUE #1284: Github token is leaked when used as a part of remote URL
Github [allows](https://github.blog/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/) usage of personal access tokens with remote URLs in the format:

```
https://<token>@github.com/owner/repo.git
```
or
```
https://<token>:x-oauth-basic@github.com/owner/repo.git
```

If such URL is used with `Remote`, if some error related to the access happens, it (in the example below token is `foobar`) can be leaked via exception message, like the following:

```
  File "start.py", line 119, in <module>
    sys.exit(run_main())
    ...
  cmdline: git push --set-upstream --force ....
  stderr: 'remote: some error.
fatal: unable to access 'https://foobar:x-oauth-basic@github.com/FooBarCorp/some-repo/': The requested URL returned error: 403'
```

It would be nice to have internal mechanism to prevent such leakage.

glennmatthews:
Fixes #1284 

- Extend `remove_password_if_present` to remove usernames from URLs as well as the existing behavior of removing passwords from URLs.
- As an extra level of caution, I extended `CommandError.__init__` to also call `remove_password_if_present` just in case an exception gets raised from a non-sanitized command by mistake.
- Add test coverage for both of the above changes.

### DIFF
--- a/git/exc.py
+++ b/git/exc.py
@@ -8,6 +8,7 @@
 from gitdb.exc import BadName  # NOQA @UnusedWildImport skipcq: PYL-W0401, PYL-W0614
 from gitdb.exc import *     # NOQA @UnusedWildImport skipcq: PYL-W0401, PYL-W0614
 from git.compat import safe_decode
+from git.util import remove_password_if_present
 
 # typing ----------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


Byron:

I particularly like that there are tests for all changes (which I assume break without these modifications).

PR #1440, opened 2022-05-17
fix: Allow adding PathLike object to index

### LINKED ISSUE #1382: repo.index.add() first argument type accepts os.PathLike but code raises TypeError
In the docs and in the code types, the first argument for `repo.index.add()` is `Sequence[Union[str, os.PathLike, git.objects.blob.Blob, git.index.typ.BaseIndexEntry, Submodule]]`.
However, actually trying to submit a list of Paths to this argument results in a TypeError, as it seems the code inside `_preprocess_add_items` does not handle Paths.

I think this could be fixed by changing line 611 inside of `_preprocess_add_items` in `base.py` to:
```
 if isinstance(item, str) or isinstance(item, os.PathLike):
```
https://github.com/gitpython-developers/GitPython/blob/fe54118ec07a68d5dc6f6108510cffc55dfca643/git/index/base.py#L611
Since `_to_relative_path()` already accepts pathlike arguments.

rdbisme:
Close #1382

Apart from the automatic linting, relevant lines are [here](https://github.com/rdbisme/GitPython/blob/index-pathlike/git/index/base.py#L587-L591) + the additional test

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -617,11 +617,11 @@ class IndexFile(LazyMixin, git_diff.Diffable, Serializable):
         paths = []
         entries = []
         # if it is a string put in list
-        if isinstance(items, str):
+        if isinstance(items, (str, os.PathLike)):
             items = [items]
 
         for item in items:
-            if isinstance(item, str):
+            if isinstance(item, (str, os.PathLike)):
                 paths.append(self._to_relative_path(item))
             elif isinstance(item, (Blob, Submodule)):
                 entries.append(BaseIndexEntry.from_blob(item))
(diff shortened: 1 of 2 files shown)


Byron:
May I ask to resubmit without the code formatting? I am happy to run it through `black` to unify the style and if it doesn't make it (too much) worse, but that should be a separate step.

All these changes are like noise to the signal making a review harder than it has to be. Thanks.

Byron (changes requested):
May I ask to adjust the commits to not do code formatting? I am happy to run it through black to unify the style and if it doesn't make it (too much) worse, but that should be a separate step.

All these changes are like noise to the signal making a review harder than it has to be. Thanks.

rdbisme:
Not sure why this is failing.

PR #1441, opened 2022-05-17
feat: If `core.autocrlf` is enabled, replace CRLF with LF when adding to index

rdbisme:
This is to mimic the same behaviour the standard git command has. 

Apart for the blackification, relevant lines are the definition of the class `_FileStore`, the addition of the method `_autocrlf`, and the adaptations in `add()` to use the `_FileStore` contextmanager.

The logic here is to replace `\r\n` with `\n` when adding to the index, then reset the local file to the original content.

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -38,6 +38,8 @@ from gitdb.db import MemoryDB
 
 import git.diff as git_diff
 import os.path as osp
+from pathlib import Path
+from typing import Optional
 
 from .fun import (
     entry_key,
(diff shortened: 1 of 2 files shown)


rdbisme:
It would be much easier to contribute if using `black` for style autolinting

Byron:
May I ask to adjust the commits to not do code formatting? I am happy to run it through `black` to unify the style and if it doesn't make it (too much) worse, but that should be a separate step.

All these changes are like noise to the signal making a review harder than it has to be. Thanks.

Byron (changes requested):
May I ask to adjust the commits to not do code formatting? I am happy to run it through black to unify the style and if it doesn't make it (too much) worse, but that should be a separate step.

All these changes are like noise to the signal making a review harder than it has to be. Thanks.

rdbisme:
Thanks for taking the time of applying black. I rebased everything on top of the latest commit and now the changes should more clear :)

Byron:
Thanks for the contribution and the intention to fix a shortcoming.
However, I don't feel comfortable to add 'a very special filter' that will inadvertently affect current users that build upon the existing, very limited behaviour. It has a huge potential to cause breakage and thus I see no way to add it in its current form.

GitPython has been done very much in that fashion - implement something to mimic the actual behaviour of git - which has caused an incredible amount of problems and surprises down the road. It's just not a paradigm that should be continued.

I am sorry, but I see no way forward for this PR, and thank you for your understanding.

rdbisme:
> Thanks for the contribution and the intention to fix a shortcoming. However, I don't feel comfortable to add 'a very special filter' that will inadvertently affect current users that build upon the existing, very limited behaviour. It has a huge potential to cause breakage and thus I see no way to add it in its current form.
> 
> GitPython has been done very much in that fashion - implement something to mimic the actual behaviour of git - which has caused an incredible amount of problems and surprises down the road. It's just not a paradigm that should be continued.
> 
> I am sorry, but I see no way forward for this PR, and thank you for your understanding.

Ok, no problem. Thanks for reviewing! :)

PR #1455, opened 2022-06-11
Re-enable Cygwin CI and get most tests passing

DWesl:
It looks like the Cygwin CI testing was disabled a while back, possibly as part of the move to GHA.  I added a separate Cygwin CI run to the GitHub actions and tried to get the tests passing.  I have three test failures left that I don't understand, and hope someone here knows what's going on.

I can move the test to a separate step in the existing GHA workflow if that would work better in this project, or move it to a separate PR to wait for the last three tests to pass while the other fixes go in.

### DIFF
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -24,6 +24,9 @@ jobs:
     - uses: cygwin/cygwin-install-action@v2
       with:
         packages: python39 python39-pip git
+    - name: Tell git to trust this repo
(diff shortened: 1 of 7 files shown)


Byron (changes requested):
Thanks a lot for putting in all this work and make `GitPython` work on cygwin once again.

Regarding the failing tests, I'd be OK with ignoring them if run on cygwin and providing a fix in other PRs. The failures are all related to submodule handling, and I think that functionality isn't necessarily widely used anymore.

PR #1459, opened 2022-06-21
Fix blob filter types

AustinScola:
Fix the types and type annotations of some of the blob filter code.

### DIFF
--- a/git/index/typ.py
+++ b/git/index/typ.py
@@ -57,7 +57,7 @@ class BlobFilter(object):
         for pathlike in self.paths:
             path: Path = pathlike if isinstance(pathlike, Path) else Path(pathlike)
             # TODO: Change to use `PosixPath.is_relative_to` once Python 3.8 is no longer supported.
-            if all(map(lambda t: t[0] == t[1], zip(path.parts, blob_path.parts))):
+            if all(i == j for i, j in zip(path.parts, blob_path.parts)):
                 return True
         return False
 
--- a/git/index/typ.py
+++ b/git/index/typ.py
@@ -4,6 +4,7 @@ from binascii import b2a_hex
 
 from .util import pack, unpack
 from git.objects import Blob
+from git.index.base import StageType
 
 
 # typing ----------------------------------------------------------------------
@@ -48,10 +49,10 @@ class BlobFilter(object):
         """
         self.paths = paths
 
-    def __call__(self, stage_blob: Blob) -> bool:
-        path = stage_blob[1].path
+    def __call__(self, stage_blob: Tuple[StageType, Blob]) -> bool:
+        path: str = str(stage_blob[1].path)
         for p in self.paths:
-            if path.startswith(p):
+            if path.startswith(str(p)):
                 return True
         # END for each path in filter paths
         return False
--- a/git/index/typ.py
+++ b/git/index/typ.py
@@ -4,7 +4,6 @@ from binascii import b2a_hex
 
 from .util import pack, unpack
 from git.objects import Blob
-from git.index.base import StageType
 
 
 # typing ----------------------------------------------------------------------
@@ -16,9 +15,11 @@ from git.types import PathLike
 if TYPE_CHECKING:
     from git.repo import Repo
 
+StageType = int
+
 # ---------------------------------------------------------------------------------
 
-__all__ = ("BlobFilter", "BaseIndexEntry", "IndexEntry")
+__all__ = ("BlobFilter", "BaseIndexEntry", "IndexEntry", "StageType")
 
 # { Invariants
 CE_NAMEMASK = 0x0FFF
--- a/git/index/typ.py
+++ b/git/index/typ.py
@@ -9,7 +9,7 @@ from git.objects import Blob
 
 # typing ----------------------------------------------------------------------
 
-from typing import NamedTuple, Sequence, TYPE_CHECKING, Tuple, Union, cast
+from typing import NamedTuple, Sequence, TYPE_CHECKING, Tuple, Union, cast, List
 
 from git.types import PathLike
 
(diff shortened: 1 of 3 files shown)


git/index/typ.py
```diff
@@ -48,10 +50,10 @@ def __init__(self, paths: Sequence[PathLike]) -> None:
         """
         self.paths = paths
 
-    def __call__(self, stage_blob: Blob) -> bool:
-        path = stage_blob[1].path
+    def __call__(self, stage_blob: Tuple[StageType, Blob]) -> bool:
+        path: str = str(stage_blob[1].path)
```
Byron, line 54: Paths can't be converted to `str` as this implies an interpreter defined encoding.

git/index/typ.py
```diff
@@ -48,10 +50,10 @@ def __init__(self, paths: Sequence[PathLike]) -> None:
...
         for p in self.paths:
-            if path.startswith(p):
+            if path.startswith(str(p)):
```
Byron, line 56: There has to be a way to do this without `str(…)`, please see above for the reason.

Byron (changes requested):
Thanks a lot, the change will be appreciated! 

There is some usage of `str(…)` conversions which requires attention, but with that fix there should be nothing in the way of merging.

git/index/typ.py
AustinScola: Can you explain what you mean by this or give an example?

Byron: Before the change it looked like this:

```
path = stage_blob[1].path
```

And after the change it should still be looking like this but possibly with type annotations added to please the type checker. Creating new strings with `str(…)` changes the way the program functions, and does so in a breaking way due to encodings being unspecified for paths on linux.

AustinScola: Changed to use paths instead of strs. Still not really sure what you mean by encodings being unspecified for paths on linux.

git/index/typ.py
```diff
@@ -48,12 +51,14 @@ def __init__(self, paths: Sequence[PathLike]) -> None:
...
-        path = stage_blob[1].path
-        for p in self.paths:
-            if path.startswith(p):
+    def __call__(self, stage_blob: Tuple[StageType, Blob]) -> bool:
+        blob_pathlike: PathLike = stage_blob[1].path
+        blob_path: Path = blob_pathlike if isinstance(blob_pathlike, Path) else Path(blob_pathlike)
+        for pathlike in self.paths:
+            path: Path = pathlike if isinstance(pathlike, Path) else Path(pathlike)
+            # TODO: Change to use `PosixPath.is_relative_to` once Python 3.8 is no longer supported.
+            if all(i == j for i, j in zip(path.parts, blob_path.parts)):
```
Byron, line 60: Is this not another way of writing `path1 == path2`, since all path components (here `.parts`)  have to be equivalent? The benefit of doing so of course is that it ignores which path separator is used, but I am having trouble to understand how this checks that `blob_path` starts with one of `self.paths`.

Byron (commented):
Thanks a lot for the changes and the continued effort! I think using `Path` _should_ be correct in terms of required encodings, at least it's as good as it gets. 

I also understand that just converting to a known type temporarily is probably the easiest than having to try to support all possible input types, so let's go with that even though it seems _potentially_ wasteful.

There is one more issue I seem to be seeing, and am looking forward to hearing your opinion.

git/index/typ.py
AustinScola: I think this should work because `zip` stops when the shortest of the arguments is exhausted. So it is only checking if _all_ the path components are equivalent in the case that the length of the `parts` are equal. But, when one path has more parts than the other, then it is checking that the longer path starts with the shorter path.

Example:
```python3
def startswith(a, b): return all(i == j for i, j in zip(a, b))

a = [1, 2, 3]
b = [1, 2]
assert startswith(a, b)
```

Byron: Thanks for the explanation, I see how this works now and why this is correct. The alternative has to remain a TODO until it becomes available, and looking at the code I am happy to suggest to not ever go there - the try:except clause seems like a deal-breaker to me. It's entirely up to you though.

Byron: Actually, let me undo what I last said. I once again think this implementation is incorrect, as these paths are not equivalent.

```python
base = [1, 2, 3]
added_path = [1, 2]
```

The above would match even though `added_path` is not contained in base.

AustinScola: Good catch, I pushed a change that addresses this and adds a test too.

test/test_blob_filter.py
```diff
@@ -0,0 +1,31 @@
...
+import pytest
+
+from git.index.typ import BlobFilter, StageType
+from git.objects import Blob
+from git.types import PathLike
+
+
+# fmt: off
+@pytest.mark.parametrize('paths, stage_type, path, expected_result', [
+    ((Path("foo"),), 0, Path("foo"), True),
```
Byron, line 15: If `stage_type` is always 0, I don't think it needs to be a parameter. Did you mean to vary the stage?

Byron (commented):
Thanks a lot for the test and the fix. It's definitely an under-tested area of GitPython as the issue wasn't detected by any test prior.

I left a comment, but when clarified, this PR is ready to be merged.

test/test_blob_filter.py
AustinScola: You are right that it doesn't need to be a parameter. I have a habbit of adding parameters preemptively that the function might use eventually. I actually don't know what the stage type is but it felt wrong to bury a hardcoded value for it in the function.

Byron: >  I actually don't know what the stage type is but it felt wrong to bury a hardcoded value for it in the function.

I think it's OK to make it a constant since it is constant, or see if varying it via a parameter affects the outcome. If not, the latter could be done and the PR can be merged. Otherwise, please go with the constant. Thanks for your help.

AustinScola: Done

Byron:
Thanks a lot for your contribution and going through the review process with me! I hope the code is more robust now.

PR #1460, opened 2022-06-24
Fix bug in TagReference.create that would duplicate the message

ruuda:
Previously if you passed the `message` kwarg, that would cause a `--message` argument to `git tag`, but then the code also added a `-m`, which caused the message to appear in the tag twice. Fix that by deleting `message` from the kwargs if we set `-m`.


Byron:
Very nicely done, thanks a lot for this contribution!

PR #1474, opened 2022-07-31
Fix incomplete typehinting for PathLike

### LINKED ISSUE #1473: git.types.PathLike is partially unknown for Pylance/mypy
Hello everyone,

I am using GitPython in my project and I use Pylance/mypy with the strict type checking settings. This setting is pretty important for my project's sanity.

As I've been using methods from GitPython that make uses of paths, I've been getting the following errors:
mypy raising an error for git.types.PathLike
As you can notice, what raises this error is `PathLike[Unknown]`, defined in `git.types.PathLike`. ([Source file & lines](https://github.com/gitpython-developers/GitPython/blob/main/git/types.py#L47-L51))

After further investigation, I've found out that this is because `os.PathLike` is considered as a generic class (Like `typing.List`, `typing.Dict` and so on)
Image showing PathLike being a generic class

The fix is fairly simple, as it is to simply allow `str` for `PathLike`, which gives this result: `os.PathLike[str]`

PathLike with the implemented fix
Actual code without any errors
(awesome! This issue's no longer a thing with this fix!)

There has some discussion about the same situation on this StackOverflow page: https://stackoverflow.com/a/69680089/13720113

I can make a PR for fixing this (Of course, with some testing, since there is Python version checks), as long as this issue is approved.

Thank you, hoping this issue get sorted soon!

Predeactor:
This adds a more rich type hinting for the `git.types.PathLike` class.
Closes #1473.

Also remove unused imports in this file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,4 +46,5 @@ Contributors are:
 -Robert Westman <robert _at_ byteflux.io>
 -Hugo van Kemenade
 -Hiroki Tokunaga <tokusan441 _at_ gmail.com>
+-Julien Mauroy <pro.julien.mauroy _at_ gmail.com>
 Portions derived from other open source works and are clearly marked.
--- a/git/types.py
+++ b/git/types.py
@@ -40,10 +40,10 @@ else:
 
 
 if sys.version_info[:2] < (3, 9):
-    PathLike = Union[str, os.PathLike[str]]
+    PathLike = Union[str, os.PathLike]
 else:
     # os.PathLike only becomes subscriptable from Python 3.9 onwards
-    PathLike = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
+    PathLike = Union[str, os.PathLike[str]]
 
 if TYPE_CHECKING:
     from git.repo import Repo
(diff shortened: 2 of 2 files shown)


Predeactor:
I'm unsure that the check has failed due to the change because I am unable to run flake8 myself for some reasons... Even after rolling back my changes
image
I'm confused, and might require a hand on how to fix the following...

Byron:
Thanks for adding `bytes`, in theory it _should_ now be ready. I myself am not sure what's going on with flake8 except that it's probably buggy in the latest version.

Maybe this PR can be re-run occasionally to see if that fixes it as it should pull the latest version and fixed version of flake at some point.

Predeactor:
Hello, please do not merge this PR as it appears the current fix might not
be correct.
As per the current fix, Pylance/mypy will complain that a str object does
not comply with the type hint. I assume it would be the same error raised
if I pass a byte object, but complaining it's not a str this time, I assume.

I need to investigate and see what else I can do, but this might be tricky.

Byron:
Thanks a lot for figuring this out with me!

Predeactor:
You're welcome!

PR #1482, opened 2022-08-22
Add commit co-authors support

itsluketwist:
Add co_authors property to the Commit object, which parses the commit message for designated co-authors, include a simple test.

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -4,6 +4,7 @@
 # This module is part of GitPython and is released under
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
 import datetime
+import re
 from subprocess import Popen, PIPE
 from gitdb import IStream
 from git.util import hex_to_bin, Actor, Stats, finalize_process
@@ -738,3 +739,24 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         return self
 
     # } END serializable implementation
+
+    @property
+    def co_authors(self) -> List[Actor]:
+        """
+        Search the commit message for any co-authors of this commit.
+        Details on co-authors: https://github.blog/2018-01-29-commit-together-with-co-authors/
+
+        :return: List of co-authors for this commit (as Actor objects).
+        """
+        co_authors = []
+
+        if self.message:
+            results = re.findall(
+                r"^Co-authored-by: ((?:\w|\-| ){0,38}) <(\S*)>$",
+                self.message,
+                re.MULTILINE,
+            )
+            for author in results:
+                co_authors.append(Actor(*author))
+
+        return co_authors
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -752,11 +752,11 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
 
         if self.message:
             results = re.findall(
-                r"^Co-authored-by: ((?:\w|\-| ){0,38} <\S*>)$",
+                r"^Co-authored-by: (.*) <(.*?)>$",
                 self.message,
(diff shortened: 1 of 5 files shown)


git/objects/commit.py
```diff
@@ -738,3 +739,24 @@ def _deserialize(self, stream: BytesIO) -> "Commit":
...
+        Search the commit message for any co-authors of this commit.
+        Details on co-authors: https://github.blog/2018-01-29-commit-together-with-co-authors/
+
+        :return: List of co-authors for this commit (as Actor objects).
+        """
+        co_authors = []
+
+        if self.message:
+            results = re.findall(
+                r"^Co-authored-by: ((?:\w|\-| ){0,38}) <(\S*)>$",
```
Byron, line 755: For consistency and probably correctness, one could either use [this regex](https://github.com/gitpython-developers/GitPython/blob/main/git/util.py#L690:L690) or instantiate an actor [like this](https://github.com/gitpython-developers/GitPython/blob/main/test/test_actor.py#L13:L13).

test/test_commit.py
```diff
@@ -509,3 +509,14 @@ def test_trailers(self):
         assert KEY_1 not in commit.trailers.keys()
         assert KEY_2 in commit.trailers.keys()
         assert commit.trailers[KEY_2] == VALUE_2
+
+    def test_commit_co_authors(self):
+        commit = copy.copy(self.rorepo.commit("4251bd5"))
+        commit.message = """Commit message
+
+Co-authored-by: Test User 1 <602352+test@users.noreply.github.com>
+Co-authored-by: test_user_2 <another_user-email@.github.com>"""
```
Byron, line 518: Especially when `Actor` code is used for parsing it would be interesting to see what happens if there is a malformed one in the list. Will that be ignored, or will the whole operation fail?

Byron (changes requested):
Thanks a lot, this looks good already!

I have some notes, but when clarified it should be ready for merging.

git/objects/commit.py
itsluketwist: Sure thing! I kept the same regex, to ensure that the correct prefix is used, and then instantiated an Actor as requested.

test/test_commit.py
itsluketwist: It should be ignored, as the regex will simply not find it, expanded the test to show this. 😄

git/objects/commit.py
```diff
@@ -752,11 +752,11 @@ def co_authors(self) -> List[Actor]:
 
         if self.message:
             results = re.findall(
-                r"^Co-authored-by: ((?:\w|\-| ){0,38}) <(\S*)>$",
+                r"^Co-authored-by: ((?:\w|\-| ){0,38} <\S*>)$",
```
Byron, line 755: To clarify, can you explain where this regex is coming from? Earlier I tried to argue that the one used by `Actor` could be used here. It's much more open than this, but I can be swayed if there is a specification of the `Co-authored-by:` field that limits it.

Byron (commented):
Thanks a lot!

The only remaining question pertains to the regex used, but once clarified this is ready to go.
Maybe you could also fix flake8 by ignoring this new check that makes CI fail, if you feel comfortable.

git/objects/commit.py
itsluketwist: Ah, my bad! Regarding the name part, this is on me, I used a regex to match against the GitHub username specs, but this is just the account name, so a more generic one is necessary - switched to match the regex used in `Actor` (with the extra `Co-authored-by: ` text and start/end line checks).

Also added a line to ignore the flake8 check - not used flake8 but I think this should be right.

Byron:

PR #1486, opened 2022-08-31
Suggested fixes by iCR, OpenRefactory, Inc.

openrefactory:
This issue was detected in branch `main` of `GitPython` project on the version with commit hash `73bde1`. This is an instance of an improper method call

**Fixes for improper method call:** 
In file: `base.py`, there is a method `from_tree` that creates a temporary file using an unsafe API `mktemp`. The use of this method is discouraged in the [Python documentation. ](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp)**iCR** suggested that a temporary file should be created using `NamedTemporaryFile` which is a [a safe API. ](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)**iCR** replaced the usage of `mktemp` with `NamedTemporaryFile`.

In file: `util.py`, there is a method `__init__` that creates a temporary file using an unsafe API `mktemp`. The use of this method is discouraged in the [Python documentation. ](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp)**iCR** suggested that a temporary file should be created using `NamedTemporaryFile` which is a [a safe API. ](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)**iCR** replaced the usage of `mktemp` with `NamedTemporaryFile`.

This issue was detected by **OpenRefactory's Intelligent Code Repair (iCR)**. We are running **iCR** on libraries in the `PyPI` repository to identify issues and fix them. More info at: [pypi.openrefactory.com](https://pypi.openrefactory.com/)

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -4,6 +4,7 @@
 # This module is part of GitPython and is released under
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
 
+from ast import Delete
 import glob
 from io import BytesIO
 import os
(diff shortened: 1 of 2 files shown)


Byron:
It's by design, the temporary file has to be opened and written by `git` in a separate process, and is later renamed. If a human is going to make this PR work though which I doubt, please feel free to open a new one with a personal note.

EliahKagan:
In case anyone finds this when searching, this was effectively fixed (among other changes):

- In `git.index.base` by [#1606](https://github.com/gitpython-developers/GitPython/pull/1606), refined in [#1751](https://github.com/gitpython-developers/GitPython/pull/1751).
- In `git.index.util` by [#1770](https://github.com/gitpython-developers/GitPython/pull/1770), refined in [#1776](https://github.com/gitpython-developers/GitPython/pull/1776).

Issue #1490, opened 2022-09-12
`invalid literal for int() with base 10: 'n'` during a diff operation


langfield:
Here is the stack trace.

```python
ERROR:git.cmd:Pumping 'stdout' of cmd(['git', 'diff-tree', 'a31de9c31a0ee03574a34146aa1a986648ac499b', '7a62039cb6ce50e5d4c8fc412fdaf10e6f92b823', '-r', '--abbrev=40', '--full-index', '-M', '--raw', '-z', '--no-color']) failed due to: ValueError("invalid literal for int() with base 10: 'n'")
Exception in thread Thread-12:
Traceback (most recent call last):
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/cmd.py", line 114, in pump_stream
  handler(line)
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/diff.py", line 570, in <lambda>
  handle_process_output(proc, lambda byt: cls._handle_diff_line(byt, repo, index),
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/diff.py", line 557, in _handle_diff_line
  diff = Diff(repo, a_path, b_path, a_blob_id, b_blob_id, old_mode, new_mode,
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/diff.py", line 303, in __init__
  self.a_mode = mode_str_to_int(a_mode) if a_mode else None
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/objects/util.py", line 80, in mode_str_to_int
  mode += int(char) << iteration * 3
ValueError: invalid literal for int() with base 10: 'n'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/home/user/conda/envs/anki/lib/python3.9/threading.py", line 973, in _bootstrap_inner
  self.run()
File "/home/user/conda/envs/anki/lib/python3.9/threading.py", line 910, in run
  self._target(*self._args, **self._kwargs)
File "/home/user/conda/envs/anki/lib/python3.9/site-packages/git/cmd.py", line 120, in pump_stream
  raise CommandError([f'<{name}-pump>'] + remove_password_if_present(cmdline), ex) from ex
git.exc.CommandError: Cmd('<stdout-pump>') failed due to: ValueError('invalid literal for int() with base 10: 'n'')
[...]

Issue #1483, opened 2022-08-25
Problem when using Diff


bobeatslollipop:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/git/cmd.py", line 114, in pump_stream
    handler(line)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/git/diff.py", line 570, in <lambda>
    handle_process_output(proc, lambda byt: cls._handle_diff_line(byt, repo, index),
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/git/diff.py", line 557, in _handle_diff_line
python-BaseException
    diff = Diff(repo, a_path, b_path, a_blob_id, b_blob_id, old_mode, new_mode,
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/git/diff.py", line 303, in __init__
    self.a_mode = mode_str_to_int(a_mode) if a_mode else None
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/git/objects/util.py", line 80, in mode_str_to_int
    mode += int(char) << iteration * 3
ValueError: invalid literal for int() with base 10: 'x'

This is my first time posting an issue, so please tell me what extra information is needed!

Byron:
It couldn't parse the output of this command:

```
git diff-tree 81a1f0c4f84aeb6ee97edea64251419515eff571 8c38dfe86157463d745c453d9a1d8c066cc9b7c5 -r --abbrev=40 --full-index -M --raw -z --no-color
```

If executed by hand, it should yield the output which ideally is posted here. That way it should become clear why it tries to parse `x` as integer.

langfield:
This was fixed by #1491.

EliahKagan:
This has just been closed as not planned, but wasn't it actually fixed in #1491?

PR #1491, opened 2022-09-12
Fix bug where colons in paths raise a `ValueError` on `diff()` calls.

langfield:
This commit introduces a potential fix for #1490 and #1483, in which an `invalid literal for int() with base 10: 'n'` exception was raised within a diff operation. Within `_handle_diff_line()`, we split the output of `git diff-tree` on colons (`:` characters), under the assumption that there are no colons within the paths of the files being diffed. On POSIX systems this is not a valid assumption. The fix is to split on `\x00:`, since a null character always precedes the colons we actually need to split on.

A test already existed for this case (`test_diff_file_with_colon()`), but it was marked as skipped.

* Split on `\x00:` instead of `:` in `_handle_diff_line()`.
* Unskip `test_diff_file_with_colon()`.

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -570,7 +570,10 @@ class Diff(object):
     def _handle_diff_line(lines_bytes: bytes, repo: "Repo", index: DiffIndex) -> None:
         lines = lines_bytes.decode(defenc)
 
-        for line in lines.split(":")[1:]:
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot! Amazing to see how close this issue was to a resolution all the time, and I am glad it's finally here now!

PR #1497, opened 2022-09-27
Ignore empty info in diff line

dwapstra:
In some cases, the git diff command does not return any output, this results in an exception in the diff line handling code.

```
  File "../lib/python3.8/site-packages/git/diff.py", line 502, in _handle_diff_line
    old_mode, new_mode, a_blob_id, b_blob_id, _change_type = meta.split(None, 4)
ValueError: not enough values to unpack (expected 5, got 1)
```

This fix ignores empty diff lines.

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -574,6 +574,9 @@ class Diff(object):
         _, _, lines = lines.partition(":")
 
         for line in lines.split("\x00:"):
+            if not line:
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot. This is probably an example for 'the perfect PR': Bite-sized, with concise description, and CI is happy, too :).

PR #1501, opened 2022-10-13
Add datetime.datetime type to commit_date and author_date

SergeantMenacingGarlic:
Update type hints to match acceptable date types.

Passing a `datetime.datetime` object for `author_date` or `commit_date` causes a mypy type error, these dates are passed to `parse_date` in `git.objects.util`, which does accept a `datetime.datetime` object.


Byron:

PR #1508, opened 2022-11-02
remove forced verbose output for pull on a remote

gal-leib:
Noticed that there is no verbose option when performing pull on a remote, and it is hardcoded to be verbose. Adding a keyword argument, similar to this [change](https://github.com/gitpython-developers/GitPython/pull/1005)

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -975,6 +975,7 @@ class Remote(LazyMixin, IterableObj):
         self,
         refspec: Union[str, List[str], None] = None,
         progress: Union[RemoteProgress, "UpdateProgress", None] = None,
+        verbose: bool = True,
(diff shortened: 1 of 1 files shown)


Byron:

CI fails apparently because `pytest` breaks with python 3.7 and under cygwin and I have no idea what the cause of this is. Maybe it's pulling in the most recent version of pytest and it's broken?

Rerunning all jobs yielded the same issue with python 3.8, so I guess it's permanent now :/ - any help with this is definitely appreciated.

Byron:
This PR has been inactive for quite a while and it's being closed due to inactivity. Feel free to re-open it once there is time to bring it up-to-date with `main`.

PR #1522, opened 2022-12-28
Document PushInfoList

skinitimski:
https://github.com/gitpython-developers/GitPython/pull/1360 was a great change, but it went undocumented, leading to [continued confusion](https://github.com/gitpython-developers/GitPython/issues/621#issuecomment-1093834110) about how to use the new feature.

This PR
* Adds docstring for `PushInfoList` so that it will show up in the API search index (currently it is [completely absent](https://gitpython.readthedocs.io/en/stable/search.html?q=PushInfoList&check_keywords=yes&area=default)).
* Changes type hint of the return value of `push` to match what is actually returned.
* Corrects the docstring for return value of `push`.

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -278,6 +278,10 @@ class PushInfo(IterableObj, object):
 
 
 class PushInfoList(IterableList[PushInfo]):
+    """
+    IterableList of PushInfo objects.
+    """
+
     def __new__(cls) -> "PushInfoList":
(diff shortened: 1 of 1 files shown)


Byron:
Thanks so much for improving the docs! I will try harder to watch out for issues like these in future. Part of the problem is me doing this from the browser exclusively which only provides a limited view.

PR #1523, opened 2022-12-29
Fix type hint on create_tag

drewcassidy:
pycharm yells at me without this


Byron:
Thanks a lot!

Please keep these patches coming as you encounter them.

PR #1525, opened 2022-12-31
tests: Use `command -v` instead of third-party `which` program

mgorny:
Use `command -v` to locate the git executable instead of `which`. The former is guaranteed to always be available according to POSIX, while which(1) is a redundant third-party tool that is slowly being phased out from Linux distributions.  In particular, Gentoo no longer installs it by default.


Byron:
Thanks a lot! Much appreciated.

mgorny:
Thanks!

PR #1531, opened 2023-01-08
use tempfile.TemporaryDirectory & fix clone_from_unsafe_protocol tests

obfusk:
* uses `tempfile.TemporaryDirectory()` instead of `tempfile.mkdtemp()` to not leave temporary files behind (though that still happens in some other tests that are harder to fix);
* fixes a typo (procol instead of protocol);
* replaces `/tmp/pwn` w/ `{tmp_file}`;
* clones into `tmp_dir / "repo"` instead of `tmp_dir`, since `tmp_dir / "pwn"` will never exist afterwards otherwise;
* adds a test w/ `ext::sh` that actually allows the protocol, to ensure we're not missing anything in the other tests.

### DIFF
--- a/test/test_remote.py
+++ b/test/test_remote.py
@@ -694,259 +694,279 @@ class TestRemote(TestBase):
 
     @with_rw_repo("HEAD")
     def test_set_unsafe_url(self, rw_repo):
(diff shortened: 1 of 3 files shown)


Byron:

PR #1532, opened 2023-01-08
Fix some resource leaks by open file handles

marlamb:


Byron:
Thanks a lot, amazing fixes!
It's odd that these survived so long, maybe this indicates the GitPython is mostly used in one-off commands rather than long running processes.

marlamb:
You are welcome, thanks for merging that quickly!

Codym48:
Thank you for committing this @marlamb and merging it @Byron!

(We only tripped over one of these after we enabled `filterwarnings("error")` and executed the pytest suite locally within a [linked worktree](https://git-scm.com/docs/git-worktree#_description). It didn't appear when running within the main worktree or within any CI pipeline.)

Any idea when the next release is scheduled? Can I volunteer to help?

Byron:
@Codym48 Thanks for your help! It would be great if https://github.com/gitpython-developers/GitPython/pull/1535 cold be merged before a release is created, it's just missing a test. You could probably submit a PR on top of that one to add it. Thank you.

Codym48:
How about something like this? https://github.com/kfreezen/GitPython/pull/1

Issue #1534, opened 2023-01-10
GitConfigParser().get_values() does not load sections properly.


kf-novi:
To replicate this bug, first do the following in your CLI

```
git config --global --add safe.directory /some/directory/here
```
Then run this python script
```
from git import GitConfigParser
from git.config import get_config_path

config = GitConfigParser(get_config_path("global"), read_only=False)
config.get_values("safe", "directory") # This throws a KeyError
```
There is a workaround:
```
config.get_value("safe", "directory")
config.get_values("safe", "directory") # This works
```

I will be taking a look at the code to see if I can submit a PR for this.

Byron:
Closing as it's fixed by #1555 .

larstobi:
I'm still getting the` KeyError: 'safe'` in version 3.1.43 (using Python 3.12.4), but only if the safe section is not already present in the git config. The above workaround with calling `.get_value()` first still works, though.
@Byron 
#1555

Byron:
Maybe you can contribute a test that reproduces the issue. From there it should be possible to fix the issue for good. Thanks a lot.

PR #1535, opened 2023-01-10
Proposed change to get_values behavior

kfreezen:
get_values() did not load existing sections before checking the `_sections` property.

**NOTE:** Haven't written a test to check for this yet.

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -796,6 +796,7 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
         :raise TypeError: in case the value could not be understood
             Otherwise the exceptions known to the ConfigParser will be raised."""
         try:
+            self.sections()
             lst = self._sections[section].getall(option)
         except Exception:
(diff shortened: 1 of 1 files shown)


kfreezen:
Proposed fix for #1534

kfreezen:
Looking at test here; I suppose we can leave it raising a KeyError. The main fix was force-loading the `_sections` attribute before checking the list.

I'll re-commit using KeyError.

Byron (changes requested):

I think the commit should link back to the issue it resolve to create a more declarative history.
Finally it should be possible to devise a test that would fail without this change to cement it. Otherwise it might be removed in future because it's unclear why eager loading of sections is necessary.

Codym48:
@kfreezen, consider accepting https://github.com/kfreezen/GitPython/pull/1 to add test coverage for this change? I verified it fails on main:

image

Byron:
Closing as it's superseded by #1555

PR #1537, opened 2023-01-12
fix files list on file rename

teknoraver:
GitPython parses the output of `git diff --no-renames` to get the files changed in a commit.
This breaks when a commit contains a file rename, because the output of `git diff` is different than expected.
This is the output of a normal commit:
```
$ git diff --numstat 6ab9810cfe6c^ 6ab9810cfe6c
5       2       drivers/clk/zynqmp/divider.c
```
And this a commit containing a rename:
```
$ git diff --numstat db401875f438^ db401875f438
1       1       tools/testing/selftests/drivers/net/mlxsw/{spectrum-2 => }/devlink_trap_tunnel_ipip6.sh
```
This can be triggered by this code:
```python
for commit in repo.iter_commits():
    print(commit.hexsha)
        for file in commit.stats.files:
            print(file)
```
Which will print for the normal commit:
```
6ab9810cfe6c8f3d8b8750c827d7870abd3751b9
'drivers/clk/zynqmp/divider.c'
```
And when there is a rename:
```
db401875f438168c5804b295b93a28c7730bb57a
('tools/testing/selftests/drivers/net/mlxsw/{spectrum-2 => '
'}/devlink_trap_tunnel_ipip6.sh')
```
Fix this by pasing the --no-renames option to `git diff` which ignores renames and print the same output as if the file was deleted from the old path and created in the new one:
```
$ git diff --numstat --no-renames db401875f438^ db401875f438
250     0       tools/testing/selftests/drivers/net/mlxsw/devlink_trap_tunnel_ipip6.sh
0       250     tools/testing/selftests/drivers/net/mlxsw/spectrum-2/devlink_trap_tunnel_ipip6.sh
````

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -324,14 +324,14 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
 
         :return: git.Stats"""
         if not self.parents:
-            text = self.repo.git.diff_tree(self.hexsha, "--", numstat=True, root=True)
+            text = self.repo.git.diff_tree(self.hexsha, "--", numstat=True, no_renames=True, root=True)
             text2 = ""
             for line in text.splitlines()[1:]:
                 (insertions, deletions, filename) = line.split("\t")
                 text2 += "%s\t%s\t%s\n" % (insertions, deletions, filename)
             text = text2
         else:
-            text = self.repo.git.diff(self.parents[0].hexsha, self.hexsha, "--", numstat=True)
+            text = self.repo.git.diff(self.parents[0].hexsha, self.hexsha, "--", numstat=True, no_renames=True)
         return Stats._list_from_string(self.repo, text)
 
     @property
--- a/test/test_commit.py
+++ b/test/test_commit.py
@@ -159,6 +159,37 @@ class TestCommit(TestCommitSerialization):
         self.assertEqual(commit.committer_tz_offset, 14400, commit.committer_tz_offset)
         self.assertEqual(commit.message, "initial project\n")
 
+    def test_renames(self):
+        commit = self.rorepo.commit("185d847ec7647fd2642a82d9205fb3d07ea71715")
+        files = commit.stats.files
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by Byron — CHANGES_REQUESTED** [review]
Thanks a lot!

It should be straightforward to add a test that motivates this changes. The test-suite uses fixture files which could be derived from an example repository.

With a test that fails without this change, the fix can be merged.

Thank you

**Comment by teknoraver** [comment]
@Byron done, and I rewrote the commit message using GitPython hashes, so is simpler for anyone to test

**Review by Byron — CHANGES_REQUESTED** [review]
Thanks a lot, this looks like the way to go!

Instead of checking what it shouldn't be (which is a huge set of possible values), could you instead check what it should be? How does a rename look like with the current interface, I'd be very interested to see that.

For posterity, here is what `git` thinks about this:

```patch
commit 185d847ec7647fd2642a82d9205fb3d07ea71715
Author: Dominic <<redacted>@gmail.com>
Date:   Mon Jul 12 17:02:21 2021 +0100

    Update and rename test_pytest.yml to Future.yml

diff --git a/.github/workflows/test_pytest.yml b/.github/workflows/Future.yml
similarity index 92%
rename from .github/workflows/test_pytest.yml
rename to .github/workflows/Future.yml
index 627e720f..39146533 100644
--- a/.github/workflows/test_pytest.yml
+++ b/.github/workflows/Future.yml
@@ -5,7 +5,9 @@ name: Future
 
 on:
   push:
-    branches: [main]
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
 
 jobs:
   build:
```

teknoraver:
This is the output of `git show`, gitpython does `git diff --numstat $commit^ $commit` which gives:
```
3       1       .github/workflows/{test_pytest.yml => Future.yml}
```
And with `--no-renames` added:
```
57      0       .github/workflows/Future.yml
0       55      .github/workflows/test_pytest.yml
```

Byron (approved):
Thanks a million, that is what I had in mind.

Thanks again for contributing!

PR #1541, opened 2023-01-15
Declare support for Python 3.11

hugovk:
Add the Trove classifier to declare support for Python 3.11 on PyPI.

Also a few other little changes to `setup.py`.

### DIFF
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -1,16 +1,12 @@
 name: test-cygwin
 
-on:
-  push:
-    branches:
-      main
-  pull_request:
-    branches:
-      main
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
     runs-on: windows-latest
+    strategy:
+      fail-fast: false
     env:
       CHERE_INVOKING: 1
       SHELLOPTS: igncr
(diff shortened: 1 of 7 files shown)


Byron (changes requested):
It looks like f-strings aren't supported consistently, but if the scope could be reduced to just he trove specifier this PR could probably be merged - CI should be green then.

hugovk:
Hi! I dropped the f-strings commit.

The Flake8 Bugbear warnings also come from files I didn't touch, because the CI is using unpinned versions and B028 is a new check released yesterday:

https://github.com/PyCQA/flake8-bugbear/releases/tag/23.1.14

(I use pre-commit for pinning, running and updating linters.)

Byron:
Oh, I see. Not pinning the version here in the GitPython CI has led to unpleasant surprises more than once. Would you be able to adjust the workflow file to install a specific (the previous) version of it and fix CI that way? Maybe the code for that is already present in the hooks you are using.

Byron (approved):
Thanks so much, this is great, much appreciated work!

I will probably not be the one to ever update the flake8 version number, and if this is something you are interested in doing occasionally, I'd welcome that as well.

Byron:
By the way, I saw on your profile that this quality of work doesn't come from nothing. Is Co-Maintaining GitPython an option for you? I really am not doing a good job as I have not caught up with Python which I stopped using entirely what feels like 10 years ago. Thanks for your consideration.

hugovk:
Hi, thanks for the offer, however my open-source plate is a little full at the moment.

PR #1545, opened 2023-01-21
Fix ignored

Lightborne:
* Add test to sanity-check basic behavior of `ignored()` method
* Add test to verify GitCommandError is raised if `git check-ignore` returns any error code other than 1
* Modify `ignred()` method to raise GitCommandError if `git check-ignore` returns any error code other than 1

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -873,8 +873,15 @@ class Repo(object):
         """
         try:
             proc: str = self.git.check_ignore(*paths)
-        except GitCommandError:
-            return []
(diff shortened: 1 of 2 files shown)


Byron:
Great work, thanks so much! I love to finally have a test for that.

For a moment I thought this might be a breaking change, just to realize that the previous implementation was just a little too 'unconditional' and this PR implements the proper fix by raising an error when the error code does *not* indicate a lack of matches.

Lightborne:
Thanks so much!

Issue #1549, opened 2023-01-27
working_dir has Optional type when it can't be None


Sineaggi:
In https://github.com/gitpython-developers/GitPython/blob/main/git/repo/base.py#L285 `working_dir` is never assigned to `None`, which means it type shouldn't be marked as `Optional`.

Byron:
What happens when the repository is bare? Then by definition it has no working dir, and I don't think `working_dir` is populated anyway.

PR #1550, opened 2023-01-27
Remove optional from two member variables

Sineaggi:
Neither `working_dir` nor `git_dir` appear to leave `__init__()` set to `None`. Let's remove the `None` initialization, and simply use them as type markers.

Fixes https://github.com/gitpython-developers/GitPython/issues/1549

### DIFF
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -98,6 +98,7 @@ class TestBase(_TestBase):
     def test_with_bare_rw_repo(self, bare_rw_repo):
         assert bare_rw_repo.config_reader("repository").getboolean("core", "bare")
         assert osp.isfile(osp.join(bare_rw_repo.git_dir, "HEAD"))
+        assert osp.isdir(bare_rw_repo.working_dir)
 
     @with_rw_repo("0.1.6")
     def test_with_rw_repo(self, rw_repo):
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -9,6 +9,7 @@ import sys
 import tempfile
 from unittest import SkipTest, skipIf
 
+from git import Repo
 from git.objects import Blob, Tree, Commit, TagObject
 from git.compat import is_win
 from git.objects.util import get_object_type_by_name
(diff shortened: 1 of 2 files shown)


Sineaggi:
The changes to `_get_config_path` and `_config_reader` were necessary as these functions are called during `__init__`, sometimes before `git_path` is initialized.

Byron (changes requested):
I'd prefer a test-driven approach. If `working_dir` cannot be none, I'd like to see what happens with bare repositories. What kind of `working_dir` are they expected to have?

Sineaggi:
A bare repo should still have working_dir afaik

test/test_base.py
```diff
@@ -98,6 +98,7 @@ def test_object_resolution(self):
     def test_with_bare_rw_repo(self, bare_rw_repo):
         assert bare_rw_repo.config_reader("repository").getboolean("core", "bare")
         assert osp.isfile(osp.join(bare_rw_repo.git_dir, "HEAD"))
+        assert osp.isdir(bare_rw_repo.working_dir)
```
Byron, line 101: Thanks! 

Could you also add an assertion of the `working_tree_dir` for completeness? It's easy to confuse both which is what I did apparently. The `working_dir` is the CWD of the git repository, either the `working_tree_dir` or the `git_dir`, which I think should be an assertion for the bare and non-bare cases.

If that assertion holds, of course, I was going by the docs.

Screenshot 2023-02-02 at 07 15 07

Byron (commented):
Thanks for following up.

Can you try to add clearer assertions to the different repo types, bare and non-bare, that make the relationship of these different `working*` directories (even more) clear?

test/test_base.py
Sineaggi: Ok so the docs in `base.py` say that working_tree_dir can return none
```python
    @property
    def working_tree_dir(self) -> Optional[PathLike]:
        """:return: The working tree directory of our git repository. If this is a bare repository, None is returned."""
        return self._working_tree_dir
```

I've added a few more assertions, please check if I understood you.

Sineaggi: Also updated the python docs of the class to mention returning None

Byron:
Thanks a lot for bearing with me, I was confused by the difference between `working_dir` and `working_tree_dir`, especially since there is a `work_dir` in `gitoxide` (the equivalent of `working_tree_dir`).

And thanks for this great contribution :)!

PR #1551, opened 2023-01-31
Enable user to override default diff -M arg

mellowed100:
Small fix to allow overriding of default diff **-M** arg. Using --find-renames=<n> or -M<n> or --no-renames will disable **-M**.

Full discussion is here:
https://github.com/gitpython-developers/GitPython/discussions/1195

### DIFF
--- a/test/test_diff.py
+++ b/test/test_diff.py
@@ -420,17 +420,17 @@ class TestDiff(TestBase):
         repo = Repo.init(rw_dir)
         file_a = osp.join(rw_dir, "file_a.txt")
         with open(file_a, "w", encoding='utf-8') as outfile:
-            outfile.write("hello world")
+            outfile.write("hello world\n")
         repo.git.add(Git.polish_url(file_a))
         repo.git.commit(message="Added file_a.txt")
 
         # remove file_a.txt
         repo.git.rm(Git.polish_url(file_a))
 
-        # create and commit file_b.txt
+        # create and commit file_b.txt with similarity index of 52
         file_b = osp.join(rw_dir, "file_b.txt")
         with open(file_b, "w", encoding='utf-8') as outfile:
-            outfile.write("hello world")
+            outfile.write("hello world\nhello world")
         repo.git.add(Git.polish_url(file_b))
         repo.git.commit(message="Removed file_a.txt. Added file_b.txt")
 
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
That looks simple enough, thanks a lot!

Is it possible to have a test leverage this newly gained capability?
Now it's possible to do that, previously it would generate an invalid command-line, so a test can probably check for a successful diff call.

mellowed100:
Let me know if this works for you. thanks!

test/test_diff.py
```diff
@@ -411,3 +411,50 @@ def test_diff_interface(self):
...
+        diffs = commit_b.diff(commit_a, no_renames=True)
+        self.assertEqual(2, len(diffs))
+
+        # check fileA.txt deleted
+        diff = diffs[0]
+        self.assertEqual(True, diff.deleted_file)
+        self.assertEqual('file_a.txt', diff.a_path)
+
+        # check fileB.txt added
+        diff = diffs[1]
```
Byron, line 458: Could you also show how the `M` argument can now be used? I am interested in seeing it myself especially since most of the work seemed to hinge around supporting it.

Thanks a lot!

Byron (changes requested):
This looks great already!

I left a comment with a tiny last request to complete the `M` journey with a show case on how callers can now use it themselves :).

test/test_diff.py
mellowed100: Yup, no problem... I'll need a few days as I have to work on some other things right now. thanks for taking a look!

mellowed100:
Hey Byron!
So I updated the test to address those changes you asked for. The initial commits of **file_a.txt** and **file_b.txt** have been modified to have a similarity index of 52%. So the default **diff -M** (50%) should still flag it as a renamed file. I added **diff -M75%** and **diff -M40%** to check for both added/deleted files and renamed files. Let me know if this works for you. Thanks!

Byron:
Thanks a lot - doing it like this is fantastic and I couldn't imagine a way to better show the effect of the `M` flag.

I will merge once CI runs through. Silly me aborted the existing Cygwin run because it seemed stuck, but history shows it's taking between 13 and 19 minutes to run there - incredible especially since the linux hosts are finished in less than 3 minutes.

mellowed100:
You're welcome! Thanks for your guidance and patience (This was my first PR). I really appreciate it! -Take care!

PR #1554, opened 2023-02-12
Fix RecursionError when iterating streams

eric-wieser:
`next(self)` was previously just a recursive call that does nothing.

This now implements `__next__` using the actual implementation, and lets the python2-compatibility `next()` just wrap it.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -701,8 +701,7 @@ class Git(LazyMixin):
 
             return line
 
-        def next(self) -> bytes:
-            return next(self)
+        next = __next__
 
         def __del__(self) -> None:
             bytes_left = self._size - self._nbr
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -695,14 +695,14 @@ class Git(LazyMixin):
             return self
 
         def __next__(self) -> bytes:
-            return next(self)
-
-        def next(self) -> bytes:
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, I have a feeling that this issue was present for a long time already!

Do you think Python 2 compatibility is still required? Support for it was dropped a while ago and if you agree it can be dropped entirely.

eric-wieser:
I have no objection to dropping the `next` method entirely; I left it only for the sake of being conservative.

git/cmd.py
```diff
@@ -695,15 +695,14 @@ def __iter__(self) -> "Git.CatFileContentStream":
...
-
-        def next(self) -> bytes:
             line = self.readline()
             if not line:
                 raise StopIteration
 
             return line
 
+        next = __next__
+
```
eric-wieser, line 705: Feel free to commit through the github UI if you want to drop this Python2 compatibility layer entirely:
```suggestion
```

Note that there may be some downstream code that relies on using `next` because `__next__` was broken though!

Byron:
Great! Let's do that and the PR can be merged. Thank you

git/cmd.py
Byron: Ah, right. Oh well, let's be conservative then.

PR #1555, opened 2023-02-16
Fix get_values() so it correctly loads section names

Codym48:
As described in https://github.com/gitpython-developers/GitPython/issues/1534, some calls to GitConfigParser().get_values() fail in main with a `KeyError` about a missing section name even though the named sections do exist within the config file.

Add a test that calls get_values() before anything else to catch this failure. Update the get_values() function to work, even if it's the first call after creating the GitConfigParser, by eagerly load the section names before trying to access them. This passes the new test.


Byron:
Thanks a lot for your help!

Issue #1533, opened 2023-01-10
commit trailers should be a list instead of a dictionary


nickbroon:
Consider for example this linux kernel commit: https://gitlab.com/linux-kernel/stable/-/commit/1382999aa0548a171a272ca817f6c38e797c458c which has following trailers that include multiple `cc` and `signed-off-by` trailers:

```
Link: https://lore.kernel.org/lkml/7cbe96cf-e0b5-ba63-d1b4-f63d2e826efa@suse.cz/


Cc: stable@vger.kernel.org # 6.1+
Reported-by: default avatarVlastimil Babka <vbabka@suse.cz>
Suggested-by: default avatarLinus Torvalds <torvalds@linux-foundation.org>
Acked-by: default avatarLuigi Semenzato <semenzato@chromium.org>
Cc: Peter Huewe <peterhuewe@gmx.de>
Cc: Jarkko Sakkinen <jarkko@kernel.org>
Cc: James Bottomley <James.Bottomley@HansenPartnership.com>
Cc: Johannes Altmanninger <aclopte@gmail.com>
Signed-off-by: Jason Donenfeld's avatarJason A. Donenfeld <Jason@zx2c4.com>
Signed-off-by: default avatarLinus Torvalds <torvalds@linux-foundation.org>
```

And `git interpret-trailers --parse ...` will return all of these.

However the `trailers()` function returns a dictionary with only the last value for each token.

itsluketwist:
@Byron - happy for me to fix this up? 😄

Byron:
Of course, that would be much appreciated. Thanks for offering.

PR #1576, opened 2023-04-21
Fix up the commit trailers functionality

itsluketwist:
Add trailers_dict and trailers_list methods to fix the commit trailers functionality.

I left the existing method as is to not break any existing code using it, leaving a comment about how it's incorrect.

Also updated the trailers tests.

Work here resolves issue https://github.com/gitpython-developers/GitPython/issues/1533

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -336,24 +336,6 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         return Stats._list_from_string(self.repo, text)
 
     @property
-    def trailers(self) -> Dict[str, str]:
-        """Get the trailers of the message as a dictionary
-
-        Git messages can contain trailer information that are similar to RFC 822
-        e-mail headers (see: https://git-scm.com/docs/git-interpret-trailers).
-
-        WARNING: This function only returns the latest instance of each trailer key
-        and will be deprecated soon. Please see either ``Commit.trailers_list`` or ``Commit.trailers_dict``.
-
-        :return:
-            Dictionary containing whitespace stripped trailer information.
-            Only the latest instance of each trailer key.
-        """
-        return {
-            k: v[0] for k, v in self.trailers_dict.items()
-        }
-
-    @property
     def trailers_list(self) -> List[str]:
         """Get the trailers of the message as a list
 
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -336,6 +336,20 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         return Stats._list_from_string(self.repo, text)
 
     @property
+    def trailers(self) -> Dict[str, str]:
+        """Get the trailers of the message as a dictionary
+
+        :note: This property is deprecated, please use either ``Commit.trailers_list`` or ``Commit.trailers_dict``.
+
+        :return:
+            Dictionary containing whitespace stripped trailer information.
+            Only contains the latest instance of each trailer key.
+        """
+        return {
+            k: v[0] for k, v in self.trailers_dict.items()
+        }
+
+    @property
     def trailers_list(self) -> List[Tuple[str, str]]:
         """Get the trailers of the message as a list
 
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -336,7 +336,7 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         return Stats._list_from_string(self.repo, text)
 
     @property
-    def trailers_list(self) -> List[str]:
+    def trailers_list(self) -> List[Tuple[str, str]]:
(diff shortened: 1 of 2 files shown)


git/objects/commit.py
```diff
@@ -335,8 +336,70 @@ def stats(self) -> Stats:
...
     @property
-    def trailers(self) -> Dict:
-        """Get the trailers of the message as dictionary
+    def trailers(self) -> Dict[str, str]:
+        """Get the trailers of the message as a dictionary
+
+        Git messages can contain trailer information that are similar to RFC 822
+        e-mail headers (see: https://git-scm.com/docs/git-interpret-trailers).
+
+        WARNING: This function only returns the latest instance of each trailer key
```
Byron, line 345: I think it's OK to deprecate it, and suggest using `trailers_dict()`. Then this warning can be removed as it will show up when using it.

git/objects/commit.py
```diff
@@ -335,8 +336,70 @@ def stats(self) -> Stats:
...
+        :return:
+            Dictionary containing whitespace stripped trailer information.
+            Only the latest instance of each trailer key.
+        """
+        return {
+            k: v[0] for k, v in self.trailers_dict.items()
+        }
+
+    @property
+    def trailers_list(self) -> List[str]:
```
Byron, line 357: To me it seems this method would be more useful if the keys and values would be returned as a tuple, i.e. `List[(str, str)]`. Is there a good reason not to do that?

PS: *It's probably OK to always trim whitespace as it should never be significant.*

git/objects/commit.py
```diff
@@ -335,8 +336,70 @@ def stats(self) -> Stats:
...
+                "key2 :    value 2 with inner spaces",
+            ]
+
+
+        :return:
+            List containing whitespace stripped trailer information.
+        """
+        cmd = ["git", "interpret-trailers", "--parse"]
+        proc: Git.AutoInterrupt = self.repo.git.execute(cmd, as_process=True, istream=PIPE)  # type: ignore
+        trailer: str = proc.communicate(str(self.message).encode())[0].decode()
```
Byron, line 393: `.decode()` will probably use the interpreter encoding which I hope is UTF8 by now for the greatest changes of success. If not, I think the expected encoding should be specified. It's worth noting that `git` doesn't enforce any encoding, even though commits may have encoding hints embedded as header field. It's a bit of a mess to deal with this correctly here, so a best-guess is all we can reasonably do here. (This is not `gitoxide`, unfortunately).

Byron (changes requested):
Thanks a lot for your help, it looks very good already!

There isn't much missing at all for getting it merged, and I am looking forward to hearing what you think. Thank you.

git/objects/commit.py
itsluketwist: Agreed - deprecated [78424b5](https://github.com/gitpython-developers/GitPython/pull/1576/commits/78424b56654ad476da4bd2faf88d3875c5265e0d)

git/objects/commit.py
itsluketwist: Agreed - done [abde3ea](https://github.com/gitpython-developers/GitPython/pull/1576/commits/abde3eafd293e8fa2ef2dc22d58ba5d80f1702e9)

git/objects/commit.py
itsluketwist: Wasn't sure exactly what you were asking for here 😅 but I've specified a decoding - [ed36bd9](https://github.com/gitpython-developers/GitPython/pull/1576/commits/ed36bd903e1fdf45c53b52dbcb1b2d8444965d98)

git/objects/commit.py
Byron: That's not deprecation 😅, that's downright removal, which is a breaking change.

After taking a look I realize now that there is no built-in facility for marking properties or functions as deprecated. So I think documenting the deprecation is as good as it gets.

Sorry for the hassle, but once that is reverted, it's ready to merge.

Byron:
Thanks a lot for bearing with me on this one, and I really appreciate the one commit per change with reference to it in the respective comment.

There is just [one tiny adjustment](https://github.com/gitpython-developers/GitPython/pull/1576#discussion_r1174524116) left and then it's ready to go.

git/objects/commit.py
itsluketwist: No worries! `trailers` method is back - [9ef07a7](https://github.com/gitpython-developers/GitPython/pull/1576/commits/9ef07a731caf39684891bce5cc92c4e91829138d)

Byron:
Thanks a lot for the contribution, much appreciated!

itsluketwist:
> Thanks a lot for the contribution, much appreciated!

No problem!

Issue https://github.com/gitpython-developers/GitPython/issues/1533 can also be closed now. 😄

PR #1578, opened 2023-04-26
get_user_id: use EUID instead of getpass.getuser() to resolve users

### LINKED ISSUE #1577: git_user_id should returnt he EUID
i think that `git.utils.get_user_id` should return the Effective user and not the.  currently the function uses `getpass.getuser()` which relies on environment variables and not getuid/geteuid functions, from the docs

> checks the environment variables LOGNAME, USER, LNAME and USERNAME, in order, and returns the value of the first one which is set to a non-empty string. If none are set, the login name from the password database is returned on systems which support the [pwd](https://docs.python.org/3/library/pwd.html#module-pwd) module, otherwise, an exception is raised.
with the following code 

```python
import git
import os
import pwd
import shutil
from git.util import get_user_id
from pathlib import Path

CHECKOUT = Path("/tmp/gist")
shutil.rmtree(CHECKOUT, ignore_errors=True)
ORIGIN_REPO = "https://gist.github.com/b4ldr/449d490b43fb3d2d01a74d28bc9a41e4"
os.seteuid(pwd.getpwnam('jbond').pw_uid)
euid = pwd.getpwuid(os.geteuid()).pw_name
print(f"git.utils.get_user_id: {get_user_id()}")
print(f"euid: {euid}")
repo = git.Repo.clone_from(ORIGIN_REPO, CHECKOUT)
print(f"directory owner: {CHECKOUT.owner()}")
```

When run as root we get the following output 
```
$ sudo python -i git_test.py                                                                  
git user id: root@storage
euid: jbond
directory owner: jbond
```

i think it would be more useful if gitpython returned the user matching the EUID e.g. `pwd.getpwuid(os.geteuid()).pw_name)`

b4ldr:
fixes #1577

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -9,10 +9,10 @@ import os.path as osp
 from .compat import is_win
 import contextlib
 from functools import wraps
-import getpass
 import logging
 import os
 import platform
+import pwd
 import subprocess
 import re
 import shutil
@@ -410,7 +410,8 @@ def is_cygwin_git(git_executable: Union[None, PathLike]) -> bool:
 
 def get_user_id() -> str:
     """:return: string identifying the currently active system user as name@node"""
-    return "%s@%s" % (getpass.getuser(), platform.node())
+    user = pwd.getpwuid(os.geteuid()).pw_name
+    return "%s@%s" % (user, platform.node())
 
 
 def finalize_process(proc: Union[subprocess.Popen, "Git.AutoInterrupt"], **kwargs: Any) -> None:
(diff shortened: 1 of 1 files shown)


Byron:
No argument is made for why this should be preferred over the current implementation. 

However, while checking what `git` does I noticed that [it does also use `getuid()`](https://github.com/git/git/blob/9857273be005833c71e2d16ba48e193113e12276/ident.c#L46).

Thus there is nothing to fix, and changing it would be a breaking change as well which is another reason not to pursue this further. Thanks for your understanding.

PR #1579, opened 2023-04-27
remove duplicate imports

adriengardou:
subprocess module is already fully imported


Byron:
Thanks for contributing, but in this case the symbol imports are intentional.
There are [many issues that seek attention though](https://github.com/gitpython-developers/GitPython/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22).

Issue #1587, opened 2023-05-13
Reuse of “exc” name in __init__.py confuses Pylint


Hawk777:
It’s not entirely clear how one is supposed to name GitPython’s exceptions:
* According to [the API reference](https://gitpython.readthedocs.io/en/stable/reference.html#module-git.exc), they live in the `git.exc` module.
* In `__init__.py`, the very first line does `from git.exp import *`, meaning they can actually just be referenced in the `git` module instead if one wishes (is that intentional/public knowledge?).

Then, further down `__init__.py`, there are two places where `except Something as exc` is done in module-level code. While those won’t execute on success (and will prevent the module from importing on failure), they do confuse Pylint: if I try to `except git.exc.InvalidGitRepositoryError` in my own code and then lint it, Pylint complains that `Instance of 'GitError' has no 'InvalidGitRepositoryError' member`, because it believes the name `git.exc` itself might actually be an instance of `GitError` (as it would be if [this `except` clause](https://github.com/gitpython-developers/GitPython/blob/main/git/__init__.py#L59) were taken).

Is it intended that we refer to exception classes from `git` rather than `git.exc`? If so, could the documentation be updated? If not—if we are not supposed to know about the `from git.exc import *`—could the variable names used in those `except` blocks be modified to not collide with a submodule name?

Byron:
No matter what the intention was, it appears the documentation could be updated to mention these can be imported using multiple paths.

Further, it appears `except E as exc` can be fixed as well, and a contribution for all of that is definitely welcome.

Hawk777:
> No matter what the intention was, it appears the documentation could be updated to mention these can be imported using multiple paths.

I was thinking, if it wasn’t intentional, it would be better not to document it since if it’s documented, people might rely on it, but if it’s not documented, it’s OK to break it in future versions. So I figured it was up to you.

> Further, it appears `except E as exc` can be fixed as well, and a contribution for all of that is definitely welcome.

PEP 8 appears to suggest using a leading underscore for module-private global variables; is that your preference here? If so I can throw together an MR that does that.

Byron:
I think the only fix possible is to turn `… as exc` into `… as _exc`, and a PR is very welcome. Thanks for your help.

PR #1590, opened 2023-05-15
Name top-level exceptions as private variables

Hawk777:
Fixes GH-1587


Byron:
Thanks a lot for your help!

PR #1603, opened 2023-07-05
fix pypi long description

eUgEntOptIc44:
Hi,
I recently found this great project on PyPi and noticed the short description.

I understand that thats caused by the following. The `setup.py` file reads the `README.md` but then the variable `long_description` remains unused. Instead the string `"""GitPython is a ... repositories"""` is used as the value of the `long_description` attribute.

I also applied a few changes to the `README.md` to increase the readability of the code blocks.

Regards,  
Jean-Luc

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,17 +51,19 @@ The installer takes care of installing them for you.
 
 If you have downloaded the source code:
 
-    python setup.py install
+```bash
+python setup.py install
+```
 
 or if you want to obtain a copy from the Pypi repository:
 
-    pip install GitPython
+```bash
+pip install GitPython
+```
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot for the fixes, to the readme in particular!

The reason for not using the actual `long_description` was that it wouldn't format markdown at all, even though everything seems to be configured.

Let's give it another try :).

Byron:
So much better, thanks again! [GitPython · PyPI.pdf](https://github.com/gitpython-developers/GitPython/files/12011572/GitPython.PyPI.pdf)

PR #1606, opened 2023-07-07
Don't rely on __del__

r-darwish:


git/index/util.py
```diff
@@ -47,12 +48,21 @@ def __init__(self, file_path: PathLike) -> None:
...
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> bool:
         if osp.isfile(self.tmp_file_path):
             if is_win and osp.exists(self.file_path):
                 os.remove(self.file_path)
             os.rename(self.tmp_file_path, self.file_path)
-        # END temp file exists
+
+        return False
```
Byron, line 65: Is there any reason exceptions shouldn't propagate?

git/index/base.py
```diff
@@ -360,20 +361,19 @@ def from_tree(cls, repo: "Repo", *treeish: Treeish, **kwargs: Any) -> "IndexFile
...
         # The TemporaryFileSwap assure the original file get put back
-        if repo.git_dir:
-            index_handler = TemporaryFileSwap(join_path_native(repo.git_dir, "index"))
         try:
-            repo.git.read_tree(*arg_list, **kwargs)
-            index = cls(repo, tmp_index)
-            index.entries  # force it to read the file as we will delete the temp-file
-            del index_handler  # release as soon as possible
+            with ExitStack() as stack:
+                if repo.git_dir:
```
Byron, line 366: I wonder why that was necessary - wouldn't `git_dir` always be set? Otherwise it wouldn't be usable, right?
If that check was unnecessary, I assume the `ExitStack` could be left out as well.

Byron (commented):
Thanks a lot, that's very valuable! Believe it or not, but when GitPython was written, `del` was still deterministic due to lack of GC. Good times.

Once a few questions are clarified, nothing should be in the way of merging this.

git/index/util.py
r-darwish: I think that `False` means it should be propageted

git/index/base.py
r-darwish: I made some refactoring here to remove the finally block

git/index/util.py
Byron: Ah, true, I must have flipped bit there. Thanks for bearing with me on this one.

Byron:
It looks like the `ExitStack` and context managers in general are *the* way around trying to rely in `__del__`. Thanks for all the improvements.

PR #1609, opened 2023-07-10
Block insecure non-multi options in clone/clone_from

Beuc:
I'm part of the Debian LTS (Long Term Support) Team and I'm working on integrating the fix for CVE-2022-24439 for GitPython in Debian.

After contacting @Byron privately we determined that #1521 would need a follow-up fix, which I hereby propose in this PR.

I also modified 2 tests to validate the change, but I'm not sure if we want to double-check (multi/non-multi) all the currenttt clone/clone_from tests, so I didn't, but I can do so if necessary.

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -1203,6 +1203,8 @@ class Repo(object):
 
         if not allow_unsafe_protocols:
             Git.check_unsafe_protocols(str(url))
+        if not allow_unsafe_options:
+            Git.check_unsafe_options(options=list(kwargs.keys()), unsafe_options=cls.unsafe_git_clone_options)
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a million!

I think it's sufficient to 'sample' only a couple of test as we know the fix is present for all clones, and I don't think that will change.

I will see to create a new release shortly.

Beuc:
Great!

I reached out to Skyk to register another CVE for this additional fix.

Beuc:
Snyk didn't answer, so I eventually reached out to MITRE who assigned CVE-2023-40267 to this issue.

PR #1617, opened 2023-07-20
Partial clean up wrt mypy and black

bodograumann:
Some of the typing errors require deeper knowledge, so I was not able to fix them in reasonable time.

- Do not typecheck submodule
- Define supported version for mypy
- Ignore remaining [unreachable] type errors
- Allow explicit casting even when slightly redundant
- Run black and exclude submodule
- Add missing type annotation
- Apply straight-forward typing fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ To typecheck, run: `mypy -p git`
 
 To test, run: `pytest`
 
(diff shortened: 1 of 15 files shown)


Byron:

PR #1618, opened 2023-07-21
Disable merge_includes in config writers

### LINKED ISSUE #1614: set_tracking_branch fails quietly when .git/config contains an include
I have git-branchless setup in this repository and for the longest time I was confounded, why `set_tracking_branch` did not work (anymore).
Turns out, git-branchless adds an `[include]` section to the `.git/config` file and this causes GitPython to quietly skip writing the config changes back that are necessary for `set_tracking_branch` to have any effect.

Could we maybe set `merge_includes=False` for the config_writer; at least in the Head class?

bodograumann:
Fixes #1614

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -600,7 +600,7 @@ class Repo(object):
             system = system wide configuration file
             global = user level configuration file
             repository = configuration file for this repository only"""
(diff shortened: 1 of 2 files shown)


Byron:
Looks good to me, thanks a lot for the fix!

This could be one of those fixes that breaks production code, but maybe we are lucky and no one is relying on this bug.

Issue #1039, opened 2020-07-20
Tutorial: first example code is very unclear and confusing


pbb72:
The first example code in the tutorial is very unclear:

```python
from git import Repo

# rorepo is a Repo instance pointing to the git-python repository.
# For all you know, the first argument to Repo is a path to the repository
# you want to work with
repo = Repo(self.rorepo.working_tree_dir)
assert not repo.bare
```

To start with, this just gives me the error message

`NameError: name 'self' is not defined`

The comment "rorepo is a Repo instance pointing to the git-python repository" doesn't help. Does this mean "rorepo" is just an example word, and I have to replace it with something from my own system? Does this reference a directory or a repository name I should create on my filesystem? The term "rorepo" doesn't come up any other places in the whole documentation.

Also the following text is only adding to the confusion. "In the above example, the directory self.rorepo.working_tree_dir equals /Users/mtrier/Development/git-python ..." NOT IT DOESN'T. Maybe it does on the specific system for the documentation writer, but it is not a general rule.

Please modify the tutorial code to apply to general users. ft code requires previous actions/setup, then please specify this.

Handballfreak:
Same issues here. In generall the documentation is more confussing than it helps.

Byron:
There can only be agreement on this, as the tutorial section prefers correctness and testability over being easy to follow. Most certainly the intend of it should be made clearer to set expectations.

Since you have been bitten by this and certainly an opinion, I would be happy to see a PR with adjustment that you would have wanted the first time around. It probably won't take much to at least set expectations, without needing to modify the way it's done as this would be way more work that I don't expect anyone to do really.

Thanks for your consideration.

pbb72:
Hi Byron, thank you for your reaction. However, I do not think you understand my situation as a reader.

I read the tutorial, and get confused by it. Obviously, this means I do not understand much about GitPython. That is the reason I am reading the tutorial. And yet you suggest I send you a PR with how the tutorial should have been written.

How do you think I can do this? I am not capable of writing a tutorial for GitPython! If I was, I wouldn't need one... All I can do, is report that it is not understandable for someone who needs a tutorial.

Sometimes expecting people who report problems, also to give you the solution, goes a bit too far!

Byron:
[...]

PR #1628, opened 2023-08-04
Clarify initial usage

tacerus:
Instead of referencing test code which cannot be used on its own, present a short example on how to initialize a repository instance from the current working directory.

Solves #1039.

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -10,20 +10,20 @@ GitPython Tutorial
 
 GitPython provides object model access to your git repository. This tutorial is composed of multiple sections, most of which explain a real-life use case.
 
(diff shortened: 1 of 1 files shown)


Byron (commented):
I do appreciate the attempt to the docs but believe it's preferable to stay consistent. If improvements should be made, I think this should be done in the form of adjusting variable names in the underlying test, or by improving the comments.

You may find interesting that with the next release there will be a new quick-start section that uses a different style, maybe it's easier to read, probably it's easier to copy-paste from.

bitcoinfinancier (approved):

Byron:
Closed due to rejection.

tacerus:
I understand, thanks for the comment.

PR #1634, opened 2023-08-27
feat: full typing for "progress" parameter in Repo class

madebylydia:
This commit also fix a few inconsistency, most especially:
— How mypy/Pylance are checking for if-statements
— Tools complaining about not subscriptable class
— Exporting imports

Tested against Python 3.7, tests were (supposedly) ran successfully.


git/types.py
```diff
@@ -8,42 +8,39 @@
...
         Literal,
-        SupportsIndex,  # noqa: F401
+        SupportsIndex as SupportsIndex,
         TypedDict,
         Protocol,
         runtime_checkable,
     )  # noqa: F401
 
-# if sys.version_info[:2] >= (3, 10):
+# if sys.version_info >= (3, 10):
```
Byron, line 38: Is this a comment that is relevant to the typing system? If not, I think it can be deleted.

Byron (approved):
Thanks a lot for these improvements!

If you have an idea on how to improve CI I'd definitely welcome another PR. Right now there [is only mypi](https://github.com/gitpython-developers/GitPython/actions/runs/5992039528/job/16251031991?pr=1634), but apparently it doesn't catch the things you are fixing.

git/types.py
madebylydia: Yes, relevant for some linting tools, like Pylance or mypy it seems.

Pylance most especially, will import packages and look for any `if-else` import, but only work if not manually reducing the tuple, you can easily notice as the text become grayed.

I think this change will also fix one of my issue I had with generating stubs with mypy.

PR #1644, opened 2023-09-05
Fix CVE-2023-41040

facutuesca:
This change adds a check during reference resolving to see if the requested reference is inside the current repository folder. If it's ouside, it raises an exception.

This fixes [CVE-2023-41040](https://github.com/advisories/GHSA-cwvm-v4w8-q58c), which allows an attacker to access files outside the repository's directory.

This closes https://github.com/gitpython-developers/GitPython/issues/1638.

### DIFF
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -168,6 +168,8 @@ class SymbolicReference(object):
         """Return: (str(sha), str(target_ref_path)) if available, the sha the file at
         rela_path points to, or None. target_ref_path is the reference we
         point to, or None"""
+        if ".." in str(ref_path):
+            raise ValueError(f"Invalid reference '{ref_path}'")
         tokens: Union[None, List[str], Tuple[str, str]] = None
         repodir = _git_dir(repo, ref_path)
         try:
--- a/test/test_refs.py
+++ b/test/test_refs.py
@@ -5,6 +5,7 @@
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
 
 from itertools import chain
+from pathlib import Path
 
 from git import (
     Reference,
@@ -20,9 +21,11 @@ from git import (
 from git.objects.tag import TagObject
 from test.lib import TestBase, with_rw_repo
 from git.util import Actor
+from gitdb.exc import BadName
 
 import git.refs as refs
 import os.path as osp
+import tempfile
 
 
 class TestRefs(TestBase):
@@ -616,3 +619,15 @@ class TestRefs(TestBase):
 
     def test_reflog(self):
         assert isinstance(self.rorepo.heads.master.log(), RefLog)
+
+    def test_refs_outside_repo(self):
+        # Create a file containing a valid reference outside the repository. Attempting
+        # to access it should raise an exception, due to it containing a parent directory
+        # reference ('..'). This tests for CVE-2023-41040.
+        git_dir = Path(self.rorepo.git_dir)
+        repo_parent_dir = git_dir.parent.parent
+        with tempfile.NamedTemporaryFile(dir=repo_parent_dir) as ref_file:
+            ref_file.write(b"91b464cd624fe22fbf54ea22b85a7e5cca507cfe")
+            ref_file.flush()
+            ref_file_name = Path(ref_file.name).name
+            self.assertRaises(BadName, self.rorepo.commit, f"../../{ref_file_name}")


Byron (changes requested):
Thanks for the initiative! Could you add some tests that fail if the change is not applied? 
Thanks

facutuesca:
> Thanks for the initiative! Could you add some tests that fail if the change is not applied? Thanks

Done!

git/refs/symbolic.py
```diff
@@ -171,7 +172,13 @@ def _get_ref_info_helper(
         tokens: Union[None, List[str], Tuple[str, str]] = None
         repodir = _git_dir(repo, ref_path)
         try:
-            with open(os.path.join(repodir, str(ref_path)), "rt", encoding="UTF-8") as fp:
+            # Make path absolute, resolving any symlinks, and check that we are still
+            # inside the repository
+            full_ref_path = Path(repodir, str(ref_path)).resolve()
+            if Path(repodir).resolve() not in full_ref_path.parents:
```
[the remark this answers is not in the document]
facutuesca: Why not resolve the symlink?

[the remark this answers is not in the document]
facutuesca: But repodir is not given by the user, rather it's computed using:
```python
repodir = _git_dir(repo, ref_path)
```
which returns (if I understand it correctly) the repo's `.git` folder. Is there a scenario where that folder can be a symlink?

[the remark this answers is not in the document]
facutuesca: I guess my point is: if your `.git` directory is a symlink to `/home/secrets`, `GitPython` _will_ have access to everything under `/home/secrets`, so no point in protecting against that in this very specific function

git/refs/symbolic.py
```diff
@@ -171,7 +172,13 @@ def _get_ref_info_helper(
         tokens: Union[None, List[str], Tuple[str, str]] = None
         repodir = _git_dir(repo, ref_path)
         try:
-            with open(os.path.join(repodir, str(ref_path)), "rt", encoding="UTF-8") as fp:
+            # Make path absolute, resolving any symlinks, and check that we are still
+            # inside the repository
+            full_ref_path = Path(repodir, str(ref_path)).resolve()
```
Byron, line 177: I think this will slow down each read of a loose ref, and maybe that can be avoided?
References mentioned in a symbolic ref are never absolute, and can be rejected on that ground. Then, when determined relative, one should normalize the `../` away to check if it would break out. So `refs/../foo` would be fine, but `refs/../foo/../../bar` would not be. All this can happen without resolving symlinks.

test/test_refs.py
```diff
@@ -616,3 +619,14 @@ def test_dereference_recursive(self):
...
+    def test_refs_outside_repo(self):
+        # Create a file containing a valid reference outside the repository. Attempting
+        # to access it should raise an exception. This tests for CVE-2023-41040.
+        git_dir = Path(self.rorepo.git_dir)
+        repo_parent_dir = git_dir.parent.parent
+        with tempfile.NamedTemporaryFile(dir=repo_parent_dir) as ref_file:
+            ref_file.write(b"91b464cd624fe22fbf54ea22b85a7e5cca507cfe")
+            ref_file.flush()
+            ref_file_name = Path(ref_file.name).name
+            self.assertRaises(BadName, self.rorepo.commit, f"../../{ref_file_name}")
```
Byron, line 632: If this raises `BadName`, would that mean it actually read the file? I would expect a `ValueError` here. I think to clarify this, the exception should be checked for a particular message.

Byron (changes requested):
Thanks for taking a look into this!

I have come up with a couple of suggestions and hope they make sense.

test/test_refs.py
facutuesca: No, if it raises `BadName`, it means that we did not read the file. We assert that `BadName` is raised because that is the correct behavior (before the fix, no exception would be raised.)

facutuesca: The reason why we expect `BadName` rather than `ValueError` is because the `ValueError` exception we raise is caught and ends up being propagated as `BadName`:
https://github.com/gitpython-developers/GitPython/blob/e50046688f734f65f452de9b8feb10189efd7c1b/git/repo/fun.py#L172-L179

git/refs/symbolic.py
facutuesca: I replaced the calls to `resolve()` with `os.path.abspath`, which normalizes `..` but does not resolve symlinks.

git/refs/symbolic.py
```diff
@@ -171,7 +172,14 @@ def _get_ref_info_helper(
         tokens: Union[None, List[str], Tuple[str, str]] = None
         repodir = _git_dir(repo, ref_path)
         try:
-            with open(os.path.join(repodir, str(ref_path)), "rt", encoding="UTF-8") as fp:
```
Byron, line 174: When looking at how `gitoxide` (and `git`) do it I noticed that this could be much simpler.
Ref-names can't be anything they want, they are very limited in what's allowed and what is not (see [this validation code as reference](https://github.com/Byron/gitoxide/blob/dff90531e49a153a38c1cb01c8fdc7ff9bad7e81/gix-validate/src/tag.rs#L48)).

It would be enough to reject any `ref_path` that has a parent-dir component in it.

What do you think?

facutuesca: Ah yeah you're right! I fixed it so that now it only checks for `..` in the `ref_path`

Byron (approved):
Thanks a lot for contributing a CVE-fix!

facutuesca:
> @facutuesca, @Byron, wait but issue is still there. Because of how os.path.join works it is possible to [exploit it](https://github.com/gitpython-developers/GitPython/blob/1256b1640fe9636d71acff969443d77ae3ff6b84/git/refs/symbolic.py#L176) with just a passing absolute path.
> 
> Like
> 
> ```
> import git
> r = git.Repo(".")
> r.commit("/dev/null")
> ```
@doc-sheet 
Ah I see. Would resolving `repodir` to an absolute path before calling `os.path.join` be enough to fix it?
```python
            abs_repodir = os.path.abspath(repodir)
            with open(os.path.join(abs_repodir, str(ref_path)), "rt", encoding="UTF-8") as fp:
```

Byron:
This would certainly be feasible if the absolute version of a repository path is stored on the repository itself as a sort of cache.

It's definitely something that would need addressing.

CC @EliahKagan

EliahKagan:
Couldn't the `ref_path` check in `_get_ref_info_helper` that this PR added just be made more robust, so it rejects non-relative paths, as well as those that contain `..`? Currently we have:

https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/refs/symbolic.py#L171-L172

I think the new cases to reject include absolute paths, but also paths that are neither absolute nor relative. The latter is possible on Windows, where you can have a path like `C:a` that refers to `a` in the "current directory" on the `C:` drive *even when `C:` is not the current drive* (see the note on [this page](https://learn.microsoft.com/en-us/windows/win32/fileio/changing-the-current-directory)), as well as a path like `/a` or `\a` that refers to `a` in the root of the current drive.

Neither is consistently considered absolute by the Python standard library. Neither `os.path.isabs` nor the `pathlib.Path.is_absolute` method consider `C:` absolute. Interestingly, they disagree on Windows about `/a` and `\a`. I don't mean the `/` and `\` versions are treated differently--those are treated the same, that's not the issue. Rather, `pathlib.Path` is aware such paths are not, in the strictest sense, absolute paths on Windows, while `os.path.isabs` is treats them as absolute. Both `C:` and `/a` or `\a` have joining behavior similar to that which [doc-sheet pointed out](https://github.com/gitpython-developers/GitPython/pull/1644#issuecomment-1726331886) pointed out true absolute paths have.

I am reminded of [some code](https://github.com/dmvassallo/EmbeddingScratchwork/blob/9fb8b1906a7372263577a97dc7dac8615dc5f5ed/embed/demos/usc.py#L78-L89) I wrote a few months ago, in a different context but also about avoiding directory traversal:

```python
path = Path(name)
if path.is_absolute():
    # Absolute paths can extract outside the target directory.
    raise zipfile.BadZipFile(f'archive has absolute path: {name!r}')
if path.root or path.drive:
    # Non-relative non-absolute paths on Windows can do the same.
    raise zipfile.BadZipFile(f'archive has non-relative path: {name!r}')
if '..' in name:
    # A ".." component, or "...", "....", etc. on some systems, can
    # traverse upward. Because we know what is reasonable in a USC
    # archive, broadly denying paths with ".." anywhere is okay.
    raise zipfile.BadZipFile(f'archive has name containing "..": {name!r}')
```

That would have to be adapted slightly--for example, we are not raising `BadZipFile`--but I think something like that could be done.

facutuesca:
@EliahKagan @doc-sheet @Byron  What about implementing the rules in https://git-scm.com/docs/git-check-ref-format/ and using them to check reference names? It looks like rules 3, 4, 6 and 10 would cover the problematic behaviors we're talking about, plus more:

3. They cannot have two consecutive dots `..` anywhere.
4. They cannot have ASCII control characters (i.e. bytes whose values are lower than \040, or \177 `DEL`), space, tilde `~`, caret `^`, **or colon `:`** anywhere.
6. They cannot begin or end with a slash `/` or contain multiple consecutive slashes
10. They cannot contain a `\`.

Byron:
Thanks everyone! 

https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/refs/symbolic.py#L171-L172

It seems the above function could be upgraded to do more complete validation similar to https://git-scm.com/docs/git-check-ref-format/ .

`gitoxide` implements this [here](https://github.com/Byron/gitoxide/blob/a9da3f84b979aa25225a819428b85bff6290da9b/gix-validate/src/reference.rs#L47), and was itself written based on `git`'s validation code - maybe it can be useful here when implementing a python version.

facutuesca:
@Byron @doc-sheet @EliahKagan I have created a PR for this here: https://github.com/gitpython-developers/GitPython/pull/1672

EliahKagan:
By the way, for the benefit of anyone who comes along to read this, the example in https://github.com/gitpython-developers/GitPython/pull/1644#issuecomment-1726331886 may not look like it shows anything, but this is because reading from `/dev/null` completes immediately and the resulting exception and traceback from GitPython doesn't obviously indicate that this happened. If a path like `/dev/full` or `/dev/random` is passed, it keeps reading and reading, and using system resources, etc. (Thus, you may *not* want to try that on your system, at least if you're not prepared to kill the process.)

So in a situation where an application that uses GitPython exposes queries on refs in specific local repositories without the intention of allowing reads from elsewhere in the filesystem, using an absolute path rather than a relative one with `..` components provided another way to exploit the vulnerability that was not prevented by the original patch. (At least that is my understanding of the situation.)

Issue #1544, opened 2023-01-21
Tests fail due to security vulnerability fix in git 2.38.1


Lightborne:
Hello,
Due to a change made in Git to address a security vulnerability, some tests are failing.

See here for a description of the change:

https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253

These are the failing tests:

- test_list_only_valid_submodules
- test_git_submodules_and_add_sm_with_new_commit

The fail signature is the same in both cases:

    cmdline: git submodule add /[redacted]/GitPython/git/ext/gitdb/gitdb/ext/smmap module
    stderr: 'Cloning into '/tmp/test_list_only_valid_submoduleshv3nprno/parent/module'...
    fatal: transport 'file' not allowed
    fatal: clone of '/[redacted]/GitPython/git/ext/gitdb/gitdb/ext/smmap' into submodule path '/tmp/test_list_only_valid_submoduleshv3nprno/parent/module' failed'

Here is a blog post discussing this issue affecting others:

https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html

I have fixed this locally by changing the submodule add command in each test from:

    repo.git.submodule("add", self._small_repo_url(), "module")

to 

    repo.git.submodule("add", Git.polish_url("https://github.com/gitpython-developers/smmap.git"), "module")

If this is an acceptable fix I can provide it in a pull request.

Lightborne:
Pull request that fixes this is here:

https://github.com/gitpython-developers/GitPython/pull/1546

Byron:
I wonder why the [fix applied on CI](https://github.com/gitpython-developers/GitPython/blob/3901d4ccdbd7c2450ff6faadebf2d33e2b246a7b/.github/workflows/pythonpackage.yml#L66) isn't feasible for you.

I'd like to avoid cloning from a remote repository, and prefer a local fix. Maybe the git configuration change that is currently applied on CI can be enforced before running any GitPython test, using some sort of global initialization? Git configuration can be affected through environment variables for example, which would certainly work here.

Lightborne:
I think you made the right call declining my PR - it was a hacky fix.  

> Maybe the git configuration change that is currently applied on CI can be enforced before running any GitPython test, using some sort of global initialization?

Would this solution modify a user's git configuration without them knowing?  I don't know how I feel about that.  

[...]

PR #1645, opened 2023-09-05
Fix 'Tree' object has no attribute '_name' when submodule path is normal path

CosmosAtlas:
When a repository have a `.gitmodules` file and the actual path defined in the git repository is not related to submodule, GitPython still assumes its a submodule.

To replicate:
```bash
# In a temp directory
mkdir test
cd test
git init
mkdir DbConnector
touch DbConnector/readme.md
touch .gitmodules
# Put the context of the next segment in .gitmodules
git add .
git commit -m "test"
```

```ini
[submodule "DbConnector"]
    path = DbConnector
    url = https://github.com/chaconinc/DbConnector
```

Then in a python section
```python
from git.repo import Repo
r = Repo('.')
r.submodules
```

the following error occurs
```
In [3]: r.submodules
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 r.submodules

File ~/.local/share/virtualenvs/GitPython-yvdsleke/lib/python3.11/site-packages/git/repo/base.py:423, in Repo.submodules(self)
    418 @property
    419 def submodules(self) -> "IterableList[Submodule]":
    420     """
    421     :return: git.IterableList(Submodule, ...) of direct submodules
    422         available from the current head"""
--> 423     return Submodule.list_items(self)

File ~/.local/share/virtualenvs/GitPython-yvdsleke/lib/python3.11/site-packages/git/util.py:1184, in IterableObj.list_items(cls, repo, *args, **kwargs)
   1175 """
   1176 Find all items of this type - subclasses can specify args and kwargs differently.
   1177 If no args are given, subclasses are obliged to return all items if no additional
   (...)
   1181
   1182 :return: list(Item,...) list of item instances"""
   1183 out_list: IterableList = IterableList(cls._id_attribute_)
-> 1184 out_list.extend(cls.iter_items(repo, *args, **kwargs))
   1185 return out_list

File ~/.local/share/virtualenvs/GitPython-yvdsleke/lib/python3.11/site-packages/git/objects/submodule/base.py:1406, in Submodule.iter_items(cls, repo, parent_commit, *Args, **kwargs)
   1401         continue
   1402     # END handle keyerror
   1403 # END handle critical error
   1404
   1405 # fill in remaining info - saves time as it doesn't have to be parsed again
-> 1406 sm._name = n
   1407 if pc != repo.commit():
   1408     sm._parent_commit = pc

AttributeError: 'Tree' object has no attribute '_name'
```

[...]

PR #1647, opened 2023-09-07
Added test for PR #1645 submodule path

CosmosAtlas:
Following from here: https://github.com/gitpython-developers/GitPython/pull/1645#issuecomment-1708908630, I've added a test in the corresponding position.

Meanwhile, I've noticed there are test failures related to [CVE-2022-39253](https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html). I did a quick fix in place by adding git configs in the Cmd.

Since I also had the chance to do a fresh start on testing, I updated the docs on where it was unclear to me and caused issues. Particularly, I highlighted that `git fetch --tags` should be ran first otherwise it would cause test failures regarding version `0.1.6` something something.

Each small change in clearly separated by commits, so if there is any change you don't like, it can be easily cherry picked.

Lastly, I would like to thank you for being so responsive in maintaining this wonderful project 😄

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,4 +51,5 @@ Contributors are:
 -Luke Twist <itsluketwist@gmail.com>
 -Joseph Hale <me _at_ jhale.dev>
 -Santos Gallegos <stsewd _at_ proton.me>
+-Wenhan Zhu <wzhu.cosmos _at_ gmail.com>
 Portions derived from other open source works and are clearly marked.
--- a/README.md
+++ b/README.md
@@ -93,9 +93,9 @@ See [Issue #525](https://github.com/gitpython-developers/GitPython/issues/525).
 
 ### RUNNING TESTS
 
-_Important_: Right after cloning this repository, please be sure to have executed
-the `./init-tests-after-clone.sh` script in the repository root. Otherwise
-you will encounter test failures.
+_Important_: Right after cloning this repository, please be sure to have
+executed `git fetch --tags` followed by the `./init-tests-after-clone.sh`
+script in the repository root. Otherwise you will encounter test failures.
 
 On _Windows_, make sure you have `git-daemon` in your PATH. For MINGW-git, the `git-daemon.exe`
 exists in `Git\mingw64\libexec\git-core\`; CYGWIN has no daemon, but should get along fine
--- a/test/test_submodule.py
+++ b/test/test_submodule.py
@@ -907,6 +907,28 @@ class TestSubmodule(TestBase):
         # end for each dry-run mode
 
     @with_rw_directory
+    def test_ignore_non_submodule_file(self, rwdir):
+        parent = git.Repo.init(rwdir)
+
+        smp = osp.join(rwdir, "module")
+        os.mkdir(smp)
+
+        with open(osp.join(smp, "a"), "w", encoding="utf-8") as f:
+            f.write('test\n')
+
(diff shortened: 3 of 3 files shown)


EliahKagan:
Does this (specifically, [69ab73b](https://github.com/gitpython-developers/GitPython/pull/1647/commits/69ab73b4a56669e22ba6060c9283b1d207767749)) close #1544?

I didn't notice that this PR had already recently been opened, when I subsequently opened #1648. This and #1648 partially overlap in what they fix: they both edit `test_submodule.py`, taking different approaches, but solving the same problem. I am not sure which approach is better. I've stated the differences between our PRs that relate to #1544, as I understand them, in an edit at the top of the #1648.

The two PRs also have non-overlapping parts, in the changes they make outside `test_submodule.py`.

If this closes #1544, then it seems to me that it may make sense to extend it to also remove the commands setting `protocol.file.allow` in the `pythonpackage.yml` and `cygwin-test.yml` workflows.

CosmosAtlas:
For future proofness I think #1648 may be better, as I suggested in the original comment, I basically simply edited it for it to work.

Meanwhile, to address the same issue, there's also another approach that seems promising, but I couldn't get it working when tried for a bit.

GitPython supports editing the config file, so in the testing environment, it's possible to just modify the config for the local repo through something like the following 

```
with repo.config_writer() as writer:
    writer.set_value("protocol.file.allow", "always")
```

EliahKagan:
If a `config_writer`-based approach were used, would it be used on the superproject (the repository that is going to have a submodule)? If so, I don't think that will work, because submodules don't inherit *local* configuration from the superprojects they are added to. I wasn't sure if that applied even to allowed protocols--intuitively it feels like it should not--but it seems it does:

```text
ek@Kip:~/source/repos/example-superproject-ii (main =)$ ls /repos/example-submodule.git
branches  config  description  HEAD  hooks  info  objects  refs
ek@Kip:~/source/repos/example-superproject-ii (main =)$ git submodule add /repos/example-submodule.git
Cloning into '/home/ek/source/repos/example-superproject-ii/example-submodule'...
fatal: transport 'file' not allowed
fatal: clone of '/repos/example-submodule.git' into submodule path '/home/ek/source/repos/example-superproject-ii/example-submodule' failed
ek@Kip:~/source/repos/example-superproject-ii (main =)[128]$ git config protocol.file.allow always
ek@Kip:~/source/repos/example-superproject-ii (main =)$ git submodule add /repos/example-submodule.git
Cloning into '/home/ek/source/repos/example-superproject-ii/example-submodule'...
fatal: transport 'file' not allowed
fatal: clone of '/repos/example-submodule.git' into submodule path '/home/ek/source/repos/example-superproject-ii/example-submodule' failed
```

CosmosAtlas:
Thanks for the update. It explained why I couldn't get it working (because it shouldn't!).

Byron (changes requested):
Thanks a lot for the test, it's much appreciated! Tests are the reason this project is still on life-support, after all.

Could you rebase onto `main` and see if the additional `-c"protocol…` line can the be removed?

Thanks you.

----

> Lastly, I would like to thank you for being so responsive in maintaining this wonderful project 😄

Thank you :). I just don't know if it's a *wonderful* project, but I'll leave that to you ;).

Issue #1052, opened 2020-09-02
Race condition in multithreading environment due to usage of os.chdir


rcarpa:
Hi, 

I recently ran into a race condition when using two different instances of `git.Repo` objects in two different threads. 

Inside `repo.index.add()`, some functions are decorated with the `git_working_dir` decorator. This decorator changes the process working directory to the one of repository, performs some system calls, then returns the working dir back to the initial value. 

If two threads perform `repo.index.add()`, even on two independent `Repo()` objects, the working directory of the entire process could be permanently modified. Moreover, system calls done by `index.add()` may fail. For example `lstat()` will be executed with a relative file name assuming a wrong absolute path.

For example: 
1. working dir (wd) is /home/user
2. git_working_dir is entered in thread 1. (/home/user is registered, wd changed to /repo1)
3. git_working_dir is entered in thread 2. (/repo1 is registered, wd changed to /repo2)
4. git_working_dir is exited in thread 1. (wd restored to /home/user)
5. git_working_dir is exited in thread 2. (wd restored to /repo1)
now the process wd is /repo1

I believe I could work around this issue by creating git object directly rather than adding to the index then committing, but I thought you may be interested to know about the issue. 

Radu

Byron:
Thanks for the detailed report. I am leaving it open for others to find, despite having no hopes of this being tackled.

Generally, GitPython is not thread-safe, but if desired, certain types of actions cold be made to allow multiple threads under the condition that these operate on their own repositories.

rcarpa:
Seems like the mentioned decorator is only used to protect calls to _store_path. Would it be difficult to ensure that this function does system calls only on absolute paths and performs conversion to relative names (to add to git) by path manipulation (prefix stripping)? 
Curious to know why it is complicated.

Byron:
That should definitely be possible, the reason it is currently implemented like this is simplicity, delegating all path handling (the devil is in the details especially windows is involved) to the operating system.

rcarpa:
Ah ... didn't think about multi-platforming at all :( . Thanks for the explanation !

PR #1650, opened 2023-09-07
Fix Windows environment variable upcasing bug

### LINKED ISSUE #1646: CVE-2023-40590 fix capitalized all environment variables on Windows
This fix:
https://github.com/gitpython-developers/GitPython/commit/6029211d729a0dd81e08fcc9c1a3ab7fe9af85c9
capitalized all environment variables on Windows. It can be illustrated by this short program:
```
import subprocess

print(subprocess.check_output(
    "set | findstr /c:SystemRoot /i", shell=True, universal_newlines=True
))

import git

print(subprocess.check_output(
    "set | findstr /c:SystemRoot /i", shell=True, universal_newlines=True
))
```

The output is:
```
SystemRoot=C:\Windows

SYSTEMROOT=C:\Windows
```

This side effect breaks our use case currently. We use gnu make in cygwin for our build, in which all environment variables are case sensitive.

The core problem was `unittest.mock.patch.dict(os.environ, {"NoDefaultCurrentDirectoryInExePath": "1"})` -- in which it will try to treat `os.environ` as a dictionary, but `os.environ` is not just a simple dictionary. It actually remembers the original casing of the environment variable. Unfortunately when reading it as dictionary it capitalize all letters.

We can also observe the same side effect with this code below:
```
import os
import subprocess
import unittest.mock

print(subprocess.check_output(
    "set | findstr /c:SystemRoot /i", shell=True, universal_newlines=True
))

with unittest.mock.patch.dict(os.environ, {"NoDefaultCurrentDirectoryInExePath": "1"}):
    pass

print(subprocess.check_output(
    "set | findstr /c:SystemRoot /i", shell=True, universal_newlines=True
))
```

The side effect is the same as above.

EliahKagan:
*This pull request might be ready, but it is still a draft because I think I should look it over once more for any serious problems.*

Fixes #1646

This fixes the bug I inadvertently introduced in #1636 where all environment variables' names are set upper-case on Windows.

It uses a simple hand-rolled context manager to patch the `NoDefaultCurrentDirectoryInExePath` variable, instead of `unittest.mock.patch.dict`. The latter was setting unrelated environment variables to the original (same) values via `os.environ`, and as a result, their names were all converted to upper-case on Windows. This is the intended behavior of `os.environ` on Windows, at least in CPython, per https://github.com/python/cpython/issues/73010 and https://github.com/python/cpython/issues/101754, but I had not been thinking of this, and more importantly I was not aware that `unittest.mock.patch.dict` effectively writes all items to the mapping when it unpatches, including items that had not been listed for patching.

Because only environment variables that are actually set through `os.environ` have their names upcased, the only variable whose name should be upcased now is `NoDefaultCurrentDirectoryInExePath`, which should be fine. (It has a single established use/meaning in Windows, where it's treated case-insensitively as environment variables in Windows *usually* are.)

---

The actual fix here is pretty simple, but robustly testing it is tricky. One pitfall in testing it remains unaddressed at this point: Either the bug doesn't affect Cygwin builds of Python (linking `cygwin1.dll`), or `is_win` evaluates to false in the `test-cygwin.yml` workflows, **so the project's current CI cannot verify the fix.** Notice, for example, that the first two commits ([d88372a](https://github.com/gitpython-developers/GitPython/pull/1650/commits/d88372a11ac145d92013dcc64b7d21a5a6ad3a91) and [7296e5c](https://github.com/gitpython-developers/GitPython/pull/1650/commits/7296e5c021450743e5fe824e94b830a73eebc4c8)) show as passing all tests, even though they should show as failing on Windows. I am unsure if that should block this PR or not, since it seems like this sort of thing might actually be common.

I could make the test run on more platforms than Windows, but I don't think that would be more robust, because on other systems it would effectively just be testing that the Python implementation of `os.environ` works as expected on non-Windows systems. Furthermore, the important test logic would differ depending on the value of `is_win`, so if the Cygwin workflow has `is_win` as false, then it would only create the false impression that CI was verifying the fix.

It might be possible to improve the situation in the CI workflows themselves. But I think that would take longer than it might be good to wait on fixing this bug, since this bug currently prevents users who rely on case-sensitive environment variables from upgrading far enough to receive fixes for either CVE-2023-40590 (3.1.33) or CVE-2023-41040 (3.1.35, forthcoming).

[**I have verified the fix locally.**](https://gist.github.com/EliahKagan/21916264b96b302e13fd105f5ba2aae6) That test failure is from *before* the fix in [c7fad20](https://github.com/gitpython-developers/GitPython/pull/1650/commits/c7fad20be5df0a86636459bf673ff9242a82e1fc). It is taken from [7296e5c](https://github.com/gitpython-developers/GitPython/pull/1650/commits/7296e5c021450743e5fe824e94b830a73eebc4c8).

Both tests shown there pass starting in the third commit ([c7fad20](https://github.com/gitpython-developers/GitPython/pull/1650/commits/c7fad20be5df0a86636459bf673ff9242a82e1fc)), as expected, and continue to pass in the fourth commit ([eebdb25](https://github.com/gitpython-developers/GitPython/pull/1650/commits/eebdb25ee6e88d8fce83ea0970bd08f5e5301f65)). Due to the new test's use of the new "fixture" script `env_case.py`, together with how the official installation instructions do not create a true editable install, if `python setup.py install` was how installation was done, and it was done prior to checking out the fix, then `python setup.py install` has to be run again for the "fixture" script to get the current code when it runs `import git`. I considered going back to the previous way that doesn't execute a separate script file, but I think that code may be harder to understand for someone who doesn't already know how the test works.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -158,6 +158,20 @@ def cwd(new_dir: PathLike) -> Generator[PathLike, None, None]:
         os.chdir(old_dir)
 
 
+@contextlib.contextmanager
+def patch_env(name: str, value: str) -> Generator[None, None, None]:
+    """Context manager to temporarily patch an environment variable."""
+    old_value = os.getenv(name)
+    os.environ[name] = value
+    try:
+        yield
+    finally:
+        if old_value is None:
+            del os.environ[name]
+        else:
+            os.environ[name] = old_value
+
+
 def rmtree(path: PathLike) -> None:
     """Remove the given recursively.
 
@@ -935,7 +949,7 @@ class LockFile(object):
             )
 
         try:
-            with open(lock_file, mode='w'):
+            with open(lock_file, mode="w"):
                 pass
         except OSError as e:
             raise IOError(str(e)) from e
--- a/git/util.py
+++ b/git/util.py
@@ -150,6 +150,7 @@ def unbare_repo(func: Callable[..., T]) -> Callable[..., T]:
 
 @contextlib.contextmanager
 def cwd(new_dir: PathLike) -> Generator[PathLike, None, None]:
+    """Context manager to temporarily change directory. Not reentrant."""
     old_dir = os.getcwd()
     os.chdir(new_dir)
     try:
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -112,16 +112,12 @@ class TestGit(TestBase):
             raise RuntimeError("test bug or strange locale: old_name invariant under upcasing")
         os.putenv(old_name, "1")  # It has to be done this lower-level way to set it lower-case.
 
-        script_lines = [
-            "import subprocess, git",
-
-            # Importing git should be enough, but this really makes sure Git.execute is called.
-            f"repo = git.Repo({self.rorepo.working_dir!r})",
-            "git.Git(repo.working_dir).execute(['git', 'version'])",
-
-            f"print(subprocess.check_output(['set', {old_name!r}], shell=True, text=True))",
+        cmdline = [
+            sys.executable,
+            fixture_path("env_case.py"),
+            self.rorepo.working_dir,
+            old_name,
         ]
-        cmdline = [sys.executable, "-c", "\n".join(script_lines)]
         pair_text = subprocess.check_output(cmdline, shell=False, text=True)
         new_name = pair_text.split("=")[0]
         self.assertEqual(new_name, old_name)
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -10,7 +10,7 @@ import shutil
 import subprocess
 import sys
 from tempfile import TemporaryDirectory, TemporaryFile
-from unittest import mock
+from unittest import mock, skipUnless
 
 from git import Git, refresh, GitCommandError, GitCommandNotFound, Repo, cmd
 from test.lib import TestBase, fixture_path
@@ -104,6 +104,27 @@ class TestGit(TestBase):
 
             with _chdir(tmpdir):
                 self.assertRegex(self.git.execute(["git", "version"]), r"^git version\b")
+
+    @skipUnless(is_win, "The regression only affected Windows, and this test logic is OS-specific.")
+    def test_it_avoids_upcasing_unrelated_environment_variable_names(self):
+        old_name = "28f425ca_d5d8_4257_b013_8d63166c8158"
+        if old_name == old_name.upper():
+            raise RuntimeError("test bug or strange locale: old_name invariant under upcasing")
+        os.putenv(old_name, "1")  # It has to be done this lower-level way to set it lower-case.
+
+        script_lines = [
+            "import subprocess, git",
+
+            # Importing git should be enough, but this really makes sure Git.execute is called.
+            f"repo = git.Repo({self.rorepo.working_dir!r})",
+            "git.Git(repo.working_dir).execute(['git', 'version'])",
+
+            f"print(subprocess.check_output(['set', {old_name!r}], shell=True, text=True))",
+        ]
+        cmdline = [sys.executable, "-c", "\n".join(script_lines)]
+        pair_text = subprocess.check_output(cmdline, shell=False, text=True)
+        new_name = pair_text.split("=")[0]
+        self.assertEqual(new_name, old_name)
 
     def test_it_accepts_stdin(self):
         filename = fixture_path("cat_file_blob")
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -4,7 +4,6 @@
 #
 # This module is part of GitPython and is released under
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php
-import contextlib
 import os
 import shutil
 import subprocess
@@ -15,22 +14,11 @@ from unittest import mock, skipUnless
 from git import Git, refresh, GitCommandError, GitCommandNotFound, Repo, cmd
 from test.lib import TestBase, fixture_path
 from test.lib import with_rw_directory
-from git.util import finalize_process
+from git.util import cwd, finalize_process
 
 import os.path as osp
 
 from git.compat import is_win
-
-
-@contextlib.contextmanager
-def _chdir(new_dir):
-    """Context manager to temporarily change directory. Not reentrant."""
-    old_dir = os.getcwd()
-    os.chdir(new_dir)
-    try:
-        yield
-    finally:
-        os.chdir(old_dir)
 
 
 class TestGit(TestBase):
@@ -102,7 +90,7 @@ class TestGit(TestBase):
                     print("#!/bin/sh", file=file)
                 os.chmod(impostor_path, 0o755)
 
-            with _chdir(tmpdir):
+            with cwd(tmpdir):
                 self.assertRegex(self.git.execute(["git", "version"]), r"^git version\b")
 
     @skipUnless(is_win, "The regression only affected Windows, and this test logic is OS-specific.")
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -14,7 +14,6 @@ from subprocess import call, Popen, PIPE, DEVNULL
 import subprocess
 import threading
 from textwrap import dedent
-import unittest.mock
 
 from git.compat import (
     defenc,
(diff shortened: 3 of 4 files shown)


git/util.py
```diff
@@ -150,6 +150,7 @@ def wrapper(self: "Remote", *args: Any, **kwargs: Any) -> T:
 
 @contextlib.contextmanager
 def cwd(new_dir: PathLike) -> Generator[PathLike, None, None]:
+    """Context manager to temporarily change directory. Not reentrant."""
```
Byron, line 153: I thought it was interesting to share my thought on the word 'reentrant' , as to me it seems it actually is reentrant and it will work deterministically. It's not advisable to use this when running tests in parallel though merely because its effects are observable within the whole process.

Please do ignore, it doesn't matter :D.

irwand:
This works for single thread, but because it temporarily change `os.environ`, this could cause unexpected race-condition with other threads that are doing something relying on `os.environ`. This bug will be even harder to detect and catch.

One of the best way to fix this is to re-implement `shutil.which()` that does not prepend current cwd into the `PATH` when looking for `git`, and then just use the `git` path we found directly. This way we don't cause unexpected side effect by temporarily changing `os.environ`

git/util.py
```diff
@@ -935,7 +950,7 @@ def _obtain_lock_or_raise(self) -> None:
             )
 
         try:
-            with open(lock_file, mode='w'):
+            with open(lock_file, mode="w"):
```
Byron, line 953: I have seen various contributors fight over these and do wonder who will win ;).

test/test_git.py
```diff
@@ -102,9 +90,26 @@ def test_it_executes_git_not_from_cwd(self):
...
 
-            with _chdir(tmpdir):
+            with cwd(tmpdir):
                 self.assertRegex(self.git.execute(["git", "version"]), r"^git version\b")
 
+    @skipUnless(is_win, "The regression only affected Windows, and this test logic is OS-specific.")
+    def test_it_avoids_upcasing_unrelated_environment_variable_names(self):
+        old_name = "28f425ca_d5d8_4257_b013_8d63166c8158"
+        if old_name == old_name.upper():
+            raise RuntimeError("test bug or strange locale: old_name invariant under upcasing")
```
Byron, line 100: An amazing thought to validate this precondition! But it's true, in Python the locale affects string encoding so programs run differently depending on the settings of their runtime environment.

test/test_git.py
```diff
@@ -102,9 +90,26 @@ def test_it_executes_git_not_from_cwd(self):
                     print("#!/bin/sh", file=file)
                 os.chmod(impostor_path, 0o755)
 
-            with _chdir(tmpdir):
+            with cwd(tmpdir):
                 self.assertRegex(self.git.execute(["git", "version"]), r"^git version\b")
 
+    @skipUnless(is_win, "The regression only affected Windows, and this test logic is OS-specific.")
+    def test_it_avoids_upcasing_unrelated_environment_variable_names(self):
+        old_name = "28f425ca_d5d8_4257_b013_8d63166c8158"
```
Byron, line 98: I see, this variable name is used to assure nobody will possibly stumble over it, nor is it likely to clash with an existing variable.
I think it's fair as nobody is iterating environment variables either.

Byron (approved):
Thanks a lot for contributing a fix for a fix for the CVE! It's much appreciated, particularly because this is a strange issue and you even found a way to add a regression test.

Regarding CI, if you have interest in adding a way to improve CI coverage so that this kind of bug would have been caught, I'd definitely appreciate another PR.

PS: I would assign 'legendary' status to your PRs - they make the impression of being a work-log of sorts, maybe one that is part of your workflow or organize work even. Amazing.

git/util.py
irwand: `black` prefers `"`, and some developers automatically runs black in their editor :)

EliahKagan:
@irwand Yes, in #1636 I had not really thought carefully enough about the situations in which a data race on `os.environ` might occur. Per https://github.com/gitpython-developers/GitPython/issues/1052#issuecomment-685405738, GitPython is not thread-safe, with some concurrent operations such as `chdir` being unsafe while GitPython code is executing. However, a project that uses GitPython and separately mutates `os.environ` for a reason unrelated to GitPython should ideally be free of data races, and likely was free of them (unless also doing `chdir`) prior to #1636.

Although reimplementing `shutil.which` is one option, I would lean toward a variation of the approach I took in #1636: instead of patching `os.environ` or otherwise changing environment variables for the current process, use another layer of subprocess indirection (on Windows only), so GitPython creates a subprocess that sets `NoDefaultCurrentDirectoryInExePath` and then itself creates the `git` subprocess. I'm not really sure what is better, though.

I think it's worthwhile to merge this PR in order to fix the immediate problem in #1646. But a new issue should probably be opened for the race condition that you have rightly identified. I don't know if you'd prefer to open it or if I should.

git/util.py
EliahKagan: Yeah, I ran `black` on the file I was changing, which is how that change came in.

test/test_git.py
EliahKagan: Another reason I named it like that is that it sticks around after the test, because I don't think there's a reasonable way to unset it (though there are probably some unreasonable ways). Setting the variable with `os.putenv` sets it case-sensitively on all systems including Windows, but it is not reflected in `os.environ` (and `nt.environ` only has the variables as they existed when the process started).

irwand:
I really appreciate the fix and have no objection to merging this in 👍 .. this will allow us to continue upgrading to newer gitpython :-) . The `os.environ` side effect is there, but not sure practically how bad that would be. Really if the industry already identifies that including cwd when looking for executables is a bad idea, maybe NoDefaultCurrentDirectoryInExePath should just be default. Feel free to open the race condition issue. From my point of view this is good enough and the race condition issue is somewhat low priority. Thank you again!

git/util.py
Byron: I see, these are tool-based preferences by now. Is it possible to drop a `black.config` file in the root to configure this? I do prefer to keep it the way it is, personally.

irwand (approved):

git/util.py
irwand: No, `black` is opinionated code formatter. One of the reasons why we like it. We all have to sacrifice something in the name of unity, so we leave the decision to the formatting overlord. :)

git/util.py
EliahKagan: @Byron I know you've suggested I ignore this, but I realized there's actual an opportunity for improvement in what I wrote here. The important thing is that `cwd` cannot be used in quite all the ways the 3.11+ [`contextlib.chdir`](https://docs.python.org/3/library/contextlib.html#contextlib.chdir) can. I've included a docstring improvement in #1657.

Functions decorated with `contextlib.contextmanager` can be used in nested `with`-statements, but each call returns a non-reentrant context manager object. Entering it when it it has already been entered produces:

```text
AttributeError: '_GeneratorContextManager' object has no attribute 'args'
```

The simplest example of reentrancy `cwd` doesn't support looks like this, though no one would do this:

```python
cm = cwd("/path/to/dir")
with cm:
    with cm:
        ...
```

However, situations resembling this (but with more statements between the `with`s) could arise in practice:

```python
cm = cwd("/path/to/dir")
with cm:
    with cwd("/path/to/other/dir"):
        with cm:
            ...
```

`cwd` doesn't support that. Probably no one would assume it does, except there may be a temptation to conceptualize it as a backport of `contextlib.chdir`, which *does* support that. Anyway, this is still fine:

```python
with cwd("/path/to/dir"):
    with cwd("/path/to/other/dir"):
        with cwd("/path/to/dir"):
            ...
```

(Btw, I hope the ping is not an annoyance. I don't think GitHub otherwise notifies for new messages in *resolved* review threads, but I'm not sure.)

Byron: No worries about the ping, I appreciate the explanation as indeed wasn't aware that for the context-manager implementation here isn't re-entrant, cannot be nested. It's mind-boggling that this limitation applies as it's not obvious. Nonetheless, it's not at all a problem here either, like you said.

EliahKagan: I think the Python design choice that makes this unintuitive is the existence of `__enter__`. Some context manager objects, like files returned by `open`, acquire their resource on *construction*, with a trivial `__enter__`. Others acquire their resource on `__enter__` (and of those, some are resuable and reentrant, some are reusable and non-reentrant, and some are neither reusable nor reentrant). But the problem context managers exist to solve, caused by lack of RAII, is that destruction, *unlike construction*, isn't deterministic.

I personally like better the approaches in C# (`Dispose` methods and `using` statements) and Java (the `AutoCloseable` interface and `try`-with-resources). Like Python, those languages lack RAII and provide special syntax to do something that would otherwise have to be done with explicit `try`-`finally`. But with the corresponding features in those languages, constructing the object is expected to acquire the resource (at least conceptually, and in most cases actually), and the only special semantics pertain to *releasing* them.

PR #1654, opened 2023-09-10
Improve Python version and OS compatibility, fixing deprecations

### LINKED ISSUE #1640: Python 3.12: ModuleNotFoundError: No module named 'setuptools'
The [Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0-release-candidate-1-released/31137?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

Python 3.12.0 final will be released in one month: https://peps.python.org/pep-0693/

---

However, when adding 3.12 to the CI like https://github.com/hugovk/GitPython/commit/427164b56ac4dec10b7d3fea27e7485643550776, it [fails](https://github.com/hugovk/GitPython/actions/runs/6063520672/job/16450971945#step:7:333) with:

```pytb
―――――――――――――――――――――― TestInstallation.test_installation ――――――――――――――――――――――

self = <test.test_installation.TestInstallation testMethod=test_installation>
rw_dir = '/tmp/test_installationo1a7a3rg'

    @with_rw_directory
    def test_installation(self, rw_dir):
        self.setUp_venv(rw_dir)
        result = subprocess.run(
            [self.pip, "install", "-r", "requirements.txt"],
            stdout=subprocess.PIPE,
            cwd=self.sources,
        )
        self.assertEqual(
            0,
            result.returncode,
            msg=result.stderr or result.stdout or "Can't install requirements",
        )
        result = subprocess.run(
            [self.python, "setup.py", "install"],
            stdout=subprocess.PIPE,
            cwd=self.sources,
        )
>       self.assertEqual(
            0,
            result.returncode,
            msg=result.stderr or result.stdout or "Can't build - setup.py failed",
        )
E       AssertionError: 0 != 1 : Can't build - setup.py failed

test/test_installation.py:39: AssertionError
----------------------------- Captured stderr call -----------------------------
  File "/home/runner/work/GitPython/GitPython/setup.py", line 2, in <module>
    from setuptools import setup, find_packages
ModuleNotFoundError: No module named 'setuptools'
```

### LINKED ISSUE #1651: Installation test always fails on native Windows
GitPython supports and can be installed on native Windows systems (not just Cygwin), but the `TestInstallation.test_installation` test in `test/test_installation.py` always fails with a `FileNotFoundError`. This is not caught on CI because, for Windows, only Cygwin currently has CI coverage. The test failure happens for two reasons:

- **The test relies on the virtual environment having a `bin` directory**, but on Windows this is called `Scripts`. This is the immediate cause of the error, but if only this is fixed, a second, closely related incompatibility will cause the test to fail.
- **The test relies on the virtual environment having `python3` and `pip3` commands**, but Windows does not have these in virtual environments. Python virtual environments on all systems have `python` and `pip` commands (even on systems such as Debian and Ubuntu where those commands are not available globally or whose global versions are part of a legacy Python 2 installation).

The following output is produced on Windows 10 using Python 3.11.5, installing GitPython from the tip of the main branch (c8e303ff), but I believe the test would fail on all Windows systems with all versions of Python in the same way:

```text
(.venv) C:\Users\ek\source\repos\GitPython [main ≡]> pytest --no-cov -k test_installation
Test session starts (platform: win32, Python 3.11.5, pytest 7.4.2, pytest-sugar 0.9.7)
rootdir: C:\Users\ek\source\repos\GitPython
configfile: pyproject.toml
testpaths: test
plugins: cov-4.1.0, sugar-0.9.7
collected 504 items / 503 deselected / 1 selected


――――――――――――――――――――――――――――――――――――――――― TestInstallation.test_installation ――――――――――――――――――――――――――――――――――――――――――

self = <test.test_installation.TestInstallation testMethod=test_installation>
rw_dir = 'C:\\Users\\ek\\AppData\\Local\\Temp\\test_installatione9rg39ma'

    @with_rw_directory
    def test_installation(self, rw_dir):
        self.setUp_venv(rw_dir)
>       result = subprocess.run(
            [self.pip, "install", "-r", "requirements.txt"],
            stdout=subprocess.PIPE,
            cwd=self.sources,
        )

test\test_installation.py:24:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.1520.0_x64__qbz5n2kfra8p0\Lib\subprocess.py:548: in run
    with Popen(*popenargs, **kwargs) as process:
C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.1520.0_x64__qbz5n2kfra8p0\Lib\subprocess.py:1026: in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Popen: returncode: None args: ['C:\\Users\\ek\\AppData\\Local\\Temp\\test_i...>
args = 'C:\\Users\\ek\\AppData\\Local\\Temp\\test_installatione9rg39ma\\bin/pip3 install -r requirements.txt'
executable = None, preexec_fn = None, close_fds = False, pass_fds = ()
cwd = 'C:\\Users\\ek\\AppData\\Local\\Temp\\test_installatione9rg39ma\\src', env = None
startupinfo = <subprocess.STARTUPINFO object at 0x000001B4D625B390>, creationflags = 0, shell = False
p2cread = Handle(1292), p2cwrite = -1, c2pread = 15, c2pwrite = Handle(1396), errread = -1, errwrite = Handle(1416)
unused_restore_signals = True, unused_gid = None, unused_gids = None, unused_uid = None, unused_umask = -1
unused_start_new_session = False, unused_process_group = -1

    def _execute_child(self, args, executable, preexec_fn, close_fds,
                       pass_fds, cwd, env,
                       startupinfo, creationflags, shell,
                       p2cread, p2cwrite,
                       c2pread, c2pwrite,
                       errread, errwrite,
                       unused_restore_signals,
                       unused_gid, unused_gids, unused_uid,
                       unused_umask,
                       unused_start_new_session, unused_process_group):
        """Execute program (MS Windows version)"""

        assert not pass_fds, "pass_fds not supported on Windows."

        if isinstance(args, str):
            pass
        elif isinstance(args, bytes):
            if shell:
                raise TypeError('bytes args is not allowed on Windows')
            args = list2cmdline([args])
        elif isinstance(args, os.PathLike):
            if shell:
                raise TypeError('path-like args is not allowed when '
                                'shell is true')
            args = list2cmdline([args])
        else:
            args = list2cmdline(args)

        if executable is not None:
            executable = os.fsdecode(executable)

        # Process startup details
        if startupinfo is None:
            startupinfo = STARTUPINFO()
        else:
            # bpo-34044: Copy STARTUPINFO since it is modified above,
            # so the caller can reuse it multiple times.
            startupinfo = startupinfo.copy()

        use_std_handles = -1 not in (p2cread, c2pwrite, errwrite)
        if use_std_handles:
            startupinfo.dwFlags |= _winapi.STARTF_USESTDHANDLES
            startupinfo.hStdInput = p2cread
            startupinfo.hStdOutput = c2pwrite
            startupinfo.hStdError = errwrite

        attribute_list = startupinfo.lpAttributeList
        have_handle_list = bool(attribute_list and
                                "handle_list" in attribute_list and
                                attribute_list["handle_list"])

        # If we were given an handle_list or need to create one
        if have_handle_list or (use_std_handles and close_fds):
            if attribute_list is None:
                attribute_list = startupinfo.lpAttributeList = {}
            handle_list = attribute_list["handle_list"] = \
                list(attribute_list.get("handle_list", []))

            if use_std_handles:
                handle_list += [int(p2cread), int(c2pwrite), int(errwrite)]

            handle_list[:] = self._filter_handle_list(handle_list)

            if handle_list:
                if not close_fds:
                    warnings.warn("startupinfo.lpAttributeList['handle_list'] "
                                  "overriding close_fds", RuntimeWarning)

                # When using the handle_list we always request to inherit
                # handles but the only handles that will be inherited are
                # the ones in the handle_list
                close_fds = False

        if shell:
            startupinfo.dwFlags |= _winapi.STARTF_USESHOWWINDOW
            startupinfo.wShowWindow = _winapi.SW_HIDE
            if not executable:
                # gh-101283: without a fully-qualified path, before Windows
                # checks the system directories, it first looks in the
                # application directory, and also the current directory if
                # NeedCurrentDirectoryForExePathW(ExeName) is true, so try
                # to avoid executing unqualified "cmd.exe".
                comspec = os.environ.get('ComSpec')
                if not comspec:
                    system_root = os.environ.get('SystemRoot', '')
                    comspec = os.path.join(system_root, 'System32', 'cmd.exe')
                    if not os.path.isabs(comspec):
                        raise FileNotFoundError('shell not found: neither %ComSpec% nor %SystemRoot% is set')
                if os.path.isabs(comspec):
                    executable = comspec
            else:
                comspec = executable

            args = '{} /c "{}"'.format (comspec, args)

        if cwd is not None:
            cwd = os.fsdecode(cwd)

        sys.audit("subprocess.Popen", executable, args, cwd, env)

        # Start the process
        try:
>           hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                                     # no special security
                                     None, None,
                                     int(not close_fds),
                                     creationflags,
                                     env,
                                     cwd,
                                     startupinfo)
E                                    FileNotFoundError: [WinError 2] The system cannot find the file specified

C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.1520.0_x64__qbz5n2kfra8p0\Lib\subprocess.py:1538: FileNotFoundError

 test/test_installation.py ⨯                                                                             100% ██████████
=============================================== short test summary info ===============================================
FAILED test/test_installation.py::TestInstallation::test_installation - FileNotFoundError: [WinError 2] The system cannot find the file specified

Results (7.00s):
       1 failed
         - test/test_installation.py:21 TestInstallation.test_installation
     503 deselected
```

I have also verified that this occurs on Python 3.12.0rc2 on Windows. #1640 affects Windows as much as other systems, but this specific issue causes the test to fail first.

I think it would make sense to fix this together with #1640, since a fix for that would include related changes to `test_installation`. I've opened #1654 for this (which also fixes #1652 and #1653).

### LINKED ISSUE #1652: Readme claims test dependencies are installed automatically
`README.md` says:

> The list of dependencies are listed in `./requirements.txt` and `./test-requirements.txt`. The installer takes care of installing them for you.

That is not accurate, because while dependencies listed in `test-requirements.txt` are included in the list passed as `tests_require` in `setup.py`, this does not actually cause them to be installed. As `README.md` later notes, they currently have to be installed with a command like `pip install -r test-requirements.txt`. That makes the above text in the readme misleading, because this is *not* the case for `requirements.txt` dependencies, which do not have to be installed manually.

(Two related things: The CI workflows currently must install from `requirements.txt` explicitly because they do not install the project itself. The `test_installation` unit test installs from `requirements.txt` before installing the project itself, which should probably not be done, because this makes it so that test would not fail if installing GitPython failed to entail installation of its dependencies, which could arise as a regression in a change to `setup.py`.)

There are at least three distinct ways this documentation bug could be fixed:

- Remove the claim that the installer takes care of this, while making no other changes.
- Remove the claim that the installer takes care of this, and remove the `test_require` keyword argument from the `setup` call in `setup.py`.
- Use an extra instead of `tests_require` so that there really is a way to install the package such that not only are the test dependencies downloaded but they are also installed automatically in the environment in which GitPython and its regular dependencies are installed. This has the substantial disadvantage that it is somewhat unintuitive to have an extra providing dependencies for functionality that is not itself part of the package people download from PyPI. But it would have the the advantage that it would fully achieve what appears to be intended or at least aspired to and that, depending on what other changes are being made, it might make the dependency installation procedure faster and simpler.

I've opened #1654, which fixes this as well as some other issues. I took the approach of making a `test` extra, but as noted in the PR description, that can be changed, including to either of the other two approaches I've suggested here.

### LINKED ISSUE #1653: Deprecated setuptools features are used
This project currently uses deprecated features of `setuptools`, some of which may be removed in the future.

- **Installing by running `setup.py` directly**, as the instructions in `README.md` say to do and as is done in `test_installation`, is deprecated. Actual breakage due to this begins in Python 3.12, where although `setuptools` is still supported as a build backend and nontrivial logic can still be used in `setup.py`, global and virtual environments do not automatically have the `setuptools` package. (See the `ensurepip` item in the [list of removals in 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#removed).) This is the cause of #1640. But the deprecation is not new as of Python 3.12. The following message appears during installation (with any Python version provided the installed `setuptools` version is not very old, but I used Python 3.11.5):

  ```text
  /home/ek/repos-wsl/GitPython/.venv/lib/python3.11/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
  !!
  
          ********************************************************************************
          Please avoid running ``setup.py`` directly.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.
  
          See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
          ********************************************************************************
  
  !!
  ```

  This could be worked around without a fix by adding `setuptools` as a development dependency, but an actual fix is straightforward, and there are some further benefits to using `pip install .` to install locally, including that `-e` can easily be passed to make the installation editable.

- **Building by running `setup.py` directly** is, if I understand correctly, also deprecated. This is a less serious problem, but a true fix for this is also straightforward: the `build` module can be added as a development dependency and used for building the sdist and wheel in `Makefile`. The `setuptools` documentation [recommends using `build`](https://setuptools.pypa.io/en/latest/userguide/quickstart.html).

- **The `tests_require` keyword argument** is deprecated and intended for removal in a future version of `setuptools`, as noted in the [`setuptools.setup` keyword arguments documentation](https://setuptools.pypa.io/en/latest/references/keywords.html).

  This is already not achieving what appears intended or least hoped for (see #1652), and one reasonable way to fix this is to remove that argument altogether. But both the reliance on a deprecated feature and the absence of an automated way to install test dependencies as a group (without `pip install -r` commands) could be remedied by using a `test` *extra* rather than `tests_require`.

Although I worry a bit about broad scope, it seems to me that it is natural to fix #1640, #1651, #1652, and this issue (which itself overlaps with #1640 and #1652, though is not limited to them) in a single pull request. I say this because there is significant overlap in the changes, and because decisions made in review should probably take the considerations of all these issues into account. I've opened #1654 for this.

Note that while it is today discouraged to implement nontrivial logic in `setup.py`, and recommended to use `pyproject.toml` and/or `setup.cfg` for what has traditionally been done in `setup.py`, *I do not believe it is deprecated to make nontrivial use of `setup.py` as this project does*. Although it would be beneficial to define build and installation rules declaratively, I think that would be a much bigger change than needed to fix this issue. I have not attempted to do anything like that at this time.

EliahKagan:
Fixes #1640
Fixes #1651
Fixes #1652
Fixes #1653

### Overview

This changes **installation instructions, test code, scripts including `setup.py`, and CI workflows** to support Python 3.12, to fix a test that always failed on native Windows systems, and to avoid using deprecated `setuptools` features or recommending their use. I've attempted to fix problems and replace the use of deprecated features in a way that increases rather than decreasing robustness, clarity, and ease of installation. Besides the practical overlap (see below), the conceptual thread that holds all these changes together is that they are about improving compatibility with current and future Python installations.

**The actual GitPython library code itself seems already compatible with 3.12**, and this PR does not change anything in `git/`. It modifies `README.md` including to change the recommend way to install GitPython from a downloaded source code archive or cloned repository; edits scripts related to installation and building, including `setup.py`; and modifies a test to fix some OS and Python version compatibilities, to make it test the changed installation procedure in the readme, and to make it slightly more robust; and modifies the two CI workflows that run tests.

It seemed to me that the four issues mentioned above, although they are distinct issues with none being a duplicate of any others, inherently overlap and are best fixed in a way that involves overlapping code changes as well as overlapping considerations for reviewing the changes. So although I am a bit concerned about the scope of this pull request, I've done these changes together in one PR. However, they are separated across a number of narrowly scoped commits, with most of the commit messages detailing the change the commit makes and its purpose.

### Documentation changes

In parts of the readme that had to be changed to replace `python setup.py` with `pip install`, I applied other updates and clarifications too. I avoided doing this in any other parts of the readme. I reorganized the installation instructions for clarity, subdividing them into separate `<h4>` sections so that the distinctions between different installation approaches is readily apparent and readers can immediately find the part of the instructions they are looking for.

When making those changes, I included the tag-fetching step that was recently added in one place but not another where it is relevant, explicitly mentioned that the instructions should usually be carried out in a virtual environment, and addressed forks so users less familiar with common GitHub workflows would not be misled into thinking they should clone the upstream repository to propose changes. (I was a bit uncomfortable doing the latter as part of this already broad PR, but it actually relates to the tag-fetching step: if the `gh` command is used for cloning, tags on the upstream repo are available even if the fork does not have them, allowing local tests to pass.)

**I did *not* also update the documentation in `doc/`.** Although this should be done, that documentation is already considerably out of date in other ways including with respect to installation and dependencies, and this PR is already fairly large in scope. Note that the old approach of running `python setup.py install` *does* still work in all the cases where it worked before. So this does not break the old documentation, it just doesn't bring it up to date.

### CI changes

This adds 3.12, which is currently at RC2, to be tested on CI, permitting `setup-python` to install prereleases for 3.12 but not for other versions. That only affects the Ubuntu workflow.

It also updates both CI test workflows to test the new installation procedure and to harmonize them with the updated `README.md` instructions, and this attempts to make them clearer both in terms of the workflow files' own readability and in terms of the output generated in the GitHub Actions web-based interface.

**This does *not* add any new Windows tests.** While it would be valuable to have native (non-Cygwin) Windows tests on CI, and it would be valuable (if it does not cause CI checks to take too long) to have Cygwin and non-Cygwin Windows tests on multiple Python versions (as well as to test on macOS), this PR does not propose any such changes. This would be nontrivial, at least in the case of native Windows tests, and in my opinion significantly beyond the scope of this PR. I also did not want to wait to fix #1640, since Python 3.12 is already at RC2, with stable 3.12.0 coming out in a month.

Because there are still no non-Cygwin Windows tests on CI, the CI tests results shown for this pull request do not demonstrate that it resolves #1651. However, I have shown the failure in that issue, and I have verified that the changes here fix it, causing `test_installation` to pass. I tested on Windows 10 with Python 3.11.5, and also on the same system with Python 3.12.0rc2 to verify that it works for the combination of 3.12 and Windows.

### One of the `setup.py` changes deserves special scrutiny in review

The main change in this PR that I anticipate might *not* be wanted is **the addition of a `test` [extra](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-extras)**. This approach seemed best to me, but only by a very narrow margin, and I am not at all sure that I am right. The reason I did this, as well as why some other approach might be preferred, are detailed in #1652.

My rationale hinges on the assumption that it is a goal for there to be a way to install the package for local development that also takes care of installing test dependencies. Although test dependencies are not installed unless the `test` extra is called for, it may nonetheless be [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for an extra to exist that provides dependencies that none of the code in the PyPI package uses. (A possible counterargument is that running the tests is a way of using the code under test, and the code under test is part of the PyPI package.)

If the `test` extra is not wanted, I would be pleased to remove it and to update the readme and CI workflows accordingly. This could still resolve #1652, because I could appropriately weaken the claim it makes about automatic dependency installation.


init-tests-after-clone.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+
+set -e
 
 if [[ -z "$TRAVIS" ]]; then
-  read -p "This operation will destroy locally modified files. Continue ? [N/y]: " answer
+  read -rp "This operation will destroy locally modified files. Continue ? [N/y]: " answer
```
EliahKagan, line 6: I've noticed I never said in the commit message or anywhere else why I added `-r` here. It's recommended to pass `-r` to `read`, except in the rare case that `read`'s very specific backslash interpretation behavior without `-r` is desired, which it's not here. My immediate motivation for adding `-r` was to get a clean [`shellcheck`](https://www.shellcheck.net/) run as a safeguard against accidentally breaking the script while making the changes at the top. Since I believe passing `-r` is an improvement, I left it in.

setup.py
```diff
@@ -122,5 +124,6 @@ def build_py_modules(basedir: str, excludes: Sequence = ()) -> Sequence:
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
```
EliahKagan, line 127: I noticed that https://github.com/gitpython-developers/gitdb/pull/88, which listed Python 3.12 for gitdb, also removed Python 3.7 from the classifiers list and set `python_requires` to have a lower bound of 3.8, on the grounds 3.7 it is end-of-life. I'm not sure if that should be done in GitPython at this time or not, and I have not made either of those changes here, though I'd be pleased to do so on request. (All of the changes in this pull request are compatible with Python 3.7, and there is no need for GitPython and gitdb to support the same version range, since GitPython is not pinned to require the very latest version of gitdb.)

.github/workflows/pythonpackage.yml
```diff
@@ -15,51 +15,70 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
```
Byron, line 18: I agree that there is no need to claim lack of support for python 3.7 as long as we are still compatible with it. Maybe py3.7 support could be re-added to GitDB as well for clarity.

Byron (approved):
That's great work, thank again.

I find your ability to explain changes downright *super-human*, along with keeping track of their internal and external dependencies, combining it all in various orthogonal issues along with a PR to fixing all of them, explaining even why this is chosen over alternatives. This is certainly appreciated, as it forces me to follow your train of thought while catching up with python. There is still a part of me that simply wants to give you write access to `main` and say: You know what's best, go ahead :).

With that said, I would definitely appreciated if native Windows testing could, one day, be added to CI. The time budget of about 20 minutes seems quite feasible, a bound that is currently set by cygwin, and I would be surprised if native windows manages to be worse.

For me it's most important that I can still make releases, and I think I will try to perform one from this branch and if it works, merge it right away.

Makefile
```diff
@@ -16,5 +16,5 @@ release: clean
 
 force_release: clean
 	git push --tags origin main
-	python3 setup.py sdist bdist_wheel
+	python -m build --sdist --wheel
```
hugovk, line 19: By default, both are built:

```suggestion
	python -m build 
```

Byron: Thanks for the suggestion (at least *I* was unaware). In this particular case though, I like keeping it explicit, and for it to remain similar to its previous state.

Byron:
It looks like the script currently expects `python` to be present, even though for me it is `python3`.

Maybe I am holding it wrong. Should I follow the installation instructions to get a local dev environment, and then try again from there? I might add that `twine` is used to perform the release, is this something that should then be added to the venv? What's the standard for that in the python world?

Thanks for your help.

EliahKagan:
I don't think we need to add `twine` to any of the requirements files, unless you prefer that it be added, because I don't think the situation where a `twine` command local to a virtual environment would work better can arise. It uploads packages, but doesn't build them, so it doesn't require the project's other build dependencies even in principle.

Although following the instructions to get a virtual environment and running `python -m build` (or `make publish` which runs it) is one option, I think it is reasonable to modify `Makefile` to make it so that it is not necessary for you to do this. If you haven't been using a virtual environment to build GitPython, I don't think you should have to start now.

Although in virtual environments (and Docker, CI, etc.) `python` should be preferred, for scripts that are useful to run with a system Python, it is subjective what is best. **So if running `make publish` for this project is meant to be compatible with a system Python when possible, then changing `Makefile` to use `python3` instead of `python` is reasonable.** Should I make this change?

You would still need `build` to be installed globally. I don't think `pipx` can help, since `build` is not available as a command (i.e., you have to run `python -m build`, not just `build`). For a system Python on Debian or a derivative of it such as Ubuntu, there is the `python3-build` package (i.e. `sudo apt install python3-build`). If necessary you could of course install it with `pip` outside any virtual environment (e.g. `pip install --user build`).

If you're on a system that has a global `python3` command and *no* global `python` command (rather than `python` being a Python 2 interpreter), and this is a system like Debian or Ubuntu, then another option may be to install the `python-is-python3` package (i.e., `sudo apt install python-is-python3`). But whether or not that is applicable in your case, and whether or not you decide to do that, it seems to me that it is very reasonable to change `python` to `python3` in `Makefile` if that is more convenient. I definitely do not think `Makefile` is one of the places where using `python` instead of `python3` is decisively preferable.

Byron:
Thanks for elaborating.

I have changed the `Makefile` to `python3` for convenience, right after installing dependencies with `pip install -e ".[test]"`. That seems to have installed `build` for the current user (`Collecting build (from GitPython==3.1.36) Downloading build-1.0.3-py3-none-any.whl (18 kB)`), but it would still not find it when running `python3 -m build`.

In order to run `twine` on my system, I already have to set the `PATH` accordingly. Now I also set the `PYTHON_PATH` to `/Users/byron/Library/Python/3.9/lib/python/site-packages:/Users/byron/Library/Python/3.9/lib` which seemingly is where `build` was installed into, but to no avail.

Maybe a venv installation is best to solve both problems. Could you add a commit that adjusts the `README.md` and/or scripts that would make it work assuming only `virtualenv`? That at least I can execute successfully.

Thank you 😅

EliahKagan:
> Could you add a commit that adjusts the `README.md` and/or scripts that would make it work assuming only `virtualenv`?

Good idea. I have now done this in [5343aa0](https://github.com/gitpython-developers/GitPython/pull/1654/commits/5343aa01e9d90481e4570797e99faf6a98ba8f6c).

There are various approaches that could be taken for this. The approach I picked was:

- Have `make release`/`make force_release` check if a virtual environment is active, and install the `build` and `twine` packages in it if so. (The environment does not need to have the project or any dependencies installed in it, though a project virtual environment can be reused for this purpose.)
- Update the publishing instructions in `README.md` with the optional step of creating and activating a virtual environment, and to mention that doing so causes the `make release` step to automatically install `build` and `twine`.
- Remove `build` from `test-requirements.txt`, since no tests currently use it and `make` now automates its installation.

I have tested this (as best as I can) on Ubuntu with Python 3.11, in virtual environments created with `virtualenv`, as well as with `python -m venv` as I am more accustomed to using. As expected, both worked, within the limits of my testing. Besides running various commands manually, I tested `make force_release` (and pressed <kbd>Ctrl</kbd>+<kbd>C</kbd> when `twine` prompted me for credentials).

I have *kept* the change of `python` to `python3` that you made in [f86f09e](https://github.com/gitpython-developers/GitPython/pull/1654/commits/f86f09e12c5616680e19b45c940144520349c9e3). I considered having it use `python` if in a virtual environment and `python3` otherwise, but I don't think that would increase portability yet. I don't have a local macOS system to test on. But I believe--and testing seems to confirm--that the ref checking in `make release` is currently nonportable. It passes `head` to `git` where `HEAD` would be required on most other \*nix systems due to case sensitivity. It also uses `sort -n` with tag names of the form `x.y.z` (on systems I have tested this on, only the `x` and `y` parts are ordered numerically). Making `Makefile` portable is something that could be done later. Even though portability is the theme of this PR, I think it's better to defer that, provided `make release` is working at least on macOS.

The installation instructions and release instructions link to two different pages on virtual environments. This is intentional, but I'd be pleased to change if it you prefer. Because `venv` is more common than `virtualenv` and more widely recommended today, and to avoid overwhelming developers less familiar with virtual environments with choices they may not be interested in, I have retained the link to `venv`-only documentation that I previously added to the installation instructions. But in the new step of the release instructions, I linked to another (also official) page that talks both about `venv` and `virtualenv`.

Byron:
Thank you so much! The trick really was to run `python3 -m venv env` to get a virtual-env, and then installing any package actually works.

I just created a new release which, in hindsight, I was unnecessary as it doesn't contain user-facing changes.

Regarding `make release`, I haven't been using this make target in a long time as it didn't really work and I could just use `make force_release` while ideally not forgetting to create a tag. Automating this further isn't needed, I believe, but won't hurt either. In theory, a python project about git should be able to use itself to deal with tags and tag creation, I'd think 😁.

EliahKagan:
> I just created a new release which, in hindsight, I was unnecessary as it doesn't contain user-facing changes.

I'm not sure if this should've had a release or not.

Starting in 3.1.36, it is possible to run `pip install "GitPython[test]"`, which downloads and installs a GitPython wheel from PyPI, along with its test dependencies. *Users should not actually do this*, because the `test` extra is not useful to use in that particular way, but it is possible. Since the `test` extra is only useful for development, I think it makes sense that it is documented only for that, and referenced only indirectly in the release description.

When I saw d99b2d4, I thought to ask if it should be a [post](https://peps.python.org/pep-0440/#post-releases) release instead. But I realized that, due to the new extra, that should probably not be done (since it is best for post releases to carry zero installed difference).

On the other hand, it might've been best to avoid it, because it's not necessary to upgrade to it, and because the resulting downstream 3.1.36 [in conda-forge](https://github.com/conda-forge/gitpython-feedstock/commits/main/) really is identical to the corresponding downstream 3.1.35. I don't know what the best practice is in this situation.

EliahKagan:
Replying to https://github.com/gitpython-developers/GitPython/pull/1654#pullrequestreview-1619375290:

> I find your ability to explain changes downright _super-human_, along with keeping track of their internal and external dependencies, combining it all in various orthogonal issues along with a PR to fixing all of them, explaining even why this is chosen over alternatives. This is certainly appreciated, as it forces me to follow your train of thought while catching up with python.

Thanks! :) :)

> There is still a part of me that simply wants to give you write access to `main` and say: You know what's best, go ahead :).

It's an honor to hear that such a thing has even crossed your mind! However, I recommend against doing any such thing, at least at this time. The reason is that I actually have only minimal knowledge *of GitPython*. I use GitPython *indirectly*--because, as a very popular library, it is a dependency of other projects I use directly, especially [nbdime](https://github.com/jupyter/nbdime/blob/43168fce57ed8f207449d5e4cc2b41e0c5e213f2/setup.py#L129). As a result, if I did have write access, it would be the current situation, just with extra steps: I would not be comfortable making any unreviewed changes.

That's the origin of my interest in GitPython. I learned of CVE-2023-40590 (#1635) from security alerts on a couple of my repositories that declare `nbdime` as a development dependency, so I decided to contribute #1636. Unfortunately, even once amended in #1650, my fix also introduces [an `os.environ` race condition](https://github.com/gitpython-developers/GitPython/pull/1650#issuecomment-1710089136), so maybe it [could be done better](https://github.com/gitpython-developers/GitPython/pull/1650#issuecomment-1710139306). Given that the still-flawed fix (like the CVE it fixes) [affects native Windows only](https://github.com/gitpython-developers/GitPython/pull/1657#issuecomment-1716870800), I would feel better experimenting with such improvements once we have...

> With that said, I would definitely appreciated if native Windows testing could, one day, be added to CI. The time budget of about 20 minutes seems quite feasible, a bound that is currently set by cygwin, and I would be surprised if native windows manages to be worse.

Yes, this is something I am very much interested in doing. (Though I certainly do not want to stop anybody else who wants to work on it!)

I'm interested in figuring out *(a)* which, of tests that do not pass for me on Windows, *should* pass because they fail due to a problem specific to my environment or fork, and, *(b)* of others that are not already skipped on Windows, if the bugs that break them on native Windows `python` can simply be fixed, #1651 is an example of such a bug. I'd rather find and fix more bugs like that--if there are any--than obscure them to get CI to pass.

I'm also hoping to find some way to make it significantly faster than the Cygwin test job (currently) is, even if not nearly as fast as the Ubuntu test jobs. If the tests can be made to run fast enough on native Windows, then a matrix strategy could be used as in the Ubuntu test workflow to test on all Python versions GitPython supports (possibly even as part of that workflow, by adding an OS dimension to the existing matrix, depending on how much Window-specific preparation ends up having to be done). Because each job gets its own runner and the runners operate in parallel, six 20-minute checks are no problem for one push, but could slow things down a lot with many pushes in a short time, due to limits on the number of GitHub Actions runners per user or organization.

If worse comes to worse, there is always the middle ground of testing a smaller number of versions on native Windows but still more than one. Perhaps 3.8 and 3.12.

hugovk:
> `make release` uses `git tag | sort -nr`. Do you happen to know if, on macOS, that sorts according to SemVer, at least for the tag names that actually exist in this repository?

macOS Ventura doesn't sort according to SemVer:

```console
❯ sort --version
2.3-Apple (154)
❯ git tag | sort -nr
3.1.9
3.1.8
3.1.7
3.1.6
3.1.5
3.1.4
3.1.36
3.1.35
3.1.34
3.1.33
3.1.32
3.1.31
3.1.30
3.1.3
3.1.29
3.1.28
3.1.27
3.1.26
3.1.25
3.1.24
3.1.23
3.1.22
3.1.20
3.1.2
3.1.19
3.1.18
3.1.17
3.1.16
3.1.15
3.1.14
3.1.13
3.1.12
3.1.11
3.1.10
3.1.1
3.1.0
3.0.9
3.0.8
3.0.7
3.0.6
3.0.5
3.0.4
3.0.3
3.0.2
3.0.1
3.0.0
2.1.9
2.1.8
2.1.7
2.1.6
2.1.5
2.1.4
2.1.3
2.1.2
2.1.15
2.1.14
2.1.13
2.1.12
2.1.11
2.1.10
2.1.1
2.1.0
2.0.9
2.0.8
2.0.7
2.0.6
2.0.5
2.0.4
2.0.3
2.0.2
2.0.1
2.0.0
1.0.2
1.0.1
1.0.0
0.3.7
0.3.6
0.3.5
0.3.4
0.3.3
0.3.2.1-patched
0.3.2.1
0.3.2-RC1
0.3.2
0.3.1-beta2
0.3.1-beta1
0.3.0-beta2
0.3.0-beta1
0.2.0-beta1
0.1.7
0.1.6
0.1.5
0.1.4-pre
0.1.4
winerr_show
```

Byron:
Thanks for sharing your "origin story" - I start liking CVEs much more now, given they made you aware of this project and caused the first contribution :).
I understand your position regarding becoming a maintainer, and if there is ever a change to that so there is benefit to make you one, just let me know :).

> I'd rather find and fix more bugs like that--if there are any--than obscure them to get CI to pass.

If I recall correctly, most test-failures on windows arise from its 'unique' way of handling open files, which is quite incompatible with the way files are handled on Unix systems. That said, by now things might have changed and it's probably worth to at least retry the ignored tests to see if some of them work now.

I also think that having any kind of native windows CI job is orthogonal to making more tests pass - the latter is entirely optional to me and definitely nothing more than 'nice to have' at this point. CI on native python on Windows would be a great step forward already.

> [..] but could slow things down a lot with many pushes in a short time, due to limits on the number of GitHub Actions runners per user or organization.

These days there isn't too much CI activity here, but if it was a problem I am willing to cancel some jobs by hand. It's something I do quite regularly in `gitoxide` where CI takes 30 to 40 minutes.

EliahKagan:
@hugovk Thanks--I find that very helpful, because it tells me `sort -n` doesn't work for this on any common systems and should be abandoned.

Based on that, I looked for alternatives to `sort -n` and `sort -V` that better support SemVer, and found [`versionsort.suffix`](https://www.git-scm.com/docs/git-config/#Documentation/git-config.txt-versionsortsuffix), which will let `git tag --sort=-v:refname` do it well enough:

```sh
git -c versionsort.suffix=-alpha -c versionsort.suffix=-beta -c versionsort.suffix=-pre -c versionsort.suffix=-rc -c versionsort.suffix=-RC tag --sort=-v:refname
```

If git's `-c` supported a form that allowed it and its operand to be passed as a single command-line argument, then brace expansion (in a shell that has it) could abbreviate that nicely. Unfortunately it doesn't. But:

```sh
git $(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC) tag --sort=-v:refname
```

Then, to also filter out non-version tags (they're at the top now, so must be dropped before `| head -n1`):

```sh
git $(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC) tag -l '[0-9]*' --sort=-v:refname
```

(Where the `[0-9]*` is a glob, not a regex, and matches the entire tag name, so it's like `grep`ing for `^[0-9]`.)

EliahKagan:
@Byron

> I understand your position regarding becoming a maintainer, and if there is ever a change to that so there is benefit to make you one, just let me know :).

Thanks! That might make sense if/when I have learned more about the codebase (though I would still likely request reviews on many changes).

> I also think that having any kind of native windows CI job is orthogonal to making more tests pass - the latter is entirely optional to me and definitely nothing more than 'nice to have' at this point. CI on native python on Windows would be a great step forward already.

That makes sense. If it's okay for native Windows CI to be set up even with tests failing that are not currently skipped or documented to fail on Windows, then I may be able to do it sooner. I think there's still some stuff I need to figure out, though. For example, none of the changes I've made that needed to be tested on Windows required me to set up `git-daemon`, and I have never done so. Since I believe Windows GitHub Actions runners include full [Git for Windows](https://gitforwindows.org/) installations, and Git for Windows does ship `git-daemon.exe` in its `/mingw64/libexec/git-core` directory (it is a stripped down MinGW system), any insights from figuring that out locally should carry over. I think some of the tests that fail on my Windows system are failing only because not everything is set up properly.

Byron:
> git $(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC) tag -l '[0-9]*' --sort=-v:refname

This is absolutely amazing! I know `git` can do a lot, but wasn't aware of this configuration option.

> That makes sense. If it's okay for native Windows CI to be set up even with tests failing that are not currently skipped or documented to fail on Windows, then I may be able to do it sooner.

Yes, absolutely. I think it's better to split the task into CI Setup and Tests themselves. That way, Tests can simply be ignored when found failing at first until the is time to see if they could also work. Additionally, I think it's perfectly alright to only test the borders of the supported version range of Python, and for that test different Windows versions. I know that GHA supports quite some parallelism, but I'd prefer to conserve the compute anyway and optimize for actual coverage. Limiting oneself somewhat when setting up additional jobs seems particularly easy here since this GitPython seems to have been fine without any of it for more than a decade now.

PR #1655, opened 2023-09-11
Update setup.py

owenwu811:


Byron:
Thanks for contributing, even though I don't quite understand what the purpose of this PR is.

The reason I am closing it is that I just merged a huge PR that affects setup.py. Please review what's there now and chances are this change here won't be needed anymore. If that's not the case, please open a new PR with the changes you need, while not hesitating to spell out what it does, for dummies, if you will :). Thanks.

PR #1657, opened 2023-09-12
Better document env_case test/fixture and cwd

EliahKagan:
This improves how some of the changes in #1650 are documented, expanding comments and a docstring.

I ended up doing this in four commits, but really it would probably be just as clear with two. If you'd like me to rebase this to have just two commits, please let me know.

### Identifying the steps in the "`env_case`" test

This is my main motivation for opening this PR, and most of the change.

I was thinking about how it was not immediately obvious why `test_it_avoids_upcasing_unrelated_environment_variable_names` (the "`env_case`" test) works and why it is done in this way.

To improve on this, I wrote comments that divide it into four steps. Two of the steps occur in the test case method, and the others occur in the supporting `env_case.py` fixture. The comments are in both files.

This only adds and changes comments, but just in case I have locally tested that it does not stop the test from passing, nor from failing when using the version of the code in `git/` from just before the merge of #1650.

### Non-reentrancy of some context manager objects

I had intended my brief mention of non-reentrancy to distinguish `git.util.cwd` from [`contextlib.chdir`](https://docs.python.org/3/library/contextlib.html#contextlib.chdir), which is fully reentrant by having each returned object maintain a stack of directories (but which is not available before Python 3.11). But this was not clear at all--I had nowhere mentioned `contextlib.chdir`!--and it risked giving the wrong impression that `cwd` couldn't be called multiple times in nested `with`-statements.

Furthermore, the newly introduced `patch_env` function is non-reentrant in exactly the same narrow sense, but because it did not mention this and `cwd` did, this gave the wrong impression that `patch_env` was reentrant in this way. (I think `patch_env` actually *doesn't* need to document this--really `cwd`'s limitation compared to `contextlib.chdir` is the only thing important enough to mention.)

I've somewhat clarified this, but maybe further improvements can be made in the future. This can be make clearer with [extended example](https://gist.github.com/EliahKagan/eb0c0d34b0b5e911f52cd044f8fac320#file-transcript-txt), but I would be very reluctant to bloat the `cwd` docstring with fully worked examples. If anything, maybe there is a way to say it shorter. Or maybe the reentrancy information should be removed altogether.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -150,7 +150,10 @@ def unbare_repo(func: Callable[..., T]) -> Callable[..., T]:
 
 @contextlib.contextmanager
 def cwd(new_dir: PathLike) -> Generator[PathLike, None, None]:
-    """Context manager to temporarily change directory. Not reentrant."""
+    """Context manager to temporarily change directory.
+
+    This is similar to contextlib.chdir introduced in Python 3.11, but the context
+    manager object returned by a single call to this function is not reentrant."""
     old_dir = os.getcwd()
     os.chdir(new_dir)
     try:
--- a/test/fixtures/env_case.py
+++ b/test/fixtures/env_case.py
@@ -1,13 +1,18 @@
+# Steps 3 and 4 for test_it_avoids_upcasing_unrelated_environment_variable_names.
+
 import subprocess
 import sys
 
+# Step 3a: Import the module, in case that upcases the environment variable name.
 import git
 
 
 _, working_dir, env_var_name = sys.argv
 
-# Importing git should be enough, but this really makes sure Git.execute is called.
+# Step 3b: Use Git.execute explicitly, in case that upcases the environment variable.
+#          (Importing git should be enough, but this ensures Git.execute is called.)
 repo = git.Repo(working_dir)  # Hold the reference.
 git.Git(repo.working_dir).execute(["git", "version"])
 
+# Step 4: Create the non-Python grandchild that accesses the variable case-sensitively.
 print(subprocess.check_output(["set", env_var_name], shell=True, text=True))
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -98,15 +98,25 @@ class TestGit(TestBase):
         old_name = "28f425ca_d5d8_4257_b013_8d63166c8158"
         if old_name == old_name.upper():
             raise RuntimeError("test bug or strange locale: old_name invariant under upcasing")
-        os.putenv(old_name, "1")  # It has to be done this lower-level way to set it lower-case.
 
+        # Step 1: Set the environment variable in this parent process. Because os.putenv is a thin
+        #         wrapper around a system API, os.environ never sees the variable in this parent
+        #         process, so the name is not upcased even on Windows.
+        os.putenv(old_name, "1")
+
+        # Step 2: Create the child process that inherits the environment variable. It will see it
+        #         in os.environ with an upcased name, but if it is not mutated through os.environ
+        #         then it will pass it on to its own child processes with the original name. The
+        #         child process will use GitPython, and we are testing that it passes the variable
+        #         with the exact original name to its own child processes.
         cmdline = [
             sys.executable,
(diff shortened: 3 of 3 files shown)


Byron:
Thanks again for your tremendous help!

I appreciate these improvements and confirm that they make clearer what's happening. It's a bit sad that we don't have a general-purpose bug-tracker yet that we can refer to in a self-contained fashion (maybe similar to [`git-appraise`](https://github.com/google/git-appraise)), as I also find myself to be hesitant to link to GitHub issues, as doing so might also be beneficial here.

> I ended up doing this in four commits, but really it would probably be just as clear with two. If you'd like me to rebase this to have just two commits, please let me know.

I'd never ask that as I believe that any `git`-commit workflow has its purpose, with different trade-offs. One day I hope there is tooling that makes leveraging this information easier than `git log` or `git blame`, but [we will be getting there](https://github.com/Byron/gitoxide).

> This only adds and changes comments, but just in case I have locally tested that it does not stop the test from passing, nor from failing when using the version of the code in `git/` from just before the merge of #1650.

That's extremely diligent of you, but even if there would have been a chance for test-failure, it's fine if CI detects it first. It's a tool and ideally, it saves some time as well.

EliahKagan:
> That's extremely diligent of you, but even if there would have been a chance for test-failure, it's fine if CI detects it first. It's a tool and ideally, it saves some time as well.

I'm a big fan of discovering test failures through CI only. The issue, here, is that the `env_case` test is for a regression that only affects *native* Windows, as elaborated below. So current CI, which uses a Cygwin build of `python`, would not catch a problem. Otherwise I would not have separately tested manually. However, given that the changes to the `env_case` code were just comments, it still probably wasn't necessary for me to test.

---

#1635 (CVE-2023-40590) and #1646 only affected GitPython on native Windows builds of `python`.

Although the reported case of #1646 mattered because of how it broke Cygwin `make`, I believe the `python` build used must have been native even there, because #1636 only patched `os.environ` when `is_win` was true, a behavior #1650 carries forward in its more limited patching. When I wrote the description of #1650, I was not sure about `git.compat.is_win` on Cygwin `python` builds. But I have since verified that it is `False`, which makes sense since it's based on `os.name`, which is `posix` rather than `nt` on Cygwin builds.

When I found that out, I was *briefly* worried CVE-2023-40590 might not be fully patched. In #1636, I had committed the fix and then the test, and mistakenly assumed--based on my local verification, which was with native Windows, not Cygwin--that CI would have shown a test failure if I had committed them in the other order. The regression test introduced in #1636 (which is now verifying the modified fix as of #1650) covers both true and false cases of `is_win`, but it would nonetheless probably not detect the bug with a Cygwin `python` build even if it were present. This is because, if the current directory were automatically included in a `subprocess.Popen` path search in Cygwin Python, it would be an inherited behavior from the Windows API and presumably would only be triggered by `.exe` files. (See https://github.com/python/cpython/issues/91558#issuecomment-1100942950.) But we only test with an `.exe` file when `is_win` is true.

So if GitPython on Cygwin `python` builds had suffered from CVE-2023-40590, it still would. Fortunately, I am pretty sure it does not. I have manually tested the `subprocess` module on a Cygwin build of `python`, both with and without `shell=True` (though with `shell=True` on a Cygwin build, the shell is Cygwin's `/bin/sh`, so I was pretty sure it wouldn't happen that way). Based on this, the Windows behavior that gave rise to CVE-2023-40590 does not happen in Cygwin `python` builds. I've also tested this in a MSYS2 build of `python`--MSYS2 being a major derivative of Cygwin--with the same result.

Note that CVE-2023-40590 *did* affect MinGW `python` builds [installed by MSYS2 `pacman`](https://www.msys2.org/docs/package-naming/) (such as MINGW64 and UCRT64 builds), because those are actually native Windows builds, but for the same reason `os.name` is `nt` on those, so `git.compat.is_win` is `True`, so #1636 and #1650 patch `NoDefaultCurrentDirectoryInExePath` in as intended.

PR #1658, opened 2023-09-12
Remove spurious executable permissions

EliahKagan:
This unsets the executable bit on `test/fixtures/diff_mode_only`, which was the only of the fixture files to be marked executable, and which is not a script.

This is a minor change, but it didn't really fit in any other PRs I was opening, so I made it its own. The benefit of this change is that it useful to be able to run `ls -l` to get a sense of what all the files are in a repository that make sense to execute as scripts. This provides useful information in the case of `init-tests-after-clone.sh` and `setup.py`, but not in the case of `test/fixtures/diff_mode_only`,

`test/fixtures/diff_mode_only` seems to have gotten marked as executable during one of [several moves](https://github.com/gitpython-developers/GitPython/commits/main/test/fixtures/diff_mode_only). This could happen due to being on a volume that doesn't support Unix-style executable permissions, accessed on a system that otherwise does, depending on how that volume is mounted. (For example, I see this on Windows when accessing the native Windows filesystem from WSL, and on Ubuntu when using an external drive formatted NTFS.) I don't think the `+x` on that file was ever intended, and removing it doesn't break any tests.

In some ways of viewing a permissions-only change, a PR's review tab wrongly claims it is to an "empty file." That is not the case here, even if the GitHub UI claims it: the file is nonempty, I'm just not changing any of its content.


Byron:
That's a great catch, thanks for fixing it!

> In some ways of viewing a permissions-only change, a PR's review tab wrongly claims it is to an "empty file." That is not the case here, even if the GitHub UI claims it: the file is nonempty, I'm just not changing any of its content.

That's quite a misleading bug and a surprise GitHub still has such blatant ones! Maybe they check for changes and if there is none, assume the file is empty. All speculation though, it's not that we could go in and fix it for them :/.

PR #1661, opened 2023-09-13
Fix up checks in Makefile and make them portable

### LINKED ISSUE #1660: "make [force_]release" attempts upload even if build fails
The `force_release` target in `Makefile` (which is also used by the `release` target after it attempts to check some things separate from this issue) had previously stopped after a failed attempt to build. But since #1654 it runs `twine` to attempt the upload whether or not the build succeeded.

The problem is that a command of the form `a || b` succeeds when `a` succeeds, but also when `a` fails but `b` succeeds:

https://github.com/gitpython-developers/GitPython/blob/25597cbc05c7e478efc77bacd0f441962dbf87d5/Makefile#L20

So that command fails only in the rare case that `echo` fails to write to standard output.

But the message about how using a virtual environment may remedy a build failure is valuable and should not be removed. Instead, the logic can be adjusted to make sure the command fails if it gets that far.

There are some other improvements I'd like to propose to `Makefile`, and I think it will be convenient to include a fix for this with them. But I wanted to open an issue for this, because I think this behavior is unexpected and could cause confusion while making a release if it is not anticipated (if the next release is made before it is fixed and something keeps the build command from succeeding).

EliahKagan:
Fixes #1660

In `.gitattributes`:

- Fix checking out `init-tests-after-clone.sh` with LF line endings on all systems (this was broken).
- Add `Makefile` to so it is also checked out with LF line endings on all systems.\
  (Windows ports of `make` will usually tolerate CRLF line endings, but `\r` or other representations of CR may, and often do, appear in output, giving an appearance of a problem that is hard to dismiss. So it's worth keeping it LF.)

For the `force_release` target:

- Don't try to upload if building failed (#1660).
- Only suggest using a virtual environment to solve a failure that occurred outside of one.
- In a virtual environment, use `python`, not `python3`, for compatibility with Windows.\
  (Windows may have a `python3` command installed outside the virtual environment, which is the most confusing case, since it gets used even from inside, so it feels like installing packages in the venv had no effect. In contrast, some Debian and derived systems, and some macOS systems, have no `python` command *outside* a venv, or it's Python 2. So keep using `python3` when no venv is detected.)

For the `release` target:

- Reference the current commit as `HEAD` rather than `head`. `HEAD` is more portable, because `git` does not treat names case-insensitively by default on all systems. `head` works on Windows and I think macOS, but not GNU/Linux or most other Unix-like systems.
- Use `git tag --sort=-v:refname` with an appropriate [`versionsort.suffix`](https://www.git-scm.com/docs/git-config/#Documentation/git-config.txt-versionsortsuffix) to [**order the version tags in the specific SemVer-inspired way this project uses**](https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717237180), and a glob to omit the non-version tags. This avoids [the correctness problem](https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717014696) with the `| sort -nr` approach, is portable, and is more precise for this project's versions than even [nonportable `sort -V` tweaked for SemVer](https://gist.github.com/loisaidasam/b1e6879f3deb495c22cc?permalink_comment_id=4070026#gistcomment-4070026), because this project has SemVer suffixes like `-pre` where `N-pre` sorts before `N`, but also non-SemVer `-patched` where `N-patched` sorts after `N`.\
  (Thanks to @hugovk for [investigating what `sort -nr` was doing on macOS](https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717014696)!)
- Decrease duplication of complex `git` commands and pipelines.
- Write the checks to fail if a command in `$(` `)` fails (by extracting them to assignments), so that if a `git` command whose output is needed fails, this doesn't cause a check to wrongly succeed.
- Check for `VERSION` and `changes.rst`, against each other and also the latest tag name.
- Check for uncommitted changes.
- Do the version/tag/reference agreement check after printing out a table showing all the relevant values, so the specific problems can always be efficiently identified.

The above includes checks for indications that each of the actions listed in the `README.md` release instructions was done, and is intended to make it so `make release` is correct, fully usable, and preferable to `make force_release` in most situations.

I have tested this on Ubuntu and Windows. When testing, I pressed <kbd>Ctrl</kbd>+<kbd>C</kbd> when `twine` prompted for credentials. I tested both the `release` and `force_release` targets. Some of the manual tests of the `release` target on Ubuntu [can be seen in this original gist](https://gist.github.com/EliahKagan/e080b25996b97cb9561e3c74a664c8da) from [before that review](https://github.com/gitpython-developers/GitPython/pull/1661#pullrequestreview-1626077838), and [**this newer gist testing `build-release.sh`**](https://gist.github.com/EliahKagan/25ac58dd7ed5b952c20529acea1a2fe1). Although not shown in those gists, I have also tested that the Makefile targets use the scripts correctly since extracting commands to them.

Here's an example of output from `make release` showing various checks and the new version/tag/ref table and failing with an error due to an inconsistency that should block the release:

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (portable-makefile=)$ make release
rm -rf build/ dist/ .eggs/ .tox/
./check-version.sh
Checking current directory.
Checking that VERSION and doc/source/changes.rst exist and have no uncommitted changes.
Checking that ALL changes are committed.
Gathering current version, latest tag, and current HEAD commit info.
VERSION file   = 3.1.36
changes.rst    = 3.1.36
Latest tag     = 3.1.36
HEAD SHA       = 4b7d7028193d3a03cc20867fa534e21528886f51
Latest tag SHA = 5297f6210f185c9df9c97392ffef4da8562b58ff
./check-version.sh: Check failed. Stopping.
make: *** [Makefile:10: release] Error 1
```

Here's an example of a failure where it doesn't get anywhere near that far:

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (portable-makefile +=)$ make release
rm -rf build/ dist/ .eggs/ .tox/
./check-version.sh
Checking current directory.
Checking that VERSION and doc/source/changes.rst exist and have no uncommitted changes.
M  VERSION
./check-version.sh: Check failed. Stopping.
make: *** [Makefile:10: release] Error 1
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,3 @@
 test/fixtures/* eol=lf
-init-tests-after-clone.sh eol=lf
-Makefile eol=lf
-check-version.sh eol=lf
-build-release.sh eol=lf
+*.sh eol=lf
+/Makefile eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 test/fixtures/* eol=lf
 init-tests-after-clone.sh eol=lf
-check-version.sh eol=lf
 Makefile eol=lf
+check-version.sh eol=lf
+build-release.sh eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 test/fixtures/* eol=lf
 init-tests-after-clone.sh eol=lf
+check-version.sh eol=lf
 Makefile eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 test/fixtures/* eol=lf
-init-tests-after-clone.sh
+init-tests-after-clone.sh eol=lf
+Makefile eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.py[co]
 *.swp
 *~
+.env/
+env/
 .venv/
 venv/
 /*.egg-info
--- a/Makefile
+++ b/Makefile
@@ -7,13 +7,29 @@ clean:
 	rm -rf build/ dist/ .eggs/ .tox/
 
 release: clean
-	# Check if latest tag is the current head we're releasing
-	@config_opts="$$(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC)" && \
+	# Check that VERSION and changes.rst exist and have no uncommitted changes
+	test -f VERSION
+	test -f doc/source/changes.rst
+	git status -s VERSION doc/source/changes.rst
+	@test -z "$$(git status -s VERSION doc/source/changes.rst)"
+
+	# Check that ALL changes are commited (can comment out if absolutely necessary)
+	git status -s
+	@test -z "$$(git status -s)"
+
+	# Check that latest tag matches version and is the current head we're releasing
+	@version_file="$$(cat VERSION)" && \
+	changes_file="$$(awk '/^[0-9]/ {print $$0; exit}' doc/source/changes.rst)" && \
+	config_opts="$$(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC)" && \
 	latest_tag=$$(git $$config_opts tag -l '[0-9]*' --sort=-v:refname | head -n1) && \
 	head_sha=$$(git rev-parse HEAD) latest_tag_sha=$$(git rev-parse "$$latest_tag") && \
-	printf '%-14s = %s\n' 'Latest tag'     "$$latest_tag" \
+	printf '%-14s = %s\n' 'VERSION file'   "$$version_file" \
+	                      'changes.rst'    "$$changes_file" \
+	                      'Latest tag'     "$$latest_tag" \
 	                      'HEAD SHA'       "$$head_sha" \
 	                      'Latest tag SHA' "$$latest_tag_sha" && \
+	test "$$version_file" = "$$changes_file" && \
+	test "$$latest_tag" = "$$version_file" && \
 	test "$$head_sha" = "$$latest_tag_sha"
 
 	make force_release
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,13 @@ release: clean
 force_release: clean
 	# IF we're in a virtual environment, add build tools
 	test -z "$$VIRTUAL_ENV" || pip install -U build twine
-	python3 -m build --sdist --wheel || echo "Use a virtual-env with 'python -m venv env && source env/bin/activate' instead"
+
+	# Build the sdist and wheel that will be uploaded to PyPI.
+	python3 -m build --sdist --wheel || \
+		test -z "$$VIRTUAL_ENV" && \
+		echo "Use a virtual-env with 'python -m venv env && source env/bin/activate' instead" && \
+		false
+
+	# Upload to PyPI and push the tag.
 	twine upload dist/*
 	git push --tags origin main
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,14 @@ clean:
 
 release: clean
 	# Check if latest tag is the current head we're releasing
-	echo "Latest tag = $$(git tag -l '[0-9]*' --sort=-v:refname | head -n1)"
-	echo "HEAD SHA       = $$(git rev-parse HEAD)"
-	echo "Latest tag SHA = $$(git tag -l '[0-9]*' --sort=-v:refname | head -n1 | xargs git rev-parse)"
-	@test "$$(git rev-parse HEAD)" = "$$(git tag -l '[0-9]*' --sort=-v:refname | head -n1 | xargs git rev-parse)"
+	@config_opts="$$(printf ' -c versionsort.suffix=-%s' alpha beta pre rc RC)" && \
+	latest_tag=$$(git $$config_opts tag -l '[0-9]*' --sort=-v:refname | head -n1) && \
+	head_sha=$$(git rev-parse HEAD) latest_tag_sha=$$(git rev-parse "$$latest_tag") && \
+	printf '%-14s = %s\n' 'Latest tag'     "$$latest_tag" \
+	                      'HEAD SHA'       "$$head_sha" \
+	                      'Latest tag SHA' "$$latest_tag_sha" && \
+	test "$$head_sha" = "$$latest_tag_sha"
+
 	make force_release
 
 force_release: clean
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,10 @@ clean:
 
 release: clean
 	# Check if latest tag is the current head we're releasing
-	echo "Latest tag = $$(git tag | sort -nr | head -n1)"
+	echo "Latest tag = $$(git tag -l '[0-9]*' --sort=-v:refname | head -n1)"
 	echo "HEAD SHA       = $$(git rev-parse HEAD)"
-	echo "Latest tag SHA = $$(git tag | sort -nr | head -n1 | xargs git rev-parse)"
-	@test "$$(git rev-parse HEAD)" = "$$(git tag | sort -nr | head -n1 | xargs git rev-parse)"
+	echo "Latest tag SHA = $$(git tag -l '[0-9]*' --sort=-v:refname | head -n1 | xargs git rev-parse)"
+	@test "$$(git rev-parse HEAD)" = "$$(git tag -l '[0-9]*' --sort=-v:refname | head -n1 | xargs git rev-parse)"
 	make force_release
 
 force_release: clean
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ clean:
 release: clean
 	# Check if latest tag is the current head we're releasing
 	echo "Latest tag = $$(git tag | sort -nr | head -n1)"
-	echo "HEAD SHA       = $$(git rev-parse head)"
+	echo "HEAD SHA       = $$(git rev-parse HEAD)"
 	echo "Latest tag SHA = $$(git tag | sort -nr | head -n1 | xargs git rev-parse)"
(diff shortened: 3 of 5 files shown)


Byron (changes requested):
Thanks a lot! I didn't even read all explanation yet as I have a single change request before doing so: Could you place the script section of the `release` section into an actual shell script? If you see fit, it *could* take an argument that allows to force the release.

With that, it will be way easier to produce a script that is free of unnecessary noise. It's my recommendation to use `set -eu -o pipefail`  to be able to get rid of `&&` as well.

Thank you.

hugovk:
As a follow up, I recommend setting up automated deploys to PyPI from GitHub Actions using the new "Trusted Publishers", to increase security and make releasing easier:

* https://github.com/pypa/gh-action-pypi-publish
* https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

EliahKagan:
@Byron Good idea. That has a number of advantages, including that shell scripts in proper files can be checked for bugs using static analyzers. I have now made such a change, or something like it--this may not be quite what you want and it can definitely be changed further.

I extracted everything in the `release` target except `make force_release` to a shell script `check-version.sh`. In addition, although you did not specifically request this, I extracted the code in `force_release` target that from before the `twine` command to *separate* script `build-release.sh`. For now, this maintains the original separation between the `release` and `force_release` targets, so I have not yet at this point done anything like having either script accept a "force" option, which I think is something that would make more sense to do if the `Makefile` as a whole is replaced by a shell script.

In both scripts, I have gotten rid of most `&&` and `\` in preference for `set` with `-e` and some other options.

Because `make` is no longer echoing commands, I added messages to explain what is going on. I considered using `set -x` but I believe that would be excessively verbose, and while it could be turned on and off at different points in the script, that would make things more complicated if one later wishes to use it for debugging (or just to get super-verbose output for some other reason).

Because not all failures--especially with `test` commands--produce output that makes it clear there was an error, I used `trap` to append a message when `check-version.sh` fails for any reason. While I was at it, I used `trap` to append the virtual-env recommendation in `build-release.sh` when the build fails outside one; this is one way to preserve the failing command's original exit code, which the old `|| false` way did not do.

Maybe the scripts should be moved and/or renamed. I put them in the root of the repository, but it may be best to reduce clutter there. I named them with a `.sh` suffix because `init-tests-after-clone.sh` has one, but it may be better to omit the suffix (which traditionally denotes files that are meant to be sourced rather than executed, whereas these scripts should not be sourced).

Byron:
> As a follow up, I recommend setting up automated deploys to PyPI from GitHub Actions using the new "Trusted Publishers", to increase security and make releasing easier:
> 
> * https://github.com/pypa/gh-action-pypi-publish
> * https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

@hugovk Thanks for the hint! I imagine that this could be a viable next step. The workflow could be that the release is prepared to the point where `check-version.sh` passes. This implies a correctly named tag which I would sign as usual. Then, on push, said tag, CI could make a release automatically.

When thinking about it, at this point most of the work was already done and running `make release` locally isn't a problem, with the only 'inconvenience' that I have to type `__token__` and then paste a passcode. These few seconds could be saved at the expense of complexity, abstraction, and giving some secrets to the cloud. And I think this is a price too high overall for the little time saved.

However, I truly think this is the way forward once multiple people should be able to make releases without entrusting them with write access to the python packages on pypi.

EliahKagan:
> Something you might find interesting is the `$'\n works here'` syntax in `bash` that helps with inserting newlines into `echo` at will.

In #1693 I took this further in what I would regard to be that direction, using two `echo` commands since the extra newline is at the edge:

```diff
-echo $'\nThe VERSION must be the same in all locations, and so must the HEAD and tag SHA'
+echo
+echo 'The VERSION must be the same in all locations, and so must the HEAD and tag SHA'
```

However, you might prefer the way with `$'` `'` and `\n` for compactness. So I understand you may change it back or ask me to do so!

Another way to write this, which personally I do *not* think is as nice as either `$'` `'` or using two separate `echo` commands, is:

```bash
echo -e '\nThe VERSION must...'
```

The reason I think the other approaches (including `$'` `'` as you used here) are nicer is that the versatility of `echo -e` is more powerful than often intended. For example, this expands escape sequences in the value of `var`, which if untrusted, could cause problems, including giving strange instructions to some terminals that obscure output the user may rely on:

```bash
echo -e "2 loaves of bread\n3 eggs\n$var juice\na sponge"
```

In contrast, `$"` `"` exists, but it is [unrelated](https://www.gnu.org/software/bash/manual/bash.html#Locale-Translation) to expanding escape sequences, so if one accidentally uses it, one finds out when the code is exercised.

POSIX permits `echo` to always, never, or sometimes, behave as though `echo -e` were used, and some `echo`s have `-e` as the default behavior, sometimes allowing `-E` to turn it off. Actually, what POSIX actually permits is either behavior, or an arbitrary idiosyncratic combination of them. POSIX actually forbids `-e` and `-E` from being treated in these ways--they are supposed to be printed literally--but implementations widely treat them as options anyway. `-n` likewise may or may not be supported, though POSIX permits it, while also permitting it to be treated literally (or to have [any other meaning](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html)).

For these and other reasons, `echo` is nonportable, but since [`echo` in `bash` is a builtin](https://www.gnu.org/software/bash/manual/bash.html#index-echo), it is okay to use it in `bash` (so long as one does not write things like `... | xargs echo` or `find ... -exec echo ...`, which use an external `echo` executable that may behave differently). It is for this reason that, in #1693, I've used `printf` in the init script, which as part of my proposed changes there is converted to be a POSIX shell script to support more systems. But this is inapplicable to the other scripts, which use `bash`'s `echo` builtin without problems.

> I also thought that maybe there would be a better place for them in _some_ directory, but then I could also appreciate that some automation is so easily discoverable. It's probably OK to leave them as is for now.

I think we could document the `Makefile` in the readme and this would give sufficient discoverability, if in the future it is found useful to move them. But I have not included any such proposed change in #1693.

> As for the `.sh` suffix, I am among the people who take the executable bit as indication of whether they are to be sourced or executed, so (at least to me) the intent is communicated clear enough.

This actually works remarkably better even than one might expect. The permissions system in Windows is very different from Unix-like systems, and nothing that really corresponds to Unix-style executable bits exists. But when one uses a Unix-style shell on Windows, such as the Git Bash shell, the available Unix-style utilities like `ls` handle this by guessing the executable bit from the hashbang. When one is not using a Unix-style shell, one typically cannot easily run a shell script, and one *definitely* cannot source one since it's the wrong language for the shell being used. So things tend to work out.

(This is separate from WSL: a system in WSL fully supports Unix-style permissions, though like any such operating system, it will not support them when using a Windows filesystem. Windows filesystems being used in a Unix-like system are typically--both when accessing files outside WSL from inside it, and when mounting an NTFS drive on an independent Unix-like operating system--mounted in such a way as to treat *every* file as having executable bits set. Then, the output of commands like `ls -l` is not illuminating, but also, for other reasons such as line endings, one should typically not work in a Git repository that way. For this reason I have included in #1693 an explicit mention in the readme that Windows users can invoke the init script from Git Bash. This suggestion is independent of actual changes made to that file, except inasmuch as both are with the aim of improving the experience on less well supported platforms.)

> When thinking about it, at this point most of the work was already done and running `make release` locally isn't a problem, with the only 'inconvenience' that I have to type `__token__` and then paste a passcode.

Publishing with GitHub Actions could potentially replace some of the logic in these scripts. For example, a workflow could be set up that builds and published to PyPI when a release is made on GitHub, or when a push creates a tag, comes from a PR with a particular label, etc. (If one of those latter triggers is used, the workflow could even create the GitHub release as well, populating it with the default description or with a custom-generated one.)

[Here's an example](https://github.com/bazingagin/npc_gzip/blob/main/.github/workflows/publish.yml) of building and publishing to PyPI when a release is made on GitHub (though that project uses the dependency manager `poetry` to build the package). This is technically separate from the use of trusted publishing, in that publishing could be done in the workflow by running a `twine` command supplied with access credentials sourced from a repository-level GitHub Actions secret. However, these days, trusted publishing is increasingly used, which avoids the need to store PyPI access credentials in a CI secret store.

> These few seconds could be saved at the expense of complexity, abstraction, and giving some secrets to the cloud.

I think this is strictly speaking true: any user who has the power to make arbitrary changes to the repository can use that power to cause a release to be published to PyPI, when trusted publishing is used. However, no PyPI secret has to be stored in the repository--changes are validated as having originated from the GitHub repository (and more specifically as from a CI job produced from a particular workflow whose file name has been put in via the PyPI web interface while setting it up), without involving *separate* stored credentials.

Byron:
Thanks for illuminating what happens to executable bits on Windows under various conditions - I see that it's definitely not portable. I think I am slowly starting to be less ignorant towards Windows, once again.

Regarding publishing through GitHub actions, I see as main benefit that it would allow other maintainers to create valid publishes, and this is what I would want to use it for as well.

EliahKagan:
When using Windows, I usually use PowerShell rather than a Unix-style shell (as my interactive shell), so `ls` doesn't show me Unix-style permissions (inferred from hashbangs or otherwise). But `git` itself can tell me what permissions are recorded in its index, and also allows me to change them. `git ls-files -s` gives output in a form I find cumbersome, but I use [a custom `git` command](https://github.com/EliahKagan/gitscripts/blob/master/git-lsx) that omits the hash column and less interesting mode bits, and shows executable files and symlinks with colorization and indicator sigils (like `ls` shows filenames when `--color=auto` and `-F`/`--classify` are passed).

> Regarding publishing through GitHub actions, I see as main benefit that it would allow other maintainers to create valid publishes, and this is what I would want to use it for as well.

Yes, I agree. However, it may be that the process of building and publishing releases could be simplified in other ways, if and when this is done.

PR #1662, opened 2023-09-17
Fix URLs that were redirecting to another license

EliahKagan:
All the opensource.org BSD [license URLs](http://www.opensource.org/licenses/bsd-license.php) at the top of source code files in this project had [originally pointed](https://web.archive.org/web/20080907042355/http://www.opensource.org/licenses/bsd-license.php) to a page on the 3-clause BSD license that this project [used](https://github.com/gitpython-developers/GitPython/commit/33ebe7acec14b25c5f84f35a664803fcab2f7781#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7) and [continues to use](https://github.com/gitpython-developers/GitPython/blob/main/LICENSE).

But over time the site was apparently reorganized and the link became a redirect to the page about the 2-clause BSD license. Because it is identified only as the "BSD license" in the comments in this project that contain the links, this unfortunately makes it so those top-of-file comments all wrongly claim that the project is 2-clause BSD licensed.

To be more specific, [bisecting archivals](https://web.archive.org/web/20110501000000*/http://www.opensource.org/licenses/bsd-license.php) reveals that the change apparently occurred sometime between [29 April 2011](https://web.archive.org/web/20110429232045/http://www.opensource.org/licenses/bsd-license.php) and [5 June 2011](https://web.archive.org/web/20110605023935/http://www.opensource.org/licenses/bsd-license.php).

This fixes the links by replacing them with the current URL of the opensource.org page on the 3-clause BSD license. The current URL contains "bsd-3-clause" in it, so this specific problem is unlikely to recur with that URL (and even if it did, the text "bsd-3-clause is information that may clue readers in to what is going on).

There may be other changes that could be made in the future to further clarify the license. For example, in `setup.py`, the license name is passed as `BSD`. There may be a more specific name that can be used and that PyPI and/or common tools will recognize for the 3-clause BSD license name. (If any [SPDX](https://spdx.org/licenses/) name can be used, which I think it can, then this is definitely so. However, a more specific *classifier* does not appear available.) But that's not actually a bug--it's not referring to a specific different license--and this PR does not make any such changes. Other than how I permitted my editor to remove trailing whitespace in the files it was saving, this pull request is strictly limited to updating those URLs.

This also affects numerous URLs in `gitdb`, for which I have opened https://github.com/gitpython-developers/gitdb/pull/96. It does not affect `smmap`.

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -2,7 +2,7 @@
 # Copyright (C) 2008, 2009 Michael Trier (mtrier@gmail.com) and contributors
 #
 # This module is part of GitPython and is released under
-# the BSD License: http://www.opensource.org/licenses/bsd-license.php
+# the BSD License: https://opensource.org/license/bsd-3-clause/
 # flake8: noqa
 # @PydevCodeAnalysisIgnore
 from git.exc import *  # @NoMove @IgnorePep8
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -2,7 +2,7 @@
 # Copyright (C) 2008, 2009 Michael Trier (mtrier@gmail.com) and contributors
 #
 # This module is part of GitPython and is released under
-# the BSD License: http://www.opensource.org/licenses/bsd-license.php
+# the BSD License: https://opensource.org/license/bsd-3-clause/
 from __future__ import annotations
 import re
 import contextlib
--- a/git/compat.py
+++ b/git/compat.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2008, 2009 Michael Trier (mtrier@gmail.com) and contributors
 #
 # This module is part of GitPython and is released under
-# the BSD License: http://www.opensource.org/licenses/bsd-license.php
+# the BSD License: https://opensource.org/license/bsd-3-clause/
 """utilities to help provide compatibility with python 3"""
 # flake8: noqa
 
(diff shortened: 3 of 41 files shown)


Byron:
Thanks a million! It's quite amazing for how long some of these bugs can hide, but also, that a site that hosts licenses is willing to change a 3-clause BSD into a 2-clause BSD license, which seems like a breaking change.

EliahKagan:
I would consider that a *bug* rather than a breaking change, whether or not it was intentional. I suspect it was not intentional, or at least that the effect was not intended. Looking at archive.org saves of the page, it looks like it had at one time been the only "BSD license" listed on opensource.org, back when the OSI had formally approved the 3-clause license but not yet the 2-clause license, or something like that. The page was updated to mention prominently that, in choosing a license, the third clause can be removed and the result is still an open source license. Then it later became a page for the 2-clause license. So the page may have appeared, to people maintaining the site, to be a general page not specific to any one version of the BSD license.

My guess is that, in addition, the mindset at the time was oriented toward informing people of what OSI-approved licenses they might choose for their software, rather than identifying a specific license to communicate that it has been used. I don't think this diminishes the bug, but it may help explain it, because that view would explain the idea that noting that there is a 3-clause version and linking to it (which the page does) would be sufficient. That is enough to inform people of what licenses exist, but not enough to inform people of what license someone is trying to tell them about.

Anyway, I think this could still be fixed in opensource.org. The old URL is a redirect, and it could be made to go to a disambiguation page, or to go to its current target but with some `?name=value` query string added that causes the server to include a note at the top, and so forth. I might at some point manage to look into whether anyone has reported this as a bug and how to do so.

PR #1665, opened 2023-09-18
Omit py_modules in setup

EliahKagan:
This removes the `py_modules` keyword argument in the call to `setup`, and further shortens/simplifies `setup.py` by removing the newly unused `build_py_modules` function.

The `packages` keyword argument already covers this, because we have no loose modules that are included in the distribution, and the call to `find_packages`:

- Omits everything in `test/` because it is directed to do so in the call.
- Omits the `gitdb/` directory (currently existing as a git submodule, not to be confused with Python submodules). This is because the `ext/` directory that contains it does not itself directly contain an `__init__.py` file, so `ext/` is not a traditional package, yet `ext/` is contained and found inside the directory `git/` that *is* a traditional package, so the `ext/` directory is not a namespace package either.
- Includes all other modules, recursively, because they are all in a recursive traditional package structure under `git/` that `find_packages` recognizes.

To verify that this includes the same files in the built wheel and sdist distributions, I have listed the contents of the wheel with `unzip -l` and the sdist .tar.gz file with `tar tf` both before and after this change, verifying they list all the same entries before and after.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,6 @@ from typing import Sequence
 from setuptools import setup, find_packages
 from setuptools.command.build_py import build_py as _build_py
 from setuptools.command.sdist import sdist as _sdist
-import fnmatch
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! It's amazing to see that, at least by now, the manual package collection isn't at all necessary anymore and I wonder why it ever was. Oh well 😅.

PR #1666, opened 2023-09-18
Don't track code coverage temporary files

EliahKagan:
While running tests with coverage enabled (as is automatically the case in this project), files of the form `.coverage.MachineName.####.######` are created temporarily. For example, on my system, one of the names was recently `.coverage.Glub.7791.638878`.

When tests complete normally (whether or not there are failures), these files are automatically removed. However, when tests are cancelled with SIGINT (<kbd>Ctrl</kbd>+<kbd>C</kbd>), they are sometimes retained.

This adds another pattern to `.gitignore` so that, in addition to not tracking the `.coverage` file that is retained after tests, these other temporary files are also not tracked.

Further considerations:

- This also has the benefit of avoiding a situation at development time where one runs tests and also stages deliberate code changes at the same time, causing these files to be unintentionally staged along with the code. Relatedly, editors that frequently update with information about source control changes will no longer show these entries flashing in and out of existence, and (least importantly) such editors may run slightly faster in some cases.
- I considered simply changing existing `.coverage` pattern to `.coverage*`, but I decided against it, because if someone ever introduces a [`.coveragerc`](https://coverage.readthedocs.io/en/latest/config.html) file, they would expect that not to be ignored by source control. (They should probably not do this, since `pyproject.toml` is being used for this instead, but if they do, then it should not be automatically ignored.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ venv/
 /lib/GitPython.egg-info
 cover/
 .coverage
+.coverage.*
 /build
 /dist
 /doc/_build


Byron:
Thanks a lot, I perfectly agree with the solution and the reasoning.

While at it, I validated that there are no currently ignored files checked in either.

```
❯ gix index entries -s
[..]
Statistics {
    entries: 183,
    entries_after_prune: 183,
    excluded: 0,
    with_attributes: 76,
    max_attributes_per_path: 1,
    cache: Some(
        Statistics {
            platforms: 183,
            delegate: Statistics {
                num_mkdir_calls: 0,
                push_element: 198,
                push_directory: 16,
                pop_directory: 14,
            },
            attributes: Statistics {
                patterns_buffers: 0,
                pattern_files: 1,
                tried_pattern_files: 17,
            },
            ignore: Statistics {
                patterns_buffers: 0,
                pattern_files: 2,
                tried_pattern_files: 16,
            },
        },
    ),
    attributes: [
        "eol=lf",
    ],
    submodule: [],
}
All encountered attributes:
        eol=lf

GitPython ( main) [!] via 🐍
```

EliahKagan:
Nice! I had not thought to do that.

PR #1668, opened 2023-09-19
Format tests with black and auto-exclude untracked paths

EliahKagan:
This expands the use of `black` to cover the whole codebase (still excluding submodules).

- Replaced `exclude` with `extend-exclude` so `black`'s default exclusions, which include anything specified in `.gitignore`, remain in effect as well as the specific path(s) specified to be excluded. This allows `black .` to be run to format everything that should be formatted and nothing that shouldn't.
- Made a few manual formatting adjustments that lead to more readable code in a few places once formatted with `black`. (This does not have to be redone; the idea is to be in the common subset of what was intended and what `black` considers its style.)
- Changed `black git` to `black .` in the instructions in `README.md` and in the command in `tox.ini` that runs via `tox` or `tox -e black`.
- Actually ran `black .`, formatting the tests in `black` style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ mypy -p git
 For automatic code formatting, run:
 
 ```bash
-black git
+black .
 ```
 
 Configuration for flake8 is in the `./.flake8` file.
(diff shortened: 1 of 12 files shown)


Byron (approved):
Thanks a lot, looks good to me.

It's interesting to see that the rest of the project apparently is formatted with it already, I wasn't aware.

Issue #1669, opened 2023-09-21
SymbolicReference.set_reference may not roll back properly on error


EliahKagan:
`SymbolicReference.set_reference` contains this code, which is meant to attempt a transactional write:

https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/refs/symbolic.py#L373-L380

However, because the `ok` flag is *initialized* to `True` instead of `False`--which strongly appears unintentional--it is always `True` even if the `ok = True` statement in the `try` block is not reached. Consequently, `lfd.rollback()` in the `finally` block is never called. This seems on first glance like a serious bug, but as detailed below I believe it might actually be quite minor in the context of what kind of cleanup GitPython generally [does or does not guarantee](https://github.com/gitpython-developers/GitPython?tab=readme-ov-file#leakage-of-system-resources).

This can be contrasted to another place where that pattern is used correctly:

https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/index/base.py#L227-L233

This is very simple to fix, since it is sufficient to change `True` to `False` in the initial assignment. So most of the work to fix it would be figuring out where a regression test should go and how it should manage to produce a case where cleanup doesn't happen immediately.

(Besides that simple fix, other fixes that also reduce the risk of such bugs coming back and allow the code to be simplified are possible. In particular, the `LockedFD` object represents a non-memory resource that always can and should be managed deterministically according to where control flow is, so it ought to implement the context manager protocol.)

I am unsure if this bug causes any problems in practice. If I understand [`LockedFD`](https://github.com/gitpython-developers/gitdb/blob/master/gitdb/util.py#L269) correctly, the write that is not rolled back is to a temporary file separate from the real target, accessed through a file object specific to that `LockedFD` object. The lock file name is determined from the target file name, so if a separate attempt to write to the target is made, that could cause problems. `LockedFD` has a `__del__` method that should be called eventually and performs the rollback. On CPython, which has reference counting as its primary garbage collection strategy, cleanup would often happen immediately as a result... though not always, even in the absence of reference cycles, because, in exception handling, an exception object holds a reference to a traceback object which holds references to all stack frames in its call stack, through which local variables thus remain accessible until control leaves the handling `except` block.

[...]

PR #1673, opened 2023-09-21
Upgrade and broaden flake8, fixing style problems and bugs

### LINKED ISSUE #1670: flake8 pre-commit config incompatible with 3.12
The configuration for `flake8` in `.pre-commit-config.yaml`--specifically that version 6.0.0 is used--is incompatible with Python 3.12, producing an incorrect "E231 missing whitespace after ':'" error:

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (scripts)$ pre-commit run --all-files
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

git/cmd.py:479:34: E231 missing whitespace after ':'
                f"The `{protocol}::` protocol looks suspicious, use `allow_unsafe_protocols=True` to allow it."
                                 ^
git/cmd.py:479:35: E231 missing whitespace after ':'
                f"The `{protocol}::` protocol looks suspicious, use `allow_unsafe_protocols=True` to allow it."
                                  ^
git/refs/log.py:247:67: E226 missing whitespace around arithmetic operator
                    raise IndexError(f"Index file ended at line {i+1}, before given index was reached")
                                                                  ^
1     E226 missing whitespace around arithmetic operator
2     E231 missing whitespace after ':'
3

check for merge conflicts................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
```

It also reports E226, which also happens only on Python 3.12, but appears intentional. But that's easy to fix (the simplest way being to add the spaces it wants, which also more resembles the overall style for the `+` operator in this project).

The wrong `:` error (false positive E231) happens because `flake8` 6.0.0 depends on `pycodestyle` 2.10.\*, which implements that check (E231) and has a parsing bug specific to Python 3.12. This was fixed in https://github.com/PyCQA/pycodestyle/pull/1148.

Upgrading to `flake8` 6.1.0 gets the fix, since that version of `flake8`, currently the latest, [depends on](https://flake8.pycqa.org/en/latest/release-notes/6.1.0.html) `pycodestyle` 2.11.\*, [which has it](https://pycodestyle.pycqa.org/en/latest/developer.html#changes):

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (scripts)$ pre-commit autoupdate
[https://github.com/PyCQA/flake8] updating 6.0.0 -> 6.1.0
[https://github.com/pre-commit/pre-commit-hooks] already up to date!
(.venv) ek@Glub:~/repos-wsl/GitPython (scripts *)$ pre-commit run --all-files
[INFO] Installing environment for https://github.com/PyCQA/flake8.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

git/objects/submodule/base.py:1406:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
            if type(sm) != git.objects.submodule.base.Submodule:
               ^
git/refs/log.py:247:67: E226 missing whitespace around arithmetic operator
                    raise IndexError(f"Index file ended at line {i+1}, before given index was reached")
                                                                  ^
git/util.py:1139:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
            if type(base) == IterableClassWatcher:
               ^
1     E226 missing whitespace around arithmetic operator
2     E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
3

check for merge conflicts................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
```

However, that reports E721 about the use of `==` to compare to a `type` object. So I think it makes sense for a fix for this to be part of a larger change that also fixes the new problems found by `flake8`, also bumps versions on the listed `flake8` plugins (other than the default ones like `pycodestyle`) and fixes new warnings from that, and possibly addresses other style issues and/or expands the scope of `flake8` checks to cover the test suite. (The benefit of the latter is that it could reveal areas where the checks ought to be configured differently.) I plan to open such a PR shortly.

### LINKED ISSUE #1671: test_git_work_tree_env renders os.environ inert in unpatching attempt
`TestRepo.test_git_work_tree_env` contains this code, which is intended to patch and unpatch two environment variables:

https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/test/test_repo.py#L1341-L1350

However, that does not unpatch the variables, and more importantly, it prevents writes through `os.environ` from ever actually setting environment variables (to be inherited automatically by subprocesses) for the rest of the process's lifetime.

```text
>>> import os
>>> type(os.environ)
<class 'os._Environ'>
>>> oldenv = os.environ.copy()
>>> type(oldenv)
<class 'dict'>
>>> os.system("printenv FOO")
256
>>> os.environ["FOO"] = "bar"
>>> os.system("printenv FOO")
bar
0
>>> os.environ = oldenv
>>> os.system("printenv FOO")
bar
0
>>> os.environ["BAZ"] = "quux"
>>> os.system("printenv BAZ")
256
```

This happens because, all types in Python being reference types, the `environ` attribute of `os` merely *refers to* the object of a special mutable mapping type that updates the process's environment variables when mutated. Calling its `copy` method constructs a `dict` object with the same items. Assigning that `dict` back to `os.environ` does not modify the original `os.environ` object in any way, but instead causes the `environ` attribute of `os` to point to the `dict` object from that point forward. From then on, writes to `os.environ` have no effect on the running process's environment variables.

Yet, this does not cause any other tests to fail, nor do any currently passing tests fail as a result of fixing it. I believe this is for three reasons, which I list in descending order of what I *guess* to be their significance:

1. When GitPython runs `git`, it doesn't rely on automatically passing its own environment variables to the `git` subprocess. Instead, it builds a modified environment based on its own and passes that explicitly. The way it finds out about its own environment is by consulting `os.environ`, which doesn't actually have to work for setting environment variables, because `Popen` is doing it:

   https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/cmd.py#L948-L955

   https://github.com/gitpython-developers/GitPython/blob/a5a646494393478c65f26cd3a921f3505219d3e1/git/cmd.py#L987-L989

2. When GitPython, or any library it uses if that library is written in Python, or the testing framework, accesses and changes environment variables for use within the process, that is also nearly always via `os.environ`, so again, it's okay if the real environment is not used.

3. The test is in `test_submodules.py`, and test modules are usually exercised in alphabetical order, with only `test_tree.py` and `test_util.py` coming after it.

This only affects the tests, and it can be solved in any way appropriate for tests. Specifically, it can be solved--and also that test code significantly simplified--by patching `os.environ` with `unittest.mock.patch.dict`. We cannot use this in the code of the `git` module, because it upcases environment variable names (#1646). But it is not a problem to use it in tests (besides the test in #1650 that #1646 is fixed), and it is already being used in a few places in the test suite.

---

The way I found out about this was that `flake8` reported it once I removed `test/` from its excluded directories:

```text
test/test_repo.py:1350:13: B003 Assigning to `os.environ` doesn't clear the environment. Subprocesses are going to see outdated variables, in disagreement with the current process. Use `os.environ.clear()` or the `env=` argument to Popen.
            os.environ = oldenv
            ^
```

The specific suggestions aren't quite applicable here (we do need to mutate `os.environ`, and to do so less drastically than with `clear`), but that message nonetheless accurately identifies the problem, which seems previously to have gone undetected. I think it is a good idea to enable `flake8` for `test/` as well as `git/` (though perhaps sometime soon we might manage to replace `flake8` with [`ruff`](https://docs.astral.sh/ruff/)). I plan to include a fix for this, and also for less serious, mostly merely stylistic issues found while examining `test/` with the help of `flake8`, in the same PR that fixes #1670.

EliahKagan:
Fixes #1670
Fixes #1671

I've upgraded `flake8` in the pre-commit configuration, which automatically upgrades its `pycodestyle` dependency. This makes it compatible with Python 3.12 (or at least we seem unaffected by any remaining incompatibilities). I've also upgraded its non-default plugins, which are listed there as additional dependencies. I have fixed a warning that was for some reason specific to 3.12 about spacing around `+`, as well as some new warnings that came in the new version of `pycodestyle` and new versions of the non-default packages. This included changing `==` to `is` and `!=` to `is not` when comparing `type` objects, since in all cases it appeared *exact* type matching was intended (rather than what an `issubclass` or, under simplification of the surrounding code, `isinstance` would check). See #1670.

I've also broadened `flake8` to check the whole project (except the `doc/` directory). That is, it now checks the test suite. It issued a number of warnings, and I fixed the code accordingly. Most of these are style changes, but it revealed #1671, which I fixed as described there (by using `unittest.mock.patch.dict`, which as noted there is okay because it is only in the tests). I then looked for bugs in `finally` cleanup logic throughout the project, in case anything else like that was present that I could identify and fix. I found no further serious bugs inside `test/` while doing this, but I did find some areas I could simplify or otherwise improve, including one place where a throwaway environment variable `FOO` was never unpatched.

I included those changes in this PR, but not any changes to code in `git/` other than to fix what `flake8` found. One place in `git/` that should be changed is a bug that merits fixing in some way, which I opened #1669 for. Other areas of improvement in `git/` related to the use of `finally` are less important, and some of them subjective. I could include them in a fix for #1669 (if I end up fixing it) or in a separate PR, but I've omitted those further changes in `git/` from this PR to keep its scope from creeping too large.

**An area that I think deserves special attention in review** is changes I made in parts of the test code that are reflected in built documentation that will be published on readthedocs:

- Where `@NoEffect` appeared in code that is copied to automatically generate usage examples in the documentation, I expanded it to `# noqa: B015  # @NoEffect`. Readability seems still intact, and I am inclined to think this is justified as long as we need to have `@NoEffect`. However, if `@NoEffect` is no longer needed, then I think configuring `flake8` to disable the specific warning in those two specific test modules would be better, since then neither `@NoEffect` nor `# noqa: B015` would need to appear anywhere in documentation example code.
- Comprehensions that perform neither mapping nor filtering are only occasionally useful, and almost never when they are list comprehensions: code like `[c for c in commits_for_file_generator]` should be `list(commits_for_file_generator)` instead. Expressions of that form appear a number of times in tests from which documentation is automatically generated, and I think changing it in this way is only beneficial. However, since it is a code change that will affect documentation (which might otherwise be unexpected), I do want to call attention to it, too.

In most cases I have retained existing `noqa` comments, and I suspect a number of them could be removed. That could be done at any time, and it was also only to limit scope that I have not tried to do it here. Besides that, however, there is a different reason I have deliberately avoided going further with `flake8`--for example, by adding it and its extra plugins as development dependencies of the project (they are still only installed by and for `pre-commit`), running it on CI, enabling more checks, or attempting to get the `flake8-type-checking` plugin listed in `requirements-dev.txt` working. The reason is that I think it would be beneficial to replace `flake8` in this project with [`ruff`](https://docs.astral.sh/ruff/), which is modern, versatile, and extremely fast. (This is also why I have not proposed that we also adopt [`isort`](https://pycqa.github.io/isort/), even though `isort` is one of my favorite tools: `ruff` might take care of that, too.)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -26,7 +26,7 @@ ignore = E265,E266,E731,E704,
          D,
          RST, RST3
 
-exclude = .tox,.venv,build,dist,doc,git/ext/,test
+exclude = .tox,.venv,build,dist,doc,git/ext/
 
 rst-roles =  # for flake8-RST-docstrings
     attr,class,func,meth,mod,obj,ref,term,var  # used by sphinx
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
+    rev: 6.1.0
     hooks:
       - id: flake8
         additional_dependencies:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,8 +5,8 @@ repos:
       - id: flake8
         additional_dependencies:
           [
-            flake8-bugbear==22.12.6,
-            flake8-comprehensions==3.10.1,
+            flake8-bugbear==23.9.16,
+            flake8-comprehensions==3.14.0,
             flake8-typing-imports==1.14.0,
           ]
         exclude: ^doc|^git/ext/|^test/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
             flake8-comprehensions==3.14.0,
             flake8-typing-imports==1.14.0,
           ]
-        exclude: ^doc|^git/ext/|^test/
+        exclude: ^doc|^git/ext/
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -1403,7 +1403,7 @@ class Submodule(IndexObject, TraversableIterableObj):
             # END handle critical error
 
             # Make sure we are looking at a submodule object
-            if type(sm) != git.objects.submodule.base.Submodule:
+            if type(sm) is not git.objects.submodule.base.Submodule:
                 continue
 
             # fill in remaining info - saves time as it doesn't have to be parsed again
--- a/git/refs/log.py
+++ b/git/refs/log.py
@@ -244,7 +244,7 @@ class RefLog(List[RefLogEntry], Serializable):
             for i in range(index + 1):
                 line = fp.readline()
                 if not line:
-                    raise IndexError(f"Index file ended at line {i+1}, before given index was reached")
+                    raise IndexError(f"Index file ended at line {i + 1}, before given index was reached")
                 # END abort on eof
             # END handle runup
 
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -206,7 +206,8 @@ class Repo(object):
         if expand_vars and re.search(self.re_envvars, epath):
             warnings.warn(
                 "The use of environment variables in paths is deprecated"
-                + "\nfor security reasons and may be removed in the future!!"
+                + "\nfor security reasons and may be removed in the future!!",
+                stacklevel=1,
             )
         epath = expand_path(epath, expand_vars)
         if epath is not None:
--- a/git/util.py
+++ b/git/util.py
@@ -1136,7 +1136,7 @@ class IterableClassWatcher(type):
 
     def __init__(cls, name: str, bases: Tuple, clsdict: Dict) -> None:
         for base in bases:
-            if type(base) == IterableClassWatcher:
+            if type(base) is IterableClassWatcher:
                 warnings.warn(
                     f"GitPython Iterable subclassed by {name}. "
                     "Iterable is deprecated due to naming clash since v3.1.18"
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -94,17 +94,16 @@ def with_rw_directory(func):
         os.mkdir(path)
         keep = False
         try:
-            try:
-                return func(self, path)
-            except Exception:
-                log.info(
-                    "Test %s.%s failed, output is at %r\n",
-                    type(self).__name__,
-                    func.__name__,
-                    path,
-                )
-                keep = True
-                raise
+            return func(self, path)
+        except Exception:
+            log.info(
+                "Test %s.%s failed, output is at %r\n",
+                type(self).__name__,
+                func.__name__,
+                path,
+            )
+            keep = True
+            raise
         finally:
             # Need to collect here to be sure all handles have been closed. It appears
             # a windows-only issue. In fact things should be deleted, as well as
@@ -147,12 +146,11 @@ def with_rw_repo(working_tree_ref, bare=False):
             prev_cwd = os.getcwd()
             os.chdir(rw_repo.working_dir)
             try:
-                try:
-                    return func(self, rw_repo)
-                except:  # noqa E722
-                    log.info("Keeping repo after failure: %s", repo_dir)
-                    repo_dir = None
-                    raise
+                return func(self, rw_repo)
+            except:  # noqa E722
+                log.info("Keeping repo after failure: %s", repo_dir)
+                repo_dir = None
+                raise
             finally:
                 os.chdir(prev_cwd)
                 rw_repo.git.clear_cache()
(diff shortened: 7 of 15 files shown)


Byron (approved):
Thanks a lot for this wonderful cleanup! I consider all of these changes improvements, particularly so if they are visible in the documentation.

About `@NoEffect`, I have no recollection of why this was added and think that less noise is more, if there is a chance to have it removed.

> The reason is that I think it would be beneficial to replace `flake8` in this project with [`ruff`](https://docs.astral.sh/ruff/), which is modern, versatile, and extremely fast. (This is also why I have not proposed that we also adopt [`isort`](https://pycqa.github.io/isort/), even though `isort` is one of my favorite tools: `ruff` might take care of that, too.)

I am all for `ruff` :), but am definitely biased as well ;). If you think it will be better, I definitely encourage you to go ahead with it.

PR #1659, opened 2023-09-12
Fix dynamically-set __all__ variable

DeflateAwning:
Fixes issue #1656 by statically setting the `__all__` variable.
Note that this could probably be done more preciely. The list was generated using the dynamic generation code currently there, printed, and then inserted back in.

PR #1677, opened 2023-09-21
Remove `@NoEffect` annotations

### LINKED ISSUE #1674: `@NoEffect` appears in general documentation examples
`@NoEffect` annotations appear in code examples [in at least one place](https://gitpython.readthedocs.io/en/stable/tutorial.html?highlight=%40NoEffect#understanding-objects) in the generated and published documentation.

This appears to have been to suppress output from some analysis tool that has been used. I don't know if they are ever still needed in the code itself or, if not, if they may ever be needed again. If not, this can be fixed by removing them. Otherwise, there may or may not be a reasonable way to have them omitted from the documentation.

In #1673, as it currently stands, I have worsened this situation by expanding them into `# noqa: B015  # @NoEffect` to also suppress the corresponding warning from `flake8`. But if they are no longer needed, then the suppression comments could be removed altogether, and instead [`per-file-ignores`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores) can be used to configure `flake8` not to issue B015 for any code in test modules from which documentation is generated. This is a broader suppression than the current one, but it would leave other checks intact even in those couple modules, and I think the tradeoff would be worthwhile.

EliahKagan:
Fixes #1674

Along with removing all `@NoEffect` annotations (there were only a handful), this:

- Adds missing asserts, where an expression statement was by itself that was intended as an assertion. This turned out to be the case in the places `@NoEffect` appeared in rendered documentation, making it so no per-file-ignores or other broadened suppressions were needed.
- Fixes misspellings (including one case affecting documentation).
- Adds a FIXME comment for investigating a free-standing expression statement with no obvious side effects that may have been meant as an assertion but [would fail if turned into one](https://gist.github.com/EliahKagan/9c49ca00036b3138458cbbfd0c008ebc).

There are some other commented `@`-style suppression annotations throughout the project. This PR deliberately does not touch those. Some of them express useful intent that should eventually be stated in a different way but currently could be a breaking change to express better. For example, some of them express that a name that is imported but not used is intended to be present, which is better expressed with a static `__all__` listing exactly the names (imports and otherwise) meant to be accessed through the module, but to usefully express this an `__all__` needs to omit names not intended to be accessed that way, which [may be a breaking change](https://github.com/gitpython-developers/GitPython/pull/1659#issuecomment-1718867511).

### DIFF
--- a/test/test_submodule.py
+++ b/test/test_submodule.py
@@ -111,7 +111,7 @@ class TestSubmodule(TestBase):
 
         # force it to reread its information
         del smold._url
-        smold.url == sm.url  # noqa: B015  # @NoEffect
+        smold.url == sm.url  # noqa: B015  # FIXME: Should this be an assertion?
 
         # test config_reader/writer methods
         sm.config_reader()
--- a/test/test_docs.py
+++ b/test/test_docs.py
@@ -167,7 +167,7 @@ class Tutorials(TestBase):
         open(new_file_path, "wb").close()  # create new file in working tree
         cloned_repo.index.add([new_file_path])  # add it to the index
         # Commit the changes to deviate masters history
-        cloned_repo.index.commit("Added a new file in the past - for later merege")
+        cloned_repo.index.commit("Added a new file in the past - for later merge")
 
         # prepare a merge
         master = cloned_repo.heads.master  # right-hand side is ahead of us, in the future
@@ -198,7 +198,7 @@ class Tutorials(TestBase):
 
         # .gitmodules was written and added to the index, which is now being committed
         cloned_repo.index.commit("Added submodule")
-        assert sm.exists() and sm.module_exists()  # this submodule is defintely available
+        assert sm.exists() and sm.module_exists()  # this submodule is definitely available
         sm.remove(module=True, configuration=False)  # remove the working tree
         assert sm.exists() and not sm.module_exists()  # the submodule itself is still available
 
@@ -263,9 +263,9 @@ class Tutorials(TestBase):
         # [8-test_references_and_objects]
         hc = repo.head.commit
         hct = hc.tree
-        hc != hct  # noqa: B015  # @NoEffect
-        hc != repo.tags[0]  # noqa: B015  # @NoEffect
-        hc == repo.head.reference.commit  # noqa: B015  # @NoEffect
+        assert hc != hct
+        assert hc != repo.tags[0]
+        assert hc == repo.head.reference.commit
         # ![8-test_references_and_objects]
 
         # [9-test_references_and_objects]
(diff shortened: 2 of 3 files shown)


test/test_submodule.py
```diff
@@ -111,7 +111,7 @@ def _do_base_tests(self, rwrepo):
 
         # force it to reread its information
         del smold._url
-        smold.url == sm.url  # noqa: B015  # @NoEffect
+        smold.url == sm.url  # noqa: B015  # FIXME: Should this be an assertion?
```
Byron, line 114: If made into an assertion it would fail, I wonder if this means that there is a bug in the submodule implementation or the assertion is simply wrong to begin with. Maybe it's an assertion that doesn't work similarly on all platforms?

Byron (approved):
Thanks a lot! 

I wonder why @NoEffect was chosen back then over making it an assertion in the first place. It's good that this is settled now though.

test/test_submodule.py
EliahKagan: I'm not fully clear on what is expected to happen. In context, we have:

https://github.com/gitpython-developers/GitPython/blob/b27a89f683cda85ebd78243c055e876282df89ee/test/test_submodule.py#L107-L114

When it is made into an assertion, `pytest` shows:

```text
E       AssertionError: assert 'git://gitorious.org/git-python/gitdb.git' == 'https://github.com/gitpython-developers/gitdb.git'
E         - https://github.com/gitpython-developers/gitdb.git
E         + git://gitorious.org/git-python/gitdb.git

test/test_submodule.py:114: AssertionError
```

Is the different remote URL part of what this intends to test? Or is this something that broke at some point (or would have broken, if it were an assertion) as a result of moving the remote to GitHub?

Byron: It's surprising to see a gitorious URL there - where would that be coming from?

When looking at this confused, I'd think it's definitely not suitable as tutorial of any kind. Maybe it's better to either revamp it into something more useful, or remove it entirely.

EliahKagan: I don't think this was part of the tutorial. In this PR, I removed `@NoEffect` everywhere in the tests, not just in lines of code that are included in the generated documentation. (This occurrence was in the submodule tests.)

Byron: Oh, thanks for the clarification, I should have known by looking at the filename in the provided code excerpt.

Since it's already a FIXME, I presume that when trying to improve the GitPython package layout and maybe make submodule tests independent of their containing repository, this will naturally be resolved.

EliahKagan: I think making the tests independent would entail fixing it. Whether or not improving the package layout would depends in more details of how that is done. It could also probably be fixed directly, but this would require figuring out where that old URL came from and what, exactly, the bounds are of what the test intends (or should intend) to test.

PR #1678, opened 2023-09-25
Add missing assert keywords

EliahKagan:
This turns two tuple expression statements--each of an equality comparison and message--that were not being checked or otherwise used, and that were intended to be assertions, into assertions.


Byron:
That just works, nice!

PR #1663, opened 2023-09-17
Assorted small fixes/improvements to root dir docs

EliahKagan:
This contains miscellaneous formatting fixes and minor proofreading in the top-level documentation files, plus making the text in the license section of `README.md` have links both to external information about the license and to the license file itself. The common theme is that these are changes to "documentation" files in the root of the repository that are individually minor but seem worth making and also don't seem like they would fit well in any other PR about anything else.

Note that the changes to the license file are just removal of excess whitespace (the extra blank line at the end, and spaces appearing at the end of lines). This does not attempt to change what license it used, *nor* the text of the license!

Some of are changes that are vaguely related to other PRs (whether existing or planned/aspirational). Changing anything related to the license in #1662 might've made that harder to review and might've given the impression to someone casually looking at the history that the license had changed (the opposite of what was really happening). I might open a future PR updating the readthedocs documentation source in `doc/source`, which may include this kind of proofreading, but it's not obviously better to do those together.

PR #1680, opened 2023-09-27
Fix new link to license in readme

EliahKagan:
In #1663, I linked from the readme to the license file, except I accidentally combined two styles of Markdown links in a way that doesn't actually work, and the link is a 404. This affected only the link to the license file *in* the repo, not to the external page about the license. I'm not sure how I missed this, but my guess is that I tested it using a case-insensitive filesystem, on which the link would incidentally have worked. (It does not work when browsing GitHub, so this PR is fixing a documentation bug.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ gpg --edit-key 4C08421980C9
 
 ### LICENSE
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks!

I missed it too, but at least now validated that the new links are indeed working :).

PR #1681, opened 2023-09-27
Drop unneeded flake8 suppressions

EliahKagan:
Some `flake8` suppressions are no longer needed and can be removed even without changing any Python code. This PR removes them, as well as the comments that were present to explain them. Other comments are retained. I've also reformatted the pre-commit and flake8-specific configuration files for readability.

In #1673 I mentioned I didn't plan to do anything more with `flake8` before trying to replace it with `ruff`, but I realized that these changes make the configuration simpler and shorter. (They *may* also make a switch to `ruff`, if/when it is done, easier, because there are fewer suppression to look up how to do. But I am not asking that this PR be evaluated on that basis.)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,38 +1,26 @@
 [flake8]
+
 show-source = True
-count= True
+count = True
 statistics = True
-# E265 = comment blocks like @{ section, which it can't handle
+
 # E266 = too many leading '#' for block comment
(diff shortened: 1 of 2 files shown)


Byron (approved):
It's like somebody cleared out the basement *(and you realize that nothing is missing really)* :D.

Thank you!

PR #1684, opened 2023-10-01
Update instructions and test helpers for git-daemon

EliahKagan:
On a current Cygwin system with git 2.39.0 (the latest version offered by the Cygwin package manager), `git-daemon` is present, with the Cygwin path `/usr/libexec/git-core/git-daemon.exe`. I believe this has been the case for a long time; as detailed below, Cygwin was not ever treated specially in the way that was assumed, yet no breakage resulted.

The `cygwin-test.yml` workflow does not take any special steps to allow `git-daemon` to work, but all tests have been passing in it even without skipping or xfailing tests that seem related to `git-daemon`.

The `git_daemon_launched` function in `test/lib/helper.py` invokes `git-daemon` directly (rather than through `git daemon`) only when `is_win` evaluates true. This only happens on native Windows systems, while Cygwin is treated the same as (other) Unix-like systems yet still works. So maybe Cygwin `git-daemon` was never a special case.

Whether or not it was, the message about `git-daemon` needing to be findable in a `PATH` search is also under an `is_win` check, and thus is never shown on Cygwin. So I've removed the Cygwin part of that message. (Because the path shown is a MinGW-style path, I have kept the wording about that being for MinGW-git, even though it is no longer needed to disambiguate the Cygwin case.)

The greatest value here is the simplification of the documentation and comments. But for consistency, and to make sure the code and comments don't disagree, I've removed the previous *suppression* of Cygwin special casing: now, Cygwin paths are allowed to be normalized using Cygwin path logic even when passed to `git daemon` as part of the tests. Both ways work on Cygwin in the tests actually being done.

The code change part of this PR only affects test code. Nothing in this pull request modifies anything in the `git` module.

### DIFF
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -302,7 +302,7 @@ def with_rw_and_rw_remote_repo(working_tree_ref):
                     cw.set("url", remote_repo_url)
 
                 with git_daemon_launched(
-                    Git.polish_url(base_daemon_path, is_cygwin=False),  # No daemon in Cygwin.
+                    Git.polish_url(base_daemon_path),
                     "127.0.0.1",
                     GIT_DAEMON_PORT,
                 ):
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -177,12 +177,10 @@ def git_daemon_launched(base_path, ip, port):
     gd = None
     try:
         if is_win:
-            ## On MINGW-git, daemon exists in .\Git\mingw64\libexec\git-core\,
-            #  but if invoked as 'git daemon', it detaches from parent `git` cmd,
-            #  and then CANNOT DIE!
-            #  So, invoke it as a single command.
-            ## Cygwin-git has no daemon.  But it can use MINGW's.
-            #
+            # On MINGW-git, daemon exists in Git\mingw64\libexec\git-core\,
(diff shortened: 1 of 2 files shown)


test/lib/helper.py
```diff
@@ -217,12 +215,11 @@ def git_daemon_launched(base_path, ip, port):
         )
         if is_win:
             msg += textwrap.dedent(
-                r"""
+                R"""
```
EliahKagan, line 218: Using a capital `R` here instead of `r` causes some editors to display the string as truly raw, rather than as a regular expression. `black` permits both case variants, even though they mean the same thing to all Python implementations, to allow for this expressiveness. I found this made the string much easier to read in VS Code. However, you may nonetheless prefer this not be done, or that it be deferred until it can be more consistently applied across the codebase. I certainly have no objection to putting this back to `r`.

Byron: That's a super-interesting feature and I wasn't aware! I think there is great value in having (at least some) editors display the string in a more readable fashion, so yes, please keep making these improvements.

Byron (approved):
That's a great catch, thanks so much!

I see how running on Windows natively gets closer with every PR ❤️.

PR #1687, opened 2023-10-02
Fix Git.execute shell use and reporting bugs

### LINKED ISSUE #1685: Git.execute treats shell=False the same as shell=None
### Expected behavior

The *intention* in [`git.cmd.Git.execute`](https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L827) is that when a `bool` value, `True` or `False`, is passed for `shell`, that value determines whether a shell is used, and that only if this argument is omitted or passed as `None` is the class or object's `USE_SHELL` attribute consulted:

https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L839

https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L903-L905

### The logic error

This is the code [in the `Popen` call](https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L987) that is intended to implement that documented logic:

https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L995

`shell is not None and shell or self.USE_SHELL` is intended to mean `self.USE_SHELL if shell is None else shell` (which could also be written `shell if shell is not None else self.USE_SHELL`). But it actually means `shell or self.USE_SHELL`:

```python
>>> def ternary_truth_table(f):
...     return {(p, q): f(p, q) for p in (True, False, None) for q in (True, False, None)}
...
>>> table1 = ternary_truth_table(lambda p, q: p is not None and p or q)
>>> table2 = ternary_truth_table(lambda p, q: p or q)
>>> table1 == table2
True
>>> table1
{(True, True): True, (True, False): True, (True, None): True, (False, True): True, (False, False): False, (False, None): None, (None, True): True, (None, False): False, (None, None): None}
```

(This resembles the [`type(z)==types.ComplexType and z.real or z`](https://mail.python.org/pipermail/python-dev/2005-September/056510.html) problem that [led to the adoption of conditional expressions](https://peps.python.org/pep-0308/#id5).)

### Possible fix

I would *like* to fix this by doing:

```diff
-                    shell=shell is not None and shell or self.USE_SHELL,
+                    shell=self.USE_SHELL if shell is None else shell,
```

Or, perhaps even more clearly:

```diff
+        if shell is None:
+            shell = self.USE_SHELL
...
-                    shell=shell is not None and shell or self.USE_SHELL,
+                    shell=shell,
```

### Could users be relying on the current (wrong) behavior?

On the surface that seems okay. The problem only happens when `USE_SHELL` is `True`. The class attribute `USE_SHELL` is set to `False` by default. It is named as a constant so, unless documented otherwise, it should really only be set during initialization, or by being monkey-patched inside GitPython's own tests. So it's tempting to think nobody should be relying on what happens if they change it.

Except... setting it from outside GitPython *is* documented. In the changelog entry for [0.3.7](https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/doc/source/changes.rst?plain=1#L669-L684):

https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/doc/source/changes.rst?plain=1#L679-L682

This is [elaborated](https://github.com/gitpython-developers/GitPython/compare/0.3.6...0.3.7#diff-35a18a749eb4d6efad45e56e78a9554926be5526e2ba2159b44311e718450e88) at the `USE_SHELL` class attribute definition:

https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/cmd.py#L282-L286

### The `pythonw` use case

In #126, a default of `shell=True` had been introduced to fix a bug on Windows where console windows would flash on the screen for `git` commands, when using the `pythonw` interpreter executable (selected to open `.pyw` files on Windows):

> The GitPython module works great, but when called from a windowless script (.pyw) each command spawns a console window. Those windows are empty and last only a second, but I think they should not be there at all.
>
> I saw it was using `Popen` to call the `git` executable and added a `shell=True` parameter to avoid spawning those windows.

I don't know if that is in practice ever still an issue or not. On my Windows 10 system, I created a Python 3.7.9 virtual environment (since 3.7 is the lowest version GitPython currently supports), installing GitPython in it, running IDLE in it using `pythonw -m idlelib.idle`, and in IDLE, running statements such as `import git` and `git.Git().execute("git version")`. No console window was created. However, it may be that this particular manual test is flawed, as IDLE, being a REPL, must be handling standard streams in a way that most other `.pyw` applications aren't. Nonetheless, the unwanted window creation seems like it should only have happened due to a bug in Python or Windows. But I am not sure.

### Impact of a fix?

My concern is that existing software may be inadvertently relying on `Git.USE_SHELL` taking precedence over the optional `shell` keyword argument to the `Git.execute` method or to any function that forwards that argument to that method.

I believe this should be fixed, because applications that build on GitPython should be able to rely on documented claims about when shells are used to create subprocesses. But I am unsure if there is anything that should be documented or warned about, or taken account in deciding exactly how to fix this and how, if at all, the impact of the fix should be documented in the changelog or elsewhere.

I plan to open a pull request very soon to propose a fix along the lines of those I suggested above. But I wanted to open an issue to explore this, in case that fix is unsatisfactory, requires refinements, or has obscure effects (perhaps even only in cases that GitPython is used incorrectly) that people might find this by searching about.

### LINKED ISSUE #1686: Git.execute debug message has misleading Popen(shell=...)
While investigating #1685, I noticed that in the debug logging `Git.execute` does to offer insight on the `Popen` call it is about to make, a user inspecting the log would reasonably infer false information about the value passed as the `shell=` keyword argument.

The log message is presented as giving information about how `Popen` is being called, not how `Git.execute` has been called: the message shows `Popen(...)`. I agree with this and I don't consider it a bug. But this means it should faithfully report what is about to be passed to `Popen` in cases where a user reading the log would otherwise be misled into thinking this is what has happened.

Reporting the actual `shell=` argument passed to `Popen` is also more useful to log than the argument passed to `Git.execute` (or its default value, as the case may be). Of course, logging both is possible, for example by adding more text like `[from: None]`, though I am somewhat inclined to think that is not worth doing.

I am not arguing that the logged `Popen(...)` message needs to only show things as they really are in the call. It already replaces passwords to decrease leakage of sensitive information into logs, and it shows an `istream_ok` string representing the most useful information about how the `istream` argument to `Git.execute` affects what is passed as the `stdin` argument to `Popen`. (It may be that this should be logged as `stdin`, I'm not sure.) These are not misleading.

It may be that knowledge that `shell=None` is not a correct thing to pass to `Popen` mitigates this. But `Popen` accepts many arguments, and it's hard to remember what all the possible values are for all of them. Furthermore, I think there is a benefit to logging exactly what is about to be passed to `Popen` here: doing so would have revealed #1685 (assuming inspection of the log with `Git.USE_SHELL` set to `True`). The current way is more prone to error, because nontrivial logic follows the logging intended to make that logic's effect clear; so the log can show `False` even when `True` is passed.

### Possible solutions

Currently, when `shell` is implicitly or explicitly `None`, the log shows lines of the form (with the `...` filled in):

```python
Popen(..., shell=None, ...)
```

I think the best approach is the simple one. This and #1685 can be fixed together by setting the `shell` variable to `self.USE_SHELL` when `shell` started out as `None`, then using `shell` both the `log.debug` and the `Popen` calls. In the log, that would give, assuming `USE_SHELL` is its default `False` value:

```python
Popen(..., shell=False, ...)
```

However, we could show both:

```python
Popen(..., shell=False [from: None], ...)
```

A mentioned above, I favor the former, simpler way. But I thought the latter was worth mentioning, because it is another way to show the information without misleading the user.

I've opened #1687, which fixes this (along with #1686) in the first way.

EliahKagan:
Fixes #1685
Fixes #1686

### Changes to `git` module code

This updates the `shell` variable itself, only when it is `None`, from `self.USE_SHELL`.

(That attribute usually gets its value from `Git.USE_SHELL` rather than being separately instance attribute. But `self.USE_SHELL` is an idiomatic way to access it. Furthermore, although people probably shouldn't set it on instances, people may expect that this is permitted.)

Now:

- `USE_SHELL` is used as a fallback only. When `shell=False` is passed, `USE_SHELL` is no longer consulted. Thus `shell=False` always keeps a shell from being used, even in the non-default case where the `USE_SHELL` attribute has been set to `True`.
- The debug message printed to the log shows the actual value that is being passed to `Popen`, because the updated shell variable is used both to produce that message and in the `Popen` call.

### New tests

This also (actually, first) adds tests, some to check that the correct value is always passed as `shell=` when `Git.execute` calls `Popen`, and others to check that whatever it passes is the same as it claims to be passing in the log. I've made them two separate test methods rather than a test method with multiple assertions, because I think the results of these particular tests are easier to understand that way (even compared to the option of using `self.subTest`). But the major shared logic is extracted to a helper function, where its important subtleties are commented. I have verified that the tests fail, in the expected way, before the bugs are fixed, and then pass, and that these are both the case even on native Windows where we don't (yet) have CI jobs.

The tests cover the six combinations of valid values of the `shell=` argument to `Git.execute` (`None`, `False`, and `True`) and `Git.USE_SHELL` (`False` and `True`). I used `ddt` for this, so there are only two test methods, even though there are six test cases. The tests are simple in principle, with the main source of complexity being the question of how to deal with how the command argument to `Popen` typically differs, including by *type*, based on the `shell` argument, yet the tests must not misbehave, nor induce extraneous misbehavior, even when the code under test has a bug that would ordinarily prevent such agreement.

On Python 3.7 only, this adds [`mock`](https://pypi.org/project/mock/) as a development dependency (in the `test` extra, not as a regular dependency). `mock` is a backport of `unittest.mock`, and a couple of my assertions use a handy feature that was introduced to the mock library in Python 3.8. Besides being harder to write without it, I believe it would be less clearly expressed. (Details are commented with the import logic.)

While adding the new tests, I also renamed `test_it_executes_git_to_shell_and_returns_result` to `test_it_executes_git_and_returns_result`, because that test case does not test that a shell is used to run `git` (and a shell is *not* used in that test).

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -974,6 +974,8 @@ class Git(LazyMixin):
         istream_ok = "None"
         if istream:
             istream_ok = "<valid stream>"
+        if shell is None:
+            shell = self.USE_SHELL
         log.debug(
             "Popen(%s, cwd=%s, universal_newlines=%s, shell=%s, istream=%s)",
             redacted_command,
@@ -992,7 +994,7 @@ class Git(LazyMixin):
                     stdin=istream or DEVNULL,
                     stderr=PIPE,
                     stdout=stdout_sink,
-                    shell=shell is not None and shell or self.USE_SHELL,
+                    shell=shell,
                     close_fds=is_posix,  # unsupported on windows
                     universal_newlines=universal_newlines,
                     creationflags=PROC_CREATIONFLAGS,
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,7 @@
 black
 coverage[toml]
-ddt>=1.1.1, !=1.4.3
+ddt >= 1.1.1, != 1.4.3
+mock ; python_version < "3.8"
 mypy
 pre-commit
 pytest
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -4,23 +4,24 @@
 #
 # This module is part of GitPython and is released under
 # the BSD License: https://opensource.org/license/bsd-3-clause/
+import contextlib
 import os
+import os.path as osp
 import shutil
 import subprocess
 import sys
 from tempfile import TemporaryDirectory, TemporaryFile
 from unittest import mock, skipUnless
 
+import ddt
+
 from git import Git, refresh, GitCommandError, GitCommandNotFound, Repo, cmd
-from test.lib import TestBase, fixture_path
-from test.lib import with_rw_directory
-from git.util import cwd, finalize_process
-
-import os.path as osp
-
 from git.compat import is_win
+from git.util import cwd, finalize_process
+from test.lib import TestBase, fixture_path, with_rw_directory
 
 
+@ddt.ddt
 class TestGit(TestBase):
     @classmethod
     def setUpClass(cls):
@@ -72,6 +73,28 @@ class TestGit(TestBase):
         # order is undefined
         res = self.git.transform_kwargs(**{"s": True, "t": True})
         self.assertEqual({"-s", "-t"}, set(res))
+
+    @ddt.data(
+        (None, False, False),
+        (None, True, True),
+        (False, True, False),
+        (False, False, False),
+        (True, False, True),
+        (True, True, True),
+    )
+    @mock.patch.object(cmd, "Popen", wraps=cmd.Popen)  # Since it is gotten via a "from" import.
+    def test_it_uses_shell_or_not_as_specified(self, case, mock_popen):
+        """A bool passed as ``shell=`` takes precedence over `Git.USE_SHELL`."""
+        value_in_call, value_from_class, expected_popen_arg = case
+        # FIXME: Check what gets logged too!
+        with mock.patch.object(Git, "USE_SHELL", value_from_class):
+            with contextlib.suppress(GitCommandError):
+                self.git.execute(
+                    "git",  # No args, so it runs with or without a shell, on all OSes.
+                    shell=value_in_call,
+                )
+        mock_popen.assert_called_once()
+        self.assertIs(mock_popen.call_args.kwargs["shell"], expected_popen_arg)
 
     def test_it_executes_git_and_returns_result(self):
         self.assertRegex(self.git.execute(["git", "version"]), r"^git version [\d\.]{2}.*$")
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -95,7 +95,7 @@ class TestGit(TestBase):
         with mock.patch.object(Git, "USE_SHELL", value_from_class):
             with contextlib.suppress(GitCommandError):
                 self.git.execute(
-                    "git",  # No args, so it runs with or without a shell, on all OSes.
+                    ["git"],  # No args, so it runs with or without a shell, on all OSes.
                     shell=value_in_call,
                 )
 
@@ -112,7 +112,7 @@ class TestGit(TestBase):
             with mock.patch.object(Git, "USE_SHELL", value_from_class):
                 with contextlib.suppress(GitCommandError):
                     self.git.execute(
-                        "git",  # No args, so it runs with or without a shell, on all OSes.
+                        ["git"],  # No args, so it runs with or without a shell, on all OSes.
                         shell=value_in_call,
                     )
 
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -5,8 +5,10 @@
 # This module is part of GitPython and is released under
 # the BSD License: https://opensource.org/license/bsd-3-clause/
 import contextlib
+import logging
 import os
 import os.path as osp
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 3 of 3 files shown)


Byron (approved):
Thanks so much for the fix which brings the behaviour in line with the documentation, and for adding regression protection.

This PR is also a wonderful example on what's possible with the `unittest` and `mock` libraries and generally what's possible with python testing.

🙏

PR #1688, opened 2023-10-03
Clarify Git.execute and Popen arguments

EliahKagan:
This is a sequel to #1687, improving how parameters to `Git.execute` are documented, and improving debug logging and tests of how they affect the arguments passed to `Popen`.

In the `Git.execute` docstring, the method is no longer described as using a shell, since this is typically (and by default) not the case; the items documenting each parameter are listed in the order of the parameters; and some copyediting is done, for clarity, consistency, spacing, and in a few cases other formatting. I copyedited some other docstrings in the module as well.

The order of names in the `execute_kwargs` set is tweaked to also match the order in which they appear as parameters to `Git.execute`. (This is just for the purpose of code clarity, as `set` objects guarantee no particular iteration order.) Since not all unintentional mismatches between this set and the defined method parameters would cause existing tests to fail, and the failures that would occur would not always immediately show the cause of the problem, I also added a test that `execute_kwargs` has the exact expected relationship to the parameters of `Git.execute`. (Though an alternative might be to generate `execute_kwargs` programmatically from `Git.execute` using a similar technique.)

One part of the test logic in #1687 was unnecessarily complicated, due to swallowing an exception produced by running `git` with no arguments. This changes that by passing `with_exceptions=False` so that exception is never generated.

Another part of the the test logic in #1687 combined claims about the code under test with custom assertion logic in a way that made it hard to see what claims were being made by reading the test code. This fixes that by generalizing, and extracting out, an `_assert_logged_for_popen` test helper method. I also took this opportunity to remove its unstated incompatibility with value representations containing regular expression metacharacters. (Please note that the new code is still only robust enough for the special purpose for which it is intended; it does not actually parse the debug message rigorously.)

As requested in https://github.com/gitpython-developers/GitPython/issues/1686#issuecomment-1743680758, I changed `istream=` to `stdin=` in the debug logging message documenting the `Popen` call. I reused the test helper in writing a new test, `test_it_logs_istream_summary_for_stdin`,
which checks both that it has the new name and that it has the expected simplified value representation. I did not change any parameters or variables called `istream` to `stdin` or any other name; rather, this changes the message to better reflect the `Popen` call it exists to document.

If I would only have changed the displayed parameter name, I would likely not have written a test, but I also wanted to change two other things, in which the test helped verify that correctness was maintained: I reordered the displayed `name=value` representations in the log message to match the relative order in which they are passed to `Popen`. And I eliminated the `istream_ok` variable (which was named like a boolean flag but was not one), because having the logic for producing the string in the logging call makes it better resemble the nonidentical but corresponding logic in the `Popen` call, so they can be compared.

Although this is peripheral to the core concept of the clarity of function arguments, I also renamed and fixed the docstring of a local function that appeared (including by claiming) to be a method.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1012,8 +1012,8 @@ class Git(LazyMixin):
         if as_process:
             return self.AutoInterrupt(proc, command)
 
-        def _kill_process(pid: int) -> None:
-            """Callback method to kill a process."""
+        def kill_process(pid: int) -> None:
+            """Callback to kill a process."""
             p = Popen(
                 ["ps", "--ppid", str(pid)],
                 stdout=PIPE,
@@ -1046,7 +1046,7 @@ class Git(LazyMixin):
 
         if kill_after_timeout is not None:
             kill_check = threading.Event()
-            watchdog = threading.Timer(kill_after_timeout, _kill_process, args=(proc.pid,))
+            watchdog = threading.Timer(kill_after_timeout, kill_process, args=(proc.pid,))
 
         # Wait for the process to return
         status = 0
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -979,7 +979,7 @@ class Git(LazyMixin):
         if shell is None:
             shell = self.USE_SHELL
         log.debug(
-            "Popen(%s, cwd=%s, universal_newlines=%s, shell=%s, istream=%s)",
+            "Popen(%s, cwd=%s, universal_newlines=%s, shell=%s, stdin=%s)",
             redacted_command,
             cwd,
             universal_newlines,
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -851,7 +851,7 @@ class Git(LazyMixin):
             program to execute is the first item in the args sequence or string.
 
         :param istream:
-            Standard input filehandle passed to subprocess.Popen.
+            Standard input filehandle passed to `subprocess.Popen`.
 
         :param with_extended_output:
             Whether to return a (status, stdout, stderr) tuple.
@@ -862,8 +862,7 @@ class Git(LazyMixin):
         :param as_process:
             Whether to return the created process instance directly from which
             streams can be read on demand. This will render with_extended_output and
-            with_exceptions ineffective - the caller will have
-            to deal with the details himself.
+            with_exceptions ineffective - the caller will have to deal with the details.
             It is important to note that the process will be placed into an AutoInterrupt
             wrapper that will interrupt the process once it goes out of scope. If you
             use the command in iterators, you should pass the whole process instance
@@ -876,25 +875,25 @@ class Git(LazyMixin):
             always be created with a pipe due to issues with subprocess.
             This merely is a workaround as data will be copied from the
             output pipe to the given output stream directly.
-            Judging from the implementation, you shouldn't use this flag !
+            Judging from the implementation, you shouldn't use this parameter!
 
         :param stdout_as_string:
-            if False, the commands standard output will be bytes. Otherwise, it will be
-            decoded into a string using the default encoding (usually utf-8).
+            If False, the command's standard output will be bytes. Otherwise, it will be
+            decoded into a string using the default encoding (usually UTF-8).
             The latter can fail, if the output contains binary data.
 
         :param kill_after_timeout:
-            To specify a timeout in seconds for the git command, after which the process
+            Specifies a timeout in seconds for the git command, after which the process
             should be killed. This will have no effect if as_process is set to True. It is
             set to None by default and will let the process run until the timeout is
             explicitly specified. This feature is not supported on Windows. It's also worth
             noting that kill_after_timeout uses SIGKILL, which can have negative side
-            effects on a repository. For example, stale locks in case of git gc could
+            effects on a repository. For example, stale locks in case of ``git gc`` could
             render the repository incapable of accepting changes until the lock is manually
             removed.
 
         :param with_stdout:
-            If True, default True, we open stdout on the created process
+            If True, default True, we open stdout on the created process.
 
         :param universal_newlines:
             if True, pipes will be opened as text, and lines are split at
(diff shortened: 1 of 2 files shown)


test/test_git.py
```diff
@@ -364,3 +374,11 @@ def counter_stderr(line):
...
         self.assertEqual(count[1], line_count)
         self.assertEqual(count[2], line_count)
+
+    def test_execute_kwargs_set_agrees_with_method(self):
+        parameter_names = inspect.signature(cmd.Git.execute).parameters.keys()
+        self_param, command_param, *most_params, extra_kwargs_param = parameter_names
+        self.assertEqual(self_param, "self")
+        self.assertEqual(command_param, "command")
+        self.assertEqual(set(most_params), cmd.execute_kwargs)  # Most important.
+        self.assertEqual(extra_kwargs_param, "subprocess_kwargs")
```
EliahKagan, line 384: Or `git.cmd.execute_kwargs` could be generated in a manner similar to this, and this test, the need to manually update it, and the note in the `Git.execute` docstring about that need, could all be done away with. Maybe something like this:

```python
execute_kwargs = inspect.signature(Git.execute).parameters.keys() - {"self", "command", "subprocess_kwargs"}
```

Right now it is defined very high up in the module, even higher than `__all__`, and this suggests that it may be intended to be available for static inspection. But I don't think any tools will statically inspect a *set* of strings like that (plus, static analysis tools *can* examine the parameters of `Git.execute`... though the incomplete lists of parameters in the `@overload`-decorated stubs that precede it confuse the situation somewhat).

`git.cmd.__all__` contains only `"Git"` and I hope that means code that uses GitPython should not be using `git.cmd.execute_kwargs` or relying on its presence. If possible, perhaps it could be made more explicitly private (`_execute_kwargs`, or if it needs to remain statically defined, maybe `_EXECUTE_KWARGS`) or, even better, removed altogether if the `Git.execute` method's arguments can be inspected efficiently enough without it where `execute_kwargs` is currently used. On the other hand, people may have come to depend on it even though the presence of `__all__` that omits it means no one *should* have depended on it.

Anyway, I do not think any of the changes I suggest in this comment need to be made in this pull request. But I wanted to mention the issue just in case.

git/cmd.py
```diff
@@ -66,10 +66,10 @@
     "with_extended_output",
     "with_exceptions",
     "as_process",
-    "stdout_as_string",
     "output_stream",
-    "with_stdout",
+    "stdout_as_string",
     "kill_after_timeout",
+    "with_stdout",
```
EliahKagan, line 72: This changes only the order in which they are given, not the contents, and the set is equal to the same value as before.

However, I am wondering if `command` should actually be added here. Although the `@overload`-decorated stubs for `Git.execute` define some parameters as keyword-only, in the actual definition no parameter is keyword-only and there is definitional symmetry between `command` and the others. Should I worry about what happens if someone has a `git-command` script that they can usually run as `git command` and tries to run `g.command()` where `g` is a `git.cmd.Git` instance? Or related situations?

Another ramification of the parameters not being keyword-only is that, because they can be passed positionally, i is a breaking change to add a new one elsewhere than the end. Still, regarding https://github.com/gitpython-developers/GitPython/issues/1686#issuecomment-1743680758, if you think a `stdin` parameter should be added as a synonym of `istream`, this could still be done even with `stdin` at the end, with the docstring items documenting the parameters referring to each other to overcome any confusion. I am inclined to say the added complexity is not worthwhile (for example, the function would have to handle the case where they are both passed and with inconsistent values). But a possible argument against this is that the `text` synonym of `universal_newlines` could also be added.

Byron: > Should I worry about what happens if someone has a `git-command` script that they can usually run as `git command` and tries to run `g.command()` where `g` is a `git.cmd.Git` instance? Or related situations?

In these situations it would be great to know why `command` was put there in the first place, and I for one do not remember. I know there never was an issue related to `command` specifically, which seems to indicate it's a limitation worth ignoring or fixing in a non-breaking fashion (which seems possible).

Regarding `istream`, I'd think maintaining stability would let me sleep better at night especially since you seem to agree that it's used consistently here and in `gitdb`. Probably along with the documentation improvements, all is well and better than it was before, which even in the past didn't seem to have caused enough confusion to cause an issue to be opened.

test/test_git.py
Byron: Removing `execute_kwargs` seems like a reduction of complexity, which would always be a valuable reduction of maintenance costs should changes need to be made.

As `execute_kwargs` was never advertised in `__all__` I'd think that it's fair to say that those who depend on it nonetheless new the risk. I think the same argument is valid knowing that nothing is ever truly private, everything can be introspected if one truly wants to, yet it's something one simply has to ignore in order to be able to make any changes to python software once released.

git/cmd.py
EliahKagan: > In these situations it would be great to know why `command` was put there in the first place, and I for one do not remember.

Do you mean why it was *omitted* from the `execute_kwargs` set? If so, my guess is that it was intended to be used positionally while the others were intended to be keyword-only. This is judging by how they are given as keyword-only arguments in the `@overload`s, where they are preceded by a `*,` item in the list of formal parameters, which is absent in the actual function definition (thus causing them to accept both positional and keyword arguments).

But it may be that I am misunderstand what you mean here. It is also possible that their keyword-only status in the `@overload`s was intentionally different from the ultimate definition and intended to provide guidance. (I'm not sure. The `@overload`s are missing most of the arguments, which confuses tooling and seems unintentional.)

Byron (approved):
Thanks a lot for this second round of improvements!

It's at the core of GitPython and any improvements is most certainly valued downstream as long as it's not subtly breaking, which you made sure of.

With that said, I remember that this module is also used to start two long-running python processes lazily when objects are accessed, along with various calls to `git` for doing pretty much anything else.

Also thanks to your work the idea of using `gitoxide` (or rather its still to be created python-bindings) has been forming in my head, which could lead to `git` invocations to be reduced to zero while probably performing as well or better. Of course, anything like it is still years away, but it's something I look forward to.

git/cmd.py
Byron: I am happy to adopt your conclusion in that this is not intentional and since there are negative side-effects, like tooling not working correctly, it's certainly something that could be improved.
If against all odds such an action does create downstream breakage, it could also be undone with a patch release.

EliahKagan: I don't think we should make the parameters in the real definition actually keyword-only, since that is a breaking change. Or, at least, I would not rush to that. It seems likely to me that, at least for the first few, people are passing them positionally.

I believe it is merely the absence of many parameters from the `@overload`-decorated stubs (which precede the real definition) that is causing VS Code not to show or autocomplete some arguments. Adding the missing parameters to the `@overload`-decorated stubs should be sufficient to fix that, and it shouldn't be a breaking change because *(a)* they are just type stubs, so it doesn't break runtime behavior, *(b)* the actual change seems compatible even relative to what the type stubs said before, and *(c)* GitPython doesn't have static typing stability currently anyway, because although `py.typed` is included in the package, `mypy` checks don't pass (nor, per #1659, do `pyright` checks pass).

Issue #914, opened 2019-08-27
3.0.2: failing tests during packaging


dvzrv:
Hi!

I'm currently packaging gitpython for Arch Linux and am about to update to version 3.0.2.
However, during packaging, I tried to run the tests and many of them fail (see attachment).

We're currently on
* git 2.23.0
* python-gitdb 2.0.5
* python-smmap 2.0.5
* python-ddt 1.2.1
* python-nose 1.3.7

I'm running the tests (after build, before install) with `nosetests -vd`.
I'm using the pypi provided sources: https://files.pythonhosted.org/packages/51/46/993beca52f3b609d148071e129235b866626eeb6056f2faffb41d9d727a7/GitPython-3.0.2.tar.gz

I would be very grateful for any hint!
[python-gitpython-3.0.2-check.log](https://github.com/gitpython-developers/GitPython/files/3546275/python-gitpython-3.0.2-check.log)

Byron:
Unfortunately, the GitPython tests indeed have very specific requirements. They expect to be run in their own git repository, with certain additional requirements being met.
Fortunately there is [a script](https://github.com/gitpython-developers/GitPython/blob/master/.travis.yml#L28) as run by travis which sets up these requirements.

Something I am not sure of is if it's enough to simply initialise a new git repository from the sources in the tarball.

I hope that helps, and please let me know if it worked or if there are additional issues.

dvzrv:
I have tried to run the tests for the `3.0.3` build, but still get *a lot* of failing and erroring tests.

As I'm building a package, I can not rely on downloading from several different sources (am I supposed to git clone all of gitpython recursively just for testing purposes?).

That being said, there are plenty of tests, that hardcode specific commits, author names, git tags, files(!), etc. to test against.
This is really bad for reproducibility and makes all of these tests basically useless for downstreams (although they would be potentially be the ones reporting integration problems, etc.)!

This is how I have to prepare the test suite, when using a tarball of gitpython (while still having 34 erroring and 9 failing tests):

```
  local TEST_TMPDIR=$(mktemp -d -t gitpython.XXX)
  local GITDB_TMPDIR=$(mktemp -d -t gitdb.XXX)
  local SMMAP_TMPDIR=$(mktemp -d -t smmap.XXX)
  git config --global user.name "Test User"
  git config --global user.email "test@user.org"
  (                                
    cd "$SMMAP_TMPDIR"
    mkdir -vp bare           
    (
      cd bare                
      git init --bare
[...]

PR #1700, opened 2023-10-09
Fix bugs affecting exception wrapping in rmtree callback

### LINKED ISSUE #1571: Failing (or warning?) with Python 3.12 alpha 7
I think things like this https://github.com/gitpython-developers/GitPython/blob/f25333525425ee1497366fd300a60127aa652d79/git/util.py#L180 is causing a problem, due to https://docs.python.org/3.12/whatsnew/3.12.html#shutil (see https://github.com/python/cpython/pull/102829 ). While I do have warnings as errors on in my test suite, this is not reporting a warning like `wheel` does, it's an error when calling `repo.index.commit("first commit")`:

```
/opt/hostedtoolcache/Python/3.12.0-alpha.7/x64/lib/python3.12/site-packages/git/refs/symbolic.py:172: in _get_ref_info_helper
    repodir = _git_dir(repo, ref_path)
        cls        = <class 'git.refs.head.Head'>
        ref_path   = 'refs/heads/main'
        repo       = <git.repo.base.Repo '/tmp/pytest-of-runner/pytest-0/test_plugin_metadata0/pkg/.git'>
        tokens     = None
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

repo = <git.repo.base.Repo '/tmp/pytest-of-runner/pytest-0/test_plugin_metadata0/pkg/.git'>
path = 'refs/heads/main'

    def _git_dir(repo: "Repo", path: Union[PathLike, None]) -> PathLike:
        """Find the git dir that's appropriate for the path"""
        name = f"{path}"
        if name in ["HEAD", "ORIG_HEAD", "FETCH_HEAD", "index", "logs"]:
            return repo.git_dir
>       return repo.common_dir
E       SystemError: <function Repo.common_dir at 0x7ffba93f3060> returned a result with an exception set

name       = 'refs/heads/main'
path       = 'refs/heads/main'
repo       = <git.repo.base.Repo '/tmp/pytest-of-runner/pytest-0/test_plugin_metadata0/pkg/.git'>

/opt/hostedtoolcache/Python/3.12.0-alpha.7/x64/lib/python3.12/site-packages/git/refs/symbolic.py:55: SystemError
```

GHA's `3.12-dev` just updated to alpha 7, which is why I am now seeing this error.

### LINKED ISSUE #1698: HIDE_WINDOWS_*_ERRORS environment variables are interpreted strangely
### Summary

As noted in #1697:

> The `HIDE_WINDOWS_KNOWN_ERRORS` and `HIDE_WINDOWS_FREEZE_ERRORS` variables are intended to be affected by the same-named environment variables, but this does not work properly. They are `True` by default, but the only way to use environment variables to set them to a false value is *to define the environment variable with an empty value* (since everything else, as a string, is truthy). In addition to being very unlikely to have been the intended behavior, this is also hard to do on Windows when using typical shells, because in both `cmd.exe` and PowerShell setting an environment variable as empty actually removes the variable altogether.

### Details

The attributes have type `bool` outside native Windows systems (where they are `False`) and in native Windows systems where the environment variables don't exist (they are `True`):

https://github.com/gitpython-developers/GitPython/blob/8107cbfeb1d56b2587d347c7f3d509ebef079950/git/util.py#L122-L126

Otherwise they are the `str` values of the same-named environment variables. They are always used as conditions, and as such, the only ways to set them `False` on a native Windows system--besides editing that code, which in practice seems to be what is done--is to write to the attributes with Python code or to set the environment variables to the empty string.

In `bash` on Windows, such as Git Bash, this is easy to do, though not at all obvious that it should be done and also very unintuitive. In shells meant for Windows, `cmd.exe` and PowerShell, setting an environment variable to the empty string causes it to be automatically *unset*, and extra effort is required to actually set the empty string.

### The tricky part

`HIDE_WINDOWS_KNOWN_ERRORS` is defined in `git.util` [**and listed in `__all__`**](https://github.com/gitpython-developers/GitPython/blob/8107cbfeb1d56b2587d347c7f3d509ebef079950/git/util.py#L114). So it is a breaking change to remove or rename it. Because code that uses the GitPython library is explicitly allowed to assume the attribute will continue to exist, it's worth considering the subjective question of what, if anything, users can reasonably assume *about* it.

Although comments shown above explains its original purpose, its semantics are not really documented anywhere, and nothing makes any statements about how it is set. Furthermore, if people are setting nonempty values that intuitively feel like they express falsehood--such as `0`--then they probably expect that this is setting it to `False` or some falsy value. Therefore, I think it's reasonable to regard the current behavior as a bug and fix it, even though actually removing or even officially deprecating the attribute should wait until later because it currently affects the behavior of `git.util.rmtree`, through [the callback](https://github.com/gitpython-developers/GitPython/blob/8107cbfeb1d56b2587d347c7f3d509ebef079950/git/util.py#L185-L196) it passes to `shutil.rmtree`.

In contrast, `HIDE_WINDOWS_FREEZE_ERRORS` is not listed in `__all__` and could be renamed, removed, or arbitrarily changed at any time. But as long as `HIDE_WINDOWS_KNOWN_ERRORS` remains in `git.util`, I don't think the presence `HIDE_WINDOWS_FREEZE_ERRORS` should be seen as any special problem.

Although I don't advocate formally deprecating the `HIDE_WINDOWS_KNOWN_ERRORS` *attribute* just yet, the environment variable is another story. In addition to being parsed differently, the presence of either environment variable could cause a warning to be printed when `git.util` is loaded (which happens when the `git` module is loaded).

### Context

This relates to the deeper issue #790. Fixing this narrower issue won't solve the problem that raising `SkipTest` as a result of failing to delete files can produce unpredictable effects in production--because it signals a condition that would typically be signaled by `PermissionError` or some other exception derived from `OSError`, and would bypass exception handling meant for that--as well as masking bugs in applications or other libraries that use GitPython by causing their automated tests to be skipped. But solving that will carry its own challenge. The main problem is not how to provide the functionality to the test suite, but instead that there may be code using GitPython that has come to rely on exceptions from deleting files being wrapped in `SkipTest`.

Although I definitely think #790 *should* be solved, and there are various approaches to mitigating disruption from doing so, I think it is worth fixing this narrower issue of how the environment variables are interpreted now. I've included such a fix in #1700 (which also would fix #1699).

### LINKED ISSUE #1699: git.util.rmtree's callback wraps too many exceptions
As noted in #790, really *no* exceptions should be wrapped in `unittest.SkipTest` in production, but there is a more specific problem with what exceptions are wrapped, which this issue is about. This issue could be fixed together with #1698 or separately, but the two are independent.

`git.util.rmtree` handles errors with this callback, which wraps them in `unittest.SkipTest`, and which explicitly reports them *as `PermissionError`*:

https://github.com/gitpython-developers/GitPython/blob/8107cbfeb1d56b2587d347c7f3d509ebef079950/git/util.py#L185-L196

The original logic for that was introduced in be44602. It occurs multiple times, but this is actually redundant because they all call `rmtree` from `git.util`, so the code in the callback defined and used in `git.util.rmtree` is sufficient. The goal--considering the context in which it was introduced, the code comments, and the explicit message--is to convert `PermissionError` to `SkipTest`.

**But it catches the much more general exception type `Exception`.** Besides `PermissionError`, there are some other subclasses of `OSError` whose instances can be raised by `shutil.rmtree`, such as `FileNotFoundError`. More surprisingly, I have found that `TypeError` can be raised if directory traversal is itself what fails, due to a function used in opening the directory not supporting being called with just a path.

Even if this logic were only operational when the project's tests are running, which is not the case, I would take the view that extra exceptions should not be caught and wrapped. However, there is a more confusing effect: **no matter what exception is wrapped, the message claims it is a `PermissionError`**, since that name is hard-coded.

I think this should be fixed by having the callback catch `PermissionError` instead of `Exception`, which I think is what anyone who is relying on the existing of wrapping exceptions in `SkipTest` is assuming. (I have found that this change does not cause any fewer tests in the test suite to fail on native Windows systems, at least when running them in Python 3.12 on my WIndows 10 system.) The duplication of that logic can be eliminated at the same time. With the same rationale as in #1698, I think it makes sense to fix this now, rather than deferring it to when a full fix for #790 can be implemented. In addition, and unlike in #1698, a fix for this would somewhat decrease the impact of #790.

I plan to include such a fix in the same pull request I open for #1698.

EliahKagan:
Fixes #1698
Fixes #1699
*May* close #1571

This takes the approaches I suggested in #1698 and #1699 for those two issues. It also adds tests for both the already-working behavior and the new behavior. For the tests of new behavior, I have manually verified they failed in the expected way, and now pass, on native Windows. (Other results can be seen on CI.)

For interpreting the `HIDE_WINDOWS_KNOWN_ERRORS` and `HIDE_WINDOWS_FREEZE_ERRORS` environment variables:

- I have kept `True` as the default (though we may be able to change that in the near future, even before fully addressing #790).
- `0`, `false`, and `no`, as well as the empty string, are treated as `False`. This is case-insensitive and ignoring leading and trailing whitespace.
- While other values are currently treated as `True`, values other than possibly padded and possibly case-varying `1`, `yes`, or `true` show a warning about being unrecognized. This warning is in addition to, not instead of, the warning that is shown from the environment variable being present *at all*.
- The attributes from those variables are always `bool`, now, too. They are never strings.

When limiting the exceptions the `rmtree` callback catches and reraises as `unittest.SkipTest` to `PermissionError`, I made a couple other changes as well:

- I eliminated the duplicated logic, which technically is an additional behavioral change, because in some cases a `PermissionError` was wrapped in a `SkipTest` that was *itself* wrapped in another `SkipTest`. But I don't think this will cause any problems because even if someone relied on that, the only thing they would likely do would have been to follow the chain to the first non-`SkipTest`.
- As noted in #1571, passing a callback as the *`onerror`* parameter to `shutil.rmtree` is deprecated starting in Python 3.12, and the documentation indicates that it is planned for removal in 3.14. So I have used the new *`onexc`* parameter on 3.12 (and later). Because this is not available on 3.11 and earlier, it continues to use `onerror` in those versions. This was easy to implement, because in this case the same callback function naturally works for both, as it does not inspect the error information it receives.

#1571 may or may not be resolved by this, which I think is to a large extent subjective. That issue does two things. It documents a problem from 3.12.0.alpha7, which I *believe* was fixed, as I commented there. It also notes the deprecated status of `shutil.rmtree`'s `onerror` argument and advocates that `onexc` be used instead. This was originally with the idea that using `onerror` caused or contributed to the observed test failures, but even though that was probably not the case, the issue might have been taken to stand for that as well.

I also fixed the type hints on the callback, renamed the callback itself as `handler` so it is no longer named after one of those two keyword arguments but not the other (which I think could cause confusing), and renamed its parameters to more closely match the documentation. As it is a local function that is not returned, this does not raise any compatibility issues. In a similar vein, I've included a number of low-stakes refactorings, mostly just to imports, in modules I was already modifying heavily, mainly in `test_util`.

This also adds a missing `xfail` and updates commented references to the callback-based skipping logic that already existed. Most tests that get skipped in this way, of which there are about eight, do not have comments like this, and I have not added them. The tests that did are those that had previously been marked with `@skipIf` decorators. Finally, I've updated the `git.util.rmtree` docstring, which will affect generated documentation.

This relates to, but only somewhat mitigates and definitely does not solve, #790.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -188,11 +188,11 @@ def rmtree(path: PathLike) -> None:
 
         try:
             func(path)  # Will scream if still not possible to delete.
-        except Exception as ex:
+        except PermissionError as ex:
             if HIDE_WINDOWS_KNOWN_ERRORS:
                 from unittest import SkipTest
 
-                raise SkipTest("FIXME: fails with: PermissionError\n  {}".format(ex)) from ex
+                raise SkipTest(f"FIXME: fails with: PermissionError\n  {ex}") from ex
             raise
 
     return shutil.rmtree(path, False, onerror)
--- a/git/util.py
+++ b/git/util.py
@@ -121,8 +121,14 @@ def _read_env_flag(name: str, default: bool) -> bool:
         name,
     )
 
-    # FIXME: This should treat some values besides "" as expressing falsehood.
-    return bool(value)
+    adjusted_value = value.strip().lower()
+
+    if adjusted_value in {"", "0", "false", "no"}:
+        return False
+    if adjusted_value in {"1", "true", "yes"}:
+        return True
+    log.warning("%s has unrecognized value %r, treating as %r.", name, value, default)
+    return default
 
 
 #: We need an easy way to see if Appveyor TCs start failing,
--- a/git/util.py
+++ b/git/util.py
@@ -5,24 +5,24 @@
 # the BSD License: https://opensource.org/license/bsd-3-clause/
 
 from abc import abstractmethod
-import os.path as osp
-from .compat import is_win
 import contextlib
 from functools import wraps
 import getpass
 import logging
 import os
+import os.path as osp
+import pathlib
 import platform
-import subprocess
 import re
 import shutil
 import stat
-from sys import maxsize
+import subprocess
+import sys
 import time
 from urllib.parse import urlsplit, urlunsplit
 import warnings
 
-# from git.objects.util import Traversable
+from .compat import is_win
 
 # typing ---------------------------------------------------------
 
@@ -42,21 +42,16 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
-    cast,
     TYPE_CHECKING,
+    cast,
     overload,
 )
-
-import pathlib
 
 if TYPE_CHECKING:
     from git.remote import Remote
     from git.repo.base import Repo
     from git.config import GitConfigParser, SectionConstraint
     from git import Git
-
-    # from git.objects.base import IndexObject
-
 
 from .types import (
     Literal,
@@ -75,7 +70,6 @@ T_IterableObj = TypeVar("T_IterableObj", bound=Union["IterableObj", "Has_id_attr
 
 # ---------------------------------------------------------------------
 
-
 from gitdb.util import (  # NOQA @IgnorePep8
     make_sha,
     LockedFD,  # @UnusedImport
@@ -87,7 +81,6 @@ from gitdb.util import (  # NOQA @IgnorePep8
     bin_to_hex,  # @UnusedImport
     hex_to_bin,  # @UnusedImport
 )
-
 
 # NOTE:  Some of the unused imports might be used/imported by others.
 # Handle once test-cases are back up and running.
@@ -182,7 +175,7 @@ def rmtree(path: PathLike) -> None:
     :note: we use shutil rmtree but adjust its behaviour to see whether files that
         couldn't be deleted are read-only. Windows will not remove them in that case"""
 
-    def onerror(function: Callable, path: PathLike, _excinfo: Any) -> None:
+    def handler(function: Callable, path: PathLike, _excinfo: Any) -> None:
         # Is the error an access error ?
         os.chmod(path, stat.S_IWUSR)
 
@@ -195,7 +188,10 @@ def rmtree(path: PathLike) -> None:
                 raise SkipTest(f"FIXME: fails with: PermissionError\n  {ex}") from ex
             raise
 
-    shutil.rmtree(path, False, onerror)
+    if sys.version_info >= (3, 12):
+        shutil.rmtree(path, onexc=handler)
+    else:
+        shutil.rmtree(path, onerror=handler)
 
 
 def rmfile(path: PathLike) -> None:
@@ -995,7 +991,7 @@ class BlockingLockFile(LockFile):
         self,
         file_path: PathLike,
         check_interval_s: float = 0.3,
-        max_block_time_s: int = maxsize,
+        max_block_time_s: int = sys.maxsize,
     ) -> None:
         """Configure the instance
 
--- a/git/util.py
+++ b/git/util.py
@@ -170,17 +170,18 @@ def patch_env(name: str, value: str) -> Generator[None, None, None]:
 
 
 def rmtree(path: PathLike) -> None:
-    """Remove the given recursively.
+    """Remove the given directory tree recursively.
 
-    :note: we use shutil rmtree but adjust its behaviour to see whether files that
-        couldn't be deleted are read-only. Windows will not remove them in that case"""
+    :note: We use :func:`shutil.rmtree` but adjust its behaviour to see whether files that
+        couldn't be deleted are read-only. Windows will not remove them in that case."""
 
     def handler(function: Callable, path: PathLike, _excinfo: Any) -> None:
-        # Is the error an access error ?
+        """Callback for :func:`shutil.rmtree`. Works either as ``onexc`` or ``onerror``."""
+        # Is the error an access error?
         os.chmod(path, stat.S_IWUSR)
 
         try:
-            function(path)  # Will scream if still not possible to delete.
+            function(path)
         except PermissionError as ex:
             if HIDE_WINDOWS_KNOWN_ERRORS:
                 from unittest import SkipTest
--- a/git/util.py
+++ b/git/util.py
@@ -109,14 +109,28 @@ __all__ = [
 
 log = logging.getLogger(__name__)
 
-# types############################################################
+
+def _read_env_flag(name: str, default: bool) -> Union[bool, str]:
+    try:
+        value = os.environ[name]
+    except KeyError:
+        return default
+
+    log.warning(
+        "The %s environment variable is deprecated. Its effect has never been documented and changes without warning.",
+        name,
+    )
+
+    # FIXME: This should always return bool, as that is how it is used.
+    # FIXME: This should treat some values besides "" as expressing falsehood.
+    return value
 
 
 #: We need an easy way to see if Appveyor TCs start failing,
 #: so the errors marked with this var are considered "acknowledged" ones, awaiting remedy,
 #: till then, we wish to hide them.
-HIDE_WINDOWS_KNOWN_ERRORS = is_win and os.environ.get("HIDE_WINDOWS_KNOWN_ERRORS", True)
-HIDE_WINDOWS_FREEZE_ERRORS = is_win and os.environ.get("HIDE_WINDOWS_FREEZE_ERRORS", True)
+HIDE_WINDOWS_KNOWN_ERRORS = is_win and _read_env_flag("HIDE_WINDOWS_KNOWN_ERRORS", True)
+HIDE_WINDOWS_FREEZE_ERRORS = is_win and _read_env_flag("HIDE_WINDOWS_FREEZE_ERRORS", True)
 
 # { Utility Methods
 
--- a/git/util.py
+++ b/git/util.py
@@ -110,7 +110,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-def _read_env_flag(name: str, default: bool) -> Union[bool, str]:
+def _read_env_flag(name: str, default: bool) -> bool:
     try:
         value = os.environ[name]
     except KeyError:
@@ -121,9 +121,8 @@ def _read_env_flag(name: str, default: bool) -> Union[bool, str]:
         name,
     )
 
-    # FIXME: This should always return bool, as that is how it is used.
     # FIXME: This should treat some values besides "" as expressing falsehood.
-    return value
+    return bool(value)
 
 
 #: We need an easy way to see if Appveyor TCs start failing,
--- a/git/util.py
+++ b/git/util.py
@@ -182,12 +182,12 @@ def rmtree(path: PathLike) -> None:
     :note: we use shutil rmtree but adjust its behaviour to see whether files that
         couldn't be deleted are read-only. Windows will not remove them in that case"""
 
-    def onerror(func: Callable, path: PathLike, exc_info: str) -> None:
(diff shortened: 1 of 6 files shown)


git/util.py
```diff
@@ -116,14 +109,33 @@
...
-# types############################################################
+
+def _read_env_flag(name: str, default: bool) -> bool:
+    try:
+        value = os.environ[name]
+    except KeyError:
+        return default
+
+    log.warning(
+        "The %s environment variable is deprecated. Its effect has never been documented and changes without warning.",
```
EliahKagan, line 120: In the case of `HIDE_WINDOWS_FREEZE_ERRORS`, even the attribute itself is nowhere documented and it's not listed in `__all__`, so to say that anything about it is deprecated is actually an understatement: it has never been part of the public interface of any component of GitPython and still isn't. It seems to me that users will not wrongly start believing it is okay to use or rely on something *because* of a claim that it is deprecated. But I could try to think of more precise yet still intuitively clear wording, if you think this is an issue.

test-requirements.txt
```diff
@@ -7,4 +7,5 @@ pre-commit
 pytest
 pytest-cov
 pytest-instafail
+pytest-subtests
```
EliahKagan, line 10: I've used the `subTest` method of `unittest.TestCase` in `test_env_vars_for_windows_tests`. This `pytest` plugin gives `pytest` full support for that mechanism. It lets `pytest` continue with the other subtests even after one fails, as the `unittest` test runner would, and show separate passes and failures for the individual subtests. Separate results are shown only when at least one fails, so when everything passes the report remains uncluttered. (It also provides a `subtests` pytest fixture, though since this is not an autouse fixture it is not usable from within a method of a class that inherits from `unittest.TestCase`.)

I think this is useful to have going forward, since we have many test cases that are large with many separate assertions of separate facts about the system under test, and as they are updated, some of them could be improved by having their separate claims divided into subtests so they can be individually described and so failures don't unnecessarily block later subtests.

However, if you'd rather this plugin be used, it can be removed. `test_env_vars_for_windows_tests` could retain subtests--they will still each run if none fails, just like multiple assertions in a test case without subtests. Or I could replace the subtests with more `@ddt` parameterization, or manually, etc.

Byron: I all sounds reasonable and in particular, since you are the one doing the work it seems fair that you can use the tooling you see as best fit. I also have no experience here and no preferences, and believe that anything that improves the tests in any way is very welcome. Thank you!

git/util.py
Byron: Thanks for playing it so very safe, it's appreciated.

Byron (approved):
Thanks again for your amazing work! GitPython is being pulled out of a deep tech-debt hole thanks to it!

> #1571 may or may not be resolved by this, which I think is to a large extent subjective.

Let's pro-actively assume it works and let them reply if they don't think so to possibly reopen the issue.

I am super-excited to see the progress towards native-windows CI :).

----

Please note that the coming weeks I try to handle merge 1 PR per day due to limited bandwidth. Me being slow isn't a sign of my (maybe perceived lack of) appreciation or care for your work. Thanks for your understanding.

test-requirements.txt
EliahKagan: Sounds good. For the conceptually unrelated reason that it would facilitate more fine-grained `xfail` marking--so we don't habitually carry XPASSing cases that aren't actually expected to have failed--l I think the test class involved here, `test.test_util.TestUtils` should be split, so that the `cygpath`-related tests can be pure `pytest` tests. The reason to *split* it is:

- `@pytest.mark.parametrize` supports fine-grained marking of individual test cases, including as `skip` or `xfail`, but that form of parameterization cannot be applied to test methods in `unittest.TestCase` subclasses. So at least those tests would currently (i.e., until the underlying causes of some cases failing are addressed) benefit from being pure `pytest` tests.
- Some of the test methods, not for `cygpath`, use the `rorepo` fixture provided by `test.lib.helper.TestBase`, which inherits from `unittest.TestCase`. Although this could be converted to a `pytest` fixture, I'd rather wait to do that until after more operating systems, at least Windows, are tested on CI, and also until I have more insight into whether it makes sense to do that at all, rather than *replacing* `rorepo` and other fixtures with new corresponding fixtures that use isolated repositories ([#914](https://github.com/gitpython-developers/GitPython/issues/914), [#1693 (review)](https://github.com/gitpython-developers/GitPython/pull/1693#pullrequestreview-1658090018)). So at least those tests should currently remain in a `unittest.TestCase` subclass.

So long as it's acceptable to have multiple test classes in the same test module, this could be done at any time, and it may facilitate some other simplifications. **I mention it here because** I think it might lead to the elimination of subtests in this particular module, either by using `@pytest.mark.parametrize` for this too or for other reasons.

If that happens, I might remove the `pytest-subtests` test dependency, even though it might be re-added later, because the alternative of [`pytest-check`](https://github.com/okken/pytest-check) may be preferable in some of the large test methods *if* they can also be converted to be pure `pytest` tests (because `pytest-check` supports a more compact syntax).

EliahKagan:
> Please note that the coming weeks I try to handle merge 1 PR per day due to limited bandwidth. Me being slow isn't a sign of my (maybe perceived lack of) appreciation or care for your work. Thanks for your understanding.

No problem!!

I'm unsure if you mean you expect to review one PR *in GitPython* daily (which would be a higher rate than I am currently managing to open them and I would likely notice no difference), or across all repositories (in which case I might notice some difference, since this would mean longer turnaround times for GitPython). But either way that is absolutely no problem! Extremely short turnaround times for reviewing PRs is something I've actually never seen, in any project, until contributing to GitPython, and I definitely do not expect that to always happen!

(Furthermore, in most cases, having one PR open would not prevent me from opening another PR to make a mostly unrelated change, nor when I have multiple PRs open is there any need for them to be reviewed or merged on the same day, week, etc., as each other, nor necessarily in the order in which they were opened.)

Byron:
> I'm unsure if you mean you expect to review one PR _in GitPython_ daily (which would be a higher rate than I am currently managing to open them and I would likely notice no difference), or across all repositories (in which case I might notice some difference, since this would mean longer turnaround times for GitPython

Hehe, I only meant 1 PR per day in GitPython - it's also currently the most active repository thanks to you so most of my time goes here. If you are ever interested how much that it, I do publish [timesheets of my public work](https://github.com/Byron/byron/blob/main/timesheets/2023.csv).

EliahKagan:
Thanks! I was aware of the timesheets' existence, yet I had somehow assumed, without checking, that most of your public work, time-wise, went into `gitoxide`.

Byron:
> [..] thanks to you so most of my time goes here.

Right - most of my *maintenance* time goes here, apologies for the lack of specificity. But even `gitoxide` takes a back-seat sometimes.

EliahKagan:
> Right - most of my *maintenance* time goes here, apologies for the lack of specificity.

That's okay, I had also assumed (erroneously, and I'm not sure why) that most of your maintenance time went into gitoxide. :)

PR #1701, opened 2023-10-11
Fix small #1662 regression due to #1659

EliahKagan:
When #1659 was updated to pick up linting configuration changes, it inadvertently undid one of the URL changes made in #1662, putting the URL in the `git.exc` module back to the one that redirects to a different BSD license from the one this project uses.

Since only that one module was affected, the fix is simple. This only changes the URL back; it doesn't undo any other #1659 changes.


DeflateAwning:
Thank you for seeing and fixing this!

Byron:
What a great catch, thanks a lot!

PR #1707, opened 2023-10-13
Update readme for milestone-less releasing

EliahKagan:
This updates the release-making instructions in the readme, to release without using milestones and to link to the GitHub release page rather than a milestone, as discussed in #1706.

I've also made those instructions a numbered list, since they consist of sequential steps, and decreased the amount of elaboration in one of the other steps, which I think will make it so the new important details in the changelog-writing step are less likely to be glossed over.

In addition to the changes here, `check-version.sh` could and, I think, should be extended to check that the top section in `changes.rst` properly links to the (forthcoming) release page on GitHub. That link contains the version number (rather than a milestone number) so it makes sense for the version check script to validate it. Furthermore, because the page will typically not exist yet, mistakes in the URL may be harder to catch. However, I have not included such a change in this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,13 +199,12 @@ Please have a look at the [contributions file][contributing].
 ### How to make a new release
 
 - Update/verify the **version** in the `VERSION` file.
-- Update/verify that the `doc/source/changes.rst` changelog file was updated.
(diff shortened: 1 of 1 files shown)


Byron:
That's exactly like I imagined it, thanks so much!

It would definitely be good to have a `check-version.sh` update, but I also think it's entirely optional.

PR #1714, opened 2023-10-18
Add missing info in Submodule.remove docstring

### LINKED ISSUE #1712: Submodule.remove `module` parameter doc is unclear or incomplete
The docstring for [`Submodule.remove`](https://github.com/gitpython-developers/GitPython/blob/main/git/objects/submodule/base.py#L941) documents the `module` parameter this way:

https://github.com/gitpython-developers/GitPython/blob/6f765a2028e45a14853bf06801104cc6d22e053e/git/objects/submodule/base.py#L951-L959

A sentence trails off in "if you have modifications in the" and it appears material is missing. This should be revised, but it is is unclear exactly what the intended meaning is, so I don't know for sure how to fix it.

The current phrasing appears to be the effect of an unsuccessful or incomplete [attempt](https://github.com/gitpython-developers/GitPython/commit/8e0e315a371cdfc80993a1532f938d56ed7acee4#diff-0d77f7389366948adb9b48c092150ac5c03033de1d89250748ae9add1e548eb5R512-R519), in 8e0e315, to fix a similar misphrasing. Compared to the [original wording](https://github.com/gitpython-developers/GitPython/commit/3d061a1a506b71234f783628ba54a7bdf79bbce9#diff-b783f55c358d34650a259d030b6ed9b67b611d205ed239027c861cad71b67d93R337-R345) in 3d061a1, the change was:

```diff
         :param module: If True, the module we point to will be deleted 
             as well. If the module is currently on a commit which is not part 
             of any branch in the remote, if the currently checked out branch 
-            is ahead of its tracking branch,  if you have modifications in the
             working tree, or untracked files,
+            is ahead of its tracking branch,  if you have modifications in the
             In case the removal of the repository fails for these reasons, the 
             submodule status will not have been altered.
             If this submodule has child-modules on its own, these will be deleted
```

So the original wording also trailed off in mid-sentence, but differently, with "working tree, or untracked files,".

I can *guess* the intended meaning:

> If True, the module checkout we point to will be deleted as well. If the module is currently on a commit which is not part of any branch in the remote, if the currently checked out branch is ahead of its tracking branch, or if there are modifications to the working tree or untracked files, then the removal will fail.

If that revision is correct, it could be clarified by a small amount of further rewording:

> If True, the checked out module we point to will be deleted as well. If that module is currently on a commit outside any branch in the remote, or if it is ahead of its tracking branch, or if there are modified or untracked files in its working tree, then the removal will fail.

However, I am not sure enough that this is the intended meaning to have put either of those changes in a PR at this time.

EliahKagan:
Fixes #1712

This rewords and adds some missing information to the docstring of `Submodule.remove`, for the `method` parameter, discussed in #1712. It uses the second suggestion presented in that issue, formatted in the style of the surrounding docstring (an 88-column wrap).

It also does some other rewording in that docstring, for clarity, at the end of that same section (on the `method` parameter), and a small punctuation fix in the section about the `force` parameter.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -948,17 +948,16 @@ class Submodule(IndexObject, TraversableIterableObj):
         """Remove this submodule from the repository. This will remove our entry
         from the .gitmodules file and the entry in the .git / config file.
 
-        :param module: If True, the module checkout we point to will be deleted
-            as well. If the module is currently on a commit which is not part
-            of any branch in the remote, if the currently checked out branch
-            working tree, or untracked files,
-            is ahead of its tracking branch, if you have modifications in the
+        :param module: If True, the checked out module we point to will be deleted as
+            well.If that module is currently on a commit outside any branch in the
+            remote, or if it is ahead of its tracking branch, or if there are modified
+            or untracked files in its working tree, then the removal will fail.
             In case the removal of the repository fails for these reasons, the
             submodule status will not have been altered.
-            If this submodule has child - modules on its own, these will be deleted
-            prior to touching the own module.
+            If this submodule has child modules of its own, these will be deleted prior
+            to touching the direct submodule.
         :param force: Enforces the deletion of the module even though it contains
-            modifications. This basically enforces a brute - force file system based
+            modifications. This basically enforces a brute-force file system based
             deletion.
         :param configuration: if True, the submodule is deleted from the configuration,
             otherwise it isn't. Although this should be enabled most of the times,


Byron:
Thanks so much for your help, it's much appreciated!

I will happily *not* make the next release a "post" release as suggested (after [looking it up](https://stackoverflow.com/questions/45877885/post-release-versioning-and-semver-2-0-sematic-verisoning)), merely because I have never done it before and am afraid that this breaks anything or anyone downstream 😅.

EliahKagan:
That makes sense, and as I've just written about in #1716, I think there's another change that *might* make sense to include in 3.1.39, if the timing works out.

PR #1719, opened 2023-10-20
Add `__all__` in git.exc

### LINKED ISSUE #1718: The public interface of git.exc is ambiguous
I believe the intent of `git.exc` is twofold:

- To publish the exceptions it defines, which it unambiguously does.
- To republish exceptions defined in `gitdb.exc`, which it... *doesn't?*

The problem is that the conceptually public names in a module are (basically) those listed in `__all__` when `__all__` is present, and since an original and continuing use of `__all__` is to control what names a `*` import binds, it might seem like whatever a `*` import binds is considered conceptually public even without `__all__`... but that is not the case.

Instead, in the absence of `__all__`, names that both do not start with an `_` and also *were not introduced merely by themselves being imported* are the conceptually public names. As [the "Public and internal interfaces" section of PEP-8](https://pep8.org/#public-and-internal-interfaces) says:

> Imported names should always be considered an implementation detail. Other modules must not rely on indirect access to such imported names unless they are an explicitly documented part of the containing module’s API<em>[...]</em>

To make such a name public, it can be included in `__all__`, but `git.exc` currently does not define `__all__`. Does it otherwise document those names as public? Well...

https://github.com/gitpython-developers/GitPython/blob/44102f30eaadcd122899f5f801f28b83bd9a5111/git/exc.py#L6

This is why I say the situation is ambiguous--this might be interpreted to mean that the names are public, in that they are the names of exceptions that are thrown throughout GitPython and would not otherwise be public in GitPython. But really this is a stretch, and rather than saying the situation is ambiguous, it might be more reasonable to say that they are definitely not public.

But I am pretty sure they are intended to be public. So they should be documented as such. The best way to do this is by defining `git.exc.__all__` and listing them there.

There then becomes the question of whether anything else that is currently imported by `from git.exc import *` belongs in `__all__`. I believe the answer is no. It's not reasonable for code outside `git.exc` to rely on those being part of `git.exc` (because they are present only due to imports, so they are nonpublic per the above-quoted rule, and the documentation of `git.exc` does not reference them even obliquely). It is especially unreasonable for code outside the `git.exc` module to rely on receiving them from a `*` import.

However, the top-level `__init__.py` does just that. *When* it began to do it depends on one's perspective. For some time, it has been including those names (such as `safe_decode` from `git.util`) in `git.__all__` because of the way `git.__all__` was dynamically generated with a comprehension. In #1659, the comprehension was removed and replaced with a listing of everything that was found to be in `__all__`.

Because `git.__all__` did, and does, exist, and it included those names, it should continue to include them for backward compatibility, even in the case of some of them from `typing` that hopefully no one is importing from the `git` module. However, I think no similar reasoning applies to `git.exc` itself, since `git.exc.__all__` has not existed. This is to say that I think it is always a bug to rely on picking up nonpublic names from `*` imports, including in `git/__init__.py`, which is currently doing that.

To keep `from git import *` working, as well as to keep `git.__all__` a correct statement of names guaranteed accessible as attributes of `git`, when fixing this in `git.exc` it will also be necessary to modify the code of `git/__init__.py` to get these names from the modules that really provide them publicly. That should be no problem, though, and I'd say it would be an improvement even by itself.

I believe `git.exc` is the clearest place in GitPython where this kind of problem exists, but every attempt to suppress an unused import lint rule is a strong hint of a similar bug, so I'm pretty sure this is far from the only case. However, I'm opening an issue about this specifically because recent changes (in #1659) combine with it to create a situation where the bug could become entrenched if not addressed fairly soon. I say this because although `git/__init__.py` has, for quite some time, being wrongly relying on those names being present, in that it has been effectively guaranteeing their presence for future patch versions by including them in `__all__` as it would be inspected by users, it is only now that this dependence can be easily discerned, and thus perhaps further relied on, by reading the code.

EliahKagan:
Fixes #1718

This adds `__all__` in `git.exc`, and adjusts `__init__.py` imports.

The `git.exc` module imports exceptions from `gitdb.exc` to republish them, as well as defining its own (also for use from outside). But because it did not define `__all__`, the intent for the exceptions it imported was unclear, since names that are introduced by imports and not present in `__all__` are not generally considered public, even when `__all__` is absent and a `*` import would reimport them.

This rectifies that by adding `__all__` and listing both imported and newly introduced exceptions explicitly in it. Although this strictly expands which names are public under typical conventions, it strictly contracts which names are imported by a `*` import, because the presence of `__all__` suppresses names not listed in it from being imported that way. However, because under typical conventions those other names are not considered public, and they were not even weakly documented as public, this should be okay.

(Even though this is not a breaking change, in that code it would break would already technically be broken... if it turns out that it is common to wrongly rely on the availability of those names, then this may need to be revisited and slightly modified.)

This brings the readily identified public interface of `git.exc` in line with what is weakly implied (and intended) by its docstring.

This also modifies `__init__.py` accordingly: The top-level git module has for some time used a `*` import on `git.exc`, causing the extra names originally meant as implementation details to be included. Because its own `__all__` was dynamically generated until c862845, #1659 also added 8edc53b to retain the formerly present names in `__all__`. So the change here imports those names from the modules that deliberately provide them, to preserve compatibility.

### DIFF
--- a/git/exc.py
+++ b/git/exc.py
@@ -5,7 +5,34 @@
 # the BSD License: https://opensource.org/license/bsd-3-clause/
 """ Module containing all exceptions thrown throughout the git package """
 
-from gitdb.exc import (  # noqa: @UnusedImport
+__all__ = [
+    # Defined in gitdb.exc:
+    "AmbiguousObjectName",
+    "BadName",
+    "BadObject",
+    "BadObjectType",
+    "InvalidDBRoot",
+    "ODBError",
+    "ParseError",
+    "UnsupportedOperation",
+    # Introduced in this module:
+    "GitError",
+    "InvalidGitRepositoryError",
+    "WorkTreeRepositoryUnsupported",
+    "NoSuchPathError",
+    "UnsafeProtocolError",
+    "UnsafeOptionError",
+    "CommandError",
+    "GitCommandNotFound",
+    "GitCommandError",
+    "CheckoutError",
+    "CacheError",
+    "UnmergedEntriesError",
+    "HookExecutionError",
+    "RepositoryDirtyError",
+]
+
+from gitdb.exc import (
     AmbiguousObjectName,
     BadName,
     BadObject,
@@ -14,7 +41,6 @@ from gitdb.exc import (  # noqa: @UnusedImport
     ODBError,
     ParseError,
     UnsupportedOperation,
-    to_hex_sha,
 )
 from git.compat import safe_decode
 from git.util import remove_password_if_present
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -6,7 +6,6 @@
 # flake8: noqa
 # @PydevCodeAnalysisIgnore
 from git.exc import *  # @NoMove @IgnorePep8
-import inspect
 import os
 import sys
 import os.path as osp
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -7,10 +7,10 @@
 # @PydevCodeAnalysisIgnore
 from git.exc import *  # @NoMove @IgnorePep8
 import os
-import sys
 import os.path as osp
+import sys
 
-from typing import Optional
+from typing import List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
 from git.types import PathLike
 
 __version__ = "git"
@@ -38,7 +38,10 @@ _init_externals()
 
 # { Imports
 
+from gitdb.util import to_hex_sha
+
 try:
+    from git.compat import safe_decode  # @NoMove @IgnorePep8
     from git.config import GitConfigParser  # @NoMove @IgnorePep8
     from git.objects import *  # @NoMove @IgnorePep8
     from git.refs import *  # @NoMove @IgnorePep8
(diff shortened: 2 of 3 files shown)


git/exc.py
```diff
@@ -5,7 +5,34 @@
...
+    "UnsafeProtocolError",
+    "UnsafeOptionError",
+    "CommandError",
+    "GitCommandNotFound",
+    "GitCommandError",
+    "CheckoutError",
+    "CacheError",
+    "UnmergedEntriesError",
+    "HookExecutionError",
+    "RepositoryDirtyError",
```
EliahKagan, line 32: Maybe these should be alphabetized? But it was easier for me to verify that they were all there when listed in the same order they are defined in this same module, and I suspect others reading and working with the code may find this easier to verify as well, so I've kept them in this order.

Byron: Both orders have their value, and I absolutely leave it to you to decide which one to prefer. Let's keep it the way it is for now then.

git/exc.py
```diff
@@ -5,7 +5,34 @@
 # the BSD License: https://opensource.org/license/bsd-3-clause/
 """ Module containing all exceptions thrown throughout the git package """
 
-from gitdb.exc import (  # noqa: @UnusedImport
+__all__ = [
+    # Defined in gitdb.exc:
```
Byron, line 9: This partitioning is very helpful.

Byron (approved):
Thanks a lot for your help!

PR #1725, opened 2023-10-30
Revise comments, docstrings, some messages, and a bit of code

EliahKagan:
This revises comments and docstrings, to improve clarity, readability, stylistic consistency, in some cases accuracy, and (for docstrings) Sphinx formatting and cross-linking. It also contains a few other related changes.

### Short summary of the most notable changes

In both comments and docstrings, my changes included rewording for clarity, as well as changes in capitalization, punctuation, and so forth.

For comments, which were in a variety of styles, I converted them to be sentences (or in some cases fragments formatted as sentences), except in a minority of cases when doing so seemed like it would worsen readability. I did this because, in addition to improving stylistic consistency and being encouraged by PEP-8, this tended to make their structure clearer (especially in cases of comments over multiple lines that were convertible to multiple separate sentences). This comment style was also one of the styles already represented throughout the codebase.

For docstrings, there were a number of places where the formatting looked one way when reading the raw docstring, but very different, and arguably less readable, as rendered by Sphinx. In cases where this was clearly unintended and the effect was significant, such as markup that was intended to represent an unordered list but instead represented a running paragraph with literal asterisks, I fixed the formatting. In cases where the intent was less clear, such as non-wrapping newlines conveying a break more major than that between sentences but less major than that between paragraphs, I preserved the style where possible but did not make changes to cause it to be newly reflected in rendered documentation.

### Kinds of other changes

There are also some other changes, described below, including some *code changes*, mainly for clarity, that I believe to be sufficiently related or where there is otherwise a reason to include them here rather than proposing them separately.

Some specific changes (*not* to code) are described in commit messages but not this PR description, but this is noted below. Specifically, a handful of changes to docstrings are for accuracy, and to allow them to carry specific rationales and to be reviewed individually, they each have individual commits and are not currently relisted here.

### Line length in docstrings

I found that many docstrings were already written, or almost entirely written, with a hard wrap of 88 columns. It makes sense that this would be done, because it is the `black` default, even though `black` does not enforce line length in docstrings, and furthermore the maximum line length this this project has been set at the much higher 120 columns. I think that a width of 120, or even 100, in docstrings, reduces readability when reading the docstrings in the code (of course it does not affect what Sphinx renders), so this is probably another reason a smaller width was used in most cases.

[...]

PR #1726, opened 2023-11-02
Use zero-argument super()

EliahKagan:
This replaces 2-argument `super(...)` calls with the zero-argument `super()` form everywhere applicable:

- All uses of two-argument `super` in the *code* are replaced.
- All but one mention of two-argument `super` in *comments* are replaced.

I had omitted this from #1725 because it may require more technical attention to fully check than other code changes there, and that PR was mostly for comments and docstrings.

### Exactly when `super()` can be used

*Since `super` is in practice one of the most confusing features of Python, I've included this section to help in verifying the changes here, and also so that they could be reverified later. However, this section is optional.*

As noted in output of `help(super)` (which is from [this code](https://github.com/python/cpython/blob/f4b5588bde656d8ad048b66a0be4cb5131f0d83f/Objects/typeobject.c#L10581), though [the other documentation](https://docs.python.org/3/library/functions.html#super) can be useful):

```text
super() -> same as super(__class__, <first argument>)
```

`__class__` is [actually literal](https://docs.python.org/3/reference/datamodel.html#creating-the-class-object) there--though the static `__class__` cell should not be confused with the dynamic `__class__` attribute--but it can also be thought of as standing for the name of the enclosing class (or narrowest enclosing class in the case of nested classes), which is actually what one typically writes in the two-argument form of `super`.

When the expression passed in place of `__class__` refers to the (narrowest enclosing) class, and the expression passed in place of `<first argument>` is the name of the first parameter of the method (typically `self` in instance methods, or `cls` in class methods or `__new__`), the zero-argument form `super()` can be used instead with the same meaning.

### The one place I couldn't replace it

The above-described situation applied *everywhere* in GitPython where `super` was actually *used*.

However, there are also comments in some places describing how something else could be done or how what was done could be done differently. In those, *most* showings of multiple-argument `super` were replaceable with `super` without changing what the code would mean and do if used in the place the comment pertained to. But then there is this, in `git.objects.util.TraversableIterableObj`:

https://github.com/gitpython-developers/GitPython/blob/6cef9c054fa15c711aed83cd5dfb19b4520e75f8/git/objects/util.py#L625-L658

The actual `super` call in the code there is replaceable, and I replaced it, with zero-argument `super()`. But the comment--or, more precisely, the string being used as a comment--shows a call with `Commit` as the class name. I'm not sure if that's intended or not. If it is intended, maybe it represents code that would be written somewhere else. So I haven't changed that, though I'm willing to do so if the matter is clarified. It could be changed either in a new commit in this PR, or in a subsequent PR.

### Other changes

There's some proofreading of the kind done in #1725 that I missed there and took care of here.

More significantly, `TestBigRepoR.setUp` contains:

https://github.com/gitpython-developers/GitPython/blob/6cef9c054fa15c711aed83cd5dfb19b4520e75f8/test/performance/lib.py#L35-L38

And `TestBigRepoRW.setUp` analogously contains:

https://github.com/gitpython-developers/GitPython/blob/6cef9c054fa15c711aed83cd5dfb19b4520e75f8/test/performance/lib.py#L67-L70

In addition to replacing those two-argument `super` calls with zero-argument `super()`, I also removed the logic to catch and swallow `AttributeError`. As covered in more detail in the 9113177 commit message, this is unnecessary because `unittest.TestCase` defines an empty `setUp` method to allow this kind of special casing to be avoided, and it may be brittle because `AttributeError` could be raised from the actual code of a `setUp` method, due to the specific way the code to swallow it is written. However, if that latter effect, which would usually be strongly unwanted, is actually the goal, then this should be undone and a suitable comment and/or clarifying refactoring done.

### What isn't included

It's unclear whether `super` calls in test classes' `setUp` and `tearDown` methods are always made in the places they are made because that is where superclass and sibling-class fixture acquisition and release behavior should occur. For example, usually, if `setUp` delegates via `super` to another class's `setUp` method at the beginning, and in a case where the corresponding `tearDown` method needs delegate via `super` to another class's `tearDown` method, then that `tearDown` call should be at the *end*, so that resource acquisition and release more often occurs in a temporally nested way that is easier to reason about and more often correct.

However, sometimes doing it differently is deliberate. Furthermore, `setUp` and `tearDown` methods are inherently weird, because, if a subclass's `setUp` calls a superclass's `setUp` and then fails with an exception, then `tearDown` won't run. (Also, in situations where `tearDown` does run, an exception in one `tearDown` method may still prevent other MRO classes' `tearDown` methods from running. Unlike the first issue, for this there isn't really any perfect thing to do. But I think the framework still makes it somewhat harder than necessary to reason about when the `tearDown` calls are manually written rather than arranged and implied by the `setUp` logic.)

For this reason, I am unsure if it is really worthwhile to try to improve ordering. If improvements are to be made, it might be best to have only `setUp` methods and have them register cleanup logic by calling [`addCleanup`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup)--or, perhaps better where applicable, [`enterContext`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.enterContext), though a backport of it for Python <3.11 would be needed. Unlike `tearDown` methods, cleanup scheduled with `addCleanup`, either directly or through `enterContext`, is deterministically performed, in the opposite of the order it was added, even if `setUp` failed with an exception after the `addCleanup` call.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -718,7 +718,7 @@ class Git(LazyMixin):
            It is meant to be the working tree directory if available, or the
            ``.git`` directory in case of bare repositories.
         """
-        super(Git, self).__init__()
+        super().__init__()
         self._working_dir = expand_path(working_dir)
         self._git_options: Union[List[str], Tuple[str, ...]] = ()
         self._persistent_git_options: List[str] = []
@@ -765,7 +765,7 @@ class Git(LazyMixin):
                 tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
             )
         else:
-            super(Git, self)._set_cache_(attr)
+            super()._set_cache_(attr)
         # END handle version info
 
     @property
--- a/git/config.py
+++ b/git/config.py
@@ -107,7 +107,7 @@ class MetaParserBuilder(abc.ABCMeta):  # noqa: B024
             # END for each base
         # END if mutating methods configuration is set
 
-        new_type = super(MetaParserBuilder, cls).__new__(cls, name, bases, clsdict)
+        new_type = super().__new__(cls, name, bases, clsdict)
         return new_type
 
 
@@ -178,7 +178,7 @@ class SectionConstraint(Generic[T_ConfigParser]):
     def __getattr__(self, attr: str) -> Any:
         if attr in self._valid_attrs_:
             return lambda *args, **kwargs: self._call_config(attr, *args, **kwargs)
-        return super(SectionConstraint, self).__getattribute__(attr)
+        return super().__getattribute__(attr)
 
     def _call_config(self, method: str, *args: Any, **kwargs: Any) -> Any:
         """Call the configuration at the given method which must take a section name
@@ -207,36 +207,36 @@ class _OMD(OrderedDict_OMD):
     """Ordered multi-dict."""
 
     def __setitem__(self, key: str, value: _T) -> None:
-        super(_OMD, self).__setitem__(key, [value])
+        super().__setitem__(key, [value])
 
     def add(self, key: str, value: Any) -> None:
         if key not in self:
-            super(_OMD, self).__setitem__(key, [value])
+            super().__setitem__(key, [value])
             return None
-        super(_OMD, self).__getitem__(key).append(value)
+        super().__getitem__(key).append(value)
 
     def setall(self, key: str, values: List[_T]) -> None:
-        super(_OMD, self).__setitem__(key, values)
+        super().__setitem__(key, values)
 
     def __getitem__(self, key: str) -> Any:
-        return super(_OMD, self).__getitem__(key)[-1]
+        return super().__getitem__(key)[-1]
 
     def getlast(self, key: str) -> Any:
-        return super(_OMD, self).__getitem__(key)[-1]
+        return super().__getitem__(key)[-1]
 
     def setlast(self, key: str, value: Any) -> None:
         if key not in self:
-            super(_OMD, self).__setitem__(key, [value])
+            super().__setitem__(key, [value])
             return
 
-        prior = super(_OMD, self).__getitem__(key)
+        prior = super().__getitem__(key)
         prior[-1] = value
 
     def get(self, key: str, default: Union[_T, None] = None) -> Union[_T, None]:
-        return super(_OMD, self).get(key, [default])[-1]
+        return super().get(key, [default])[-1]
 
     def getall(self, key: str) -> List[_T]:
-        return super(_OMD, self).__getitem__(key)
+        return super().__getitem__(key)
 
     def items(self) -> List[Tuple[str, _T]]:  # type: ignore[override]
         """List of (key, last value for key)."""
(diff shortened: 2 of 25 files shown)


Byron (approved):
Thanks so much for working your way through this codebase of nearly 26k lines of code! I can't thank you enough.

>  If it is intended, maybe it represents code that would be written somewhere else. So I haven't changed that, though I'm willing to do so if the matter is clarified. It could be changed either in a new commit in this PR, or in a subsequent PR.

I definitely don't remember what the intention originally was, but I can absolutely believe that GitPython tried to be very clever with inheritance. Thus I suggest to find a test for this case or see if adding one yields desirable results. From there it should be clear what can be done (which might as well involve changing the doc-string).

> For this reason, I am unsure if it is really worthwhile to try to improve ordering. If improvements are to be made, it might be best to have only setUp methods and have them register cleanup logic by calling [addCleanup](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup)--or, perhaps better where applicable, [enterContext](https://docs.python.org/3/library/unittest.html#unittest.TestCase.enterContext), though a backport of it for Python <3.11 would be needed. Unlike tearDown methods, cleanup scheduled with addCleanup, either directly or through enterContext, is deterministically performed, and in the opposite of the order it was added.

To me this reads like the way `super(…)` is used here can stay until there is an incentive to use `addCleanup()`, which seems like a superior way of doing this along with the guarantee that the tear-down will happen as well.

EliahKagan:
> Thanks so much for working your way through this codebase of nearly 26k lines of code! I can't thank you enough.

You're welcome! *but* I feel this praise is somewhat undeserved. As noted in https://github.com/gitpython-developers/GitPython/pull/1725#issuecomment-1791341169, the way I've taken in the code is closer to skimming the surface than a deep dive. Over time, however, I expect my knowledge and understanding of the code to deepen further.

For the change here in this PR, broad familiarity with the codebase was less relevant than in #1725, because the code that needed to be checked to verify correctness was always fairly nearby: I just needed to check, in each occurrence of `super` passing two arguments, that it occurred in a method, that the first argument was the name of the (nearest enclosing) class, and that the second argument was the method's first parameter.

> To me this reads like the way `super(…)` is used here can stay until there is an incentive to use `addCleanup()`, which seems like a superior way of doing this along with the guarantee that the tear-down will happen as well.

Another option, which would sort of make this irrelevant, would be to convert all the tests to be pure `pytest` tests. Most tests in the project use `unittest`, being implemented as methods of classes that inherit (indirectly) from `unittest.TestCase`. A small number are pure `pytest` tests, but not most, and currently none use `pytest` fixtures. `pytest` fixtures. [`pytest` fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) take care of cleanup as well as `unittest.TestCase.addCleanup`. A `pytest` fixture that requires cleanup can ensure it is performed by either *(a)* itself depending on another fixture that provides it, or *(b)* by being implemented as a generator that does cleanup after the `yield` (sort of like how one writes a `@contextlib.contextmanager`-decorated function, except without writing the `try`-`finally` oneself).

I already plan to make some of the tests in `test_util.py` pure `pytest` tests, so that the xfail markings will be granular, eliminating all XPASSes and making immediately clear from the test code itself exactly which parameter sets produce failure on which tests. (This is what I meant in #1725 under "What is omitted".) I would definitely *not* try to shoehorn a big design change affecting the entire test suite into that change. Nor do I have a drive to break things that are already working well. However, perhaps over time complete conversion of the test suite to pure `pytest` tests, with the [further `pytest` features](https://docs.pytest.org/en/7.1.x/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses) that this facilitates, would become natural.

Byron:
> However, perhaps over time complete conversion of the test suite to pure `pytest` tests, with the [further `pytest` features](https://docs.pytest.org/en/7.1.x/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses) that this facilitates, would become natural

Yes, that sounds like the way to go. Keep working what's working already, and if opportunities arise to improve it, probably by using `pytest` instead, that can be done. 
Since you are working with the code the most these days, it's really about what seems most useful to you.

PR #1729, opened 2023-11-03
Reorganize test_util and make xfail marks precise

### LINKED ISSUE #1728: Overbroad xfail marks will eventually make CI fail
### Background

#1679 included improvements to a number of tests that are known to fail on some platforms, by marking them `xfail` instead of `skip` so they are still run and their status is reported, but without a failing status causing the whole test run to fail. However, it applied [`xfail`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) to too many tests, due to limitations on granularity when applying [`pytest` marks](https://docs.pytest.org/en/latest/how-to/mark.html) to `unittest` test cases generated by `@ddt` parameterization.

https://github.com/gitpython-developers/GitPython/blob/340da6d39397253da8d0807179b0ecb5952effca/test/test_util.py#L221-L228

https://github.com/gitpython-developers/GitPython/blob/340da6d39397253da8d0807179b0ecb5952effca/test/test_util.py#L233-L245

### Upcoming impact

Although this was known and discussed in #1679, and FIXME comments about it were included in the code, the problem turns out to be somewhat more serious than I had anticipated: if not addressed, it will eventually lead to test failures in a future version of `pytest`. This is because the default behavior of an *unexpectedly passing* test--one that is marked `xfail` but passes--will most likely change in pytest 8. Because GitPython does not specify upper bounds on most of its development dependencies, and pytest is one of the development dependencies for which no upper bound is specified, pytest 8 will be automatically installed once it is (stably) released.

 Specifically, and in the absence of configuration or command-line options to `pytest` that override the behavior:

- A test marked `xfail` that fails, and fails in the expected way, produces an XFAIL status, which is treated similarly to PASS. We always want this.
- A test marked `xfail` that fails in a detectably unexpected way--where a different exception results than the one that was [expected](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#raises-parameter)--produces a FAIL status. We always want this.
- A test marked `xfail` that *passes* produces an XPASS status. How this status is treated is more complicated. The `xfail` mark supports [an optional `strict` parameter](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter). Where present, it determines whether the XPASS fails the test run like a FAIL status would, or does not fail the test run (thus behaving like PASS or XFAIL). If absent, the `xfail_strict` configuration option provides the default. Currently, as of pytest 7, `xfail_strict` defaults to `False` when not specified.

As noted in https://github.com/pytest-dev/pytest/issues/11467, which was opened by a pytest maintainer and is listed for pytest's [8.0 milestone](https://github.com/pytest-dev/pytest/milestone/35), the default is planned to be changed from `False` to `True` starting in pytest 8.0. (See also https://github.com/pytest-dev/pytest/pull/11499.)

### Possible fixes

Breakage could be avoided (at least for a while, since `strict=False` [may eventually](https://github.com/pytest-dev/pytest/issues/11467#issuecomment-1733852526) be removed as a feature) by passing `strict=False` or setting `xfail_strict=false` for `pytest` in `pyproject.toml`. It is also possible to set an upper bound like `<8` for `pytest` in `test-requirements.txt`.

However, I recommend this instead be fixed by reorganizing the tests in `test_util.py` so that the tests of `cygpath` and `decygpath`--which are the ones that have the insufficiently precise `xfail` markings that mark some generated test cases `xfail` even though they are known to pass--can be pure `pytest` tests. Because they are currently `unittest` tests, they cannot be generated by `@pytest.mark.parametrize` (hence `@ddt` is used). But if they could be generated with the `parametrize` mark then they could have per-case markings, because `parametrize` supports an optional `marks` argument. They could then have the `xfail` mark applied to exactly the cases where failure is really expected.

That approach – which I mentioned in #1679 itself and in https://github.com/gitpython-developers/GitPython/pull/1700#discussion_r1353654395, and more recently alluded to in #1725 and https://github.com/gitpython-developers/GitPython/pull/1726#issuecomment-1791541248 – has the following advantages over other approaches that effectively just suppress the problem:

- Any XPASS will be a sign that something has changed and should be looked into, thereby building on the improvements in [#1679](https://github.com/gitpython-developers/GitPython/pull/1679).
- Although we have FIXME comments, the current situation is still misleading in the test results themselves, which indicate that some tests are unexpectedly passing.
- When the default treatment of XPASS in `pytest` changes--but also even before that, once it is documented to change--the presence of expected XPASSes will be more misleading than it is already, even if GitPython is not using a version of `pytest` affected by the change. This is because that change will further solidify people's expectations about what XPASS indicates, including for people who are trying to become familiar with GitPython.
- Reorganizing the tests in `test_util.py` can also help clarify the tests of `rmtree` behavior, and help make them easier to modify. This is useful because it will allow building on [#1700](https://github.com/gitpython-developers/GitPython/pull/1700) toward an eventual complete fix for [#790](https://github.com/gitpython-developers/GitPython/issues/790). (In addition, I want to make sure the [planned](https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717239735) native Windows CI jobs don't have the effect of calcifying cleanup logic in `rmtree` that otherwise could or should change, or at least that this does not happen in ways that impinge on non-Windows platforms. I think such a reorganization will help with that, too.)

I have opened #1729, which fixes this issue by reorganizing tests in `test_util.py` in this way.

EliahKagan:
Fixes #1728

This makes `xfail` markings precise in the test suite, so that only tests that are expected to fail (and do fail on CI) are marked `xfail`. This addresses both the general problem that spurious XPASS statuses make the information from `xfail`-marked tests less informative, and the specific problem focused on in #1728.

In addition, this also refactors some other recently added tests in `test_util.py`, to achieve the goals described there. I *may* soon propose other changes that build on this in other ways, but that is not included here. Basically, this just:

- Reorganizes the tests in `test_util.py`, making some of them pure `pytest` tests.
- Uses `@pytest.mark.parametrize` with the `marks=` argument to apply `@pytest.mark.xfail` to exactly the cases that currently need it, and no others.
- Refactors some of the other tests' use of ad-hoc fixtures, removing some code reuse that I had added that was making the code harder rather than easier to maintain, and achieving useful code reuse more cleanly with a custom `pytest` fixture.

This includes two changes to the `pytest` plugins that are used. Since those affect development dependencies, I list them here with their rationales:

- Because the `parametrized` mark supports nesting, generating test cases for the Cartesian product of the specified parameters, I removed the use of subtests for now, and thus also the [`pytest-subtests`](https://pypi.org/project/pytest-subtests/) development dependency. Subtests, and that plugin, could be brought back in the future if they prove useful for another reason, but right now there are no longer any subtests in the test suite. This is the change I had anticipated in [#1700 (comment)](https://github.com/gitpython-developers/GitPython/pull/1700#discussion_r1353654395).
- I have added and used [`pytest-mock`](https://pypi.org/project/pytest-mock/) for some of the new pure `pytest` tests that had previously used nested `with`-statements. The `pytest-mock` plugin provides a `mocker` fixture that wraps the functionality of `unittest.mock` but automatically unpatches everything (and in the reverse order in which it was patched) after the test.

Since we do not yet have CI for native Windows, I have manually run the `test_util.py` tests on Windows (on Windows 10, with Python 3.12) to verify that they produce the expected statuses (including no FAIL statuses).

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -211,6 +211,11 @@ class TestEnvParsing:
         assert actual_parsed_value is expected_truth_value
 
 
+def _xfail_param(*values, **xfail_kwargs):
+    """Build a pytest.mark.parametrize parameter that carries an xfail mark."""
+    return pytest.param(*values, marks=pytest.mark.xfail(**xfail_kwargs))
+
+
 @pytest.mark.skipif(sys.platform != "cygwin", reason="Paths specifically for Cygwin.")
 class TestCygpath:
     """Tests for :func:`git.util.cygpath` and :func:`git.util.decygpath`."""
@@ -246,15 +251,11 @@ class TestCygpath:
         cwpath = cygpath(wpath)
         assert cwpath == cpath, wpath
 
-    @pytest.mark.xfail(
-        reason=R'2nd example r".\bar" -> "bar" fails, returns "./bar"',
-        raises=AssertionError,
-    )
     @pytest.mark.parametrize(
         "wpath, cpath",
         [
             (R"./bar", "bar"),
-            (R".\bar", "bar"),  # FIXME: Mark only this one xfail (or fix it).
+            _xfail_param(R".\bar", "bar", reason=R'Returns: "./bar"', raises=AssertionError),
             (R"../bar", "../bar"),
             (R"..\bar", "../bar"),
             (R"../bar/.\foo/../chu", "../bar/chu"),
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -49,8 +49,13 @@ def permission_error_tmpdir(tmp_path):
     td = tmp_path / "testdir"
     td.mkdir()
     (td / "x").write_bytes(b"")
-    (td / "x").chmod(stat.S_IRUSR)  # Set up PermissionError on Windows.
-    td.chmod(stat.S_IRUSR | stat.S_IXUSR)  # Set up PermissionError on Unix.
+
+    # Set up PermissionError on Windows, where we can't delete read-only files.
+    (td / "x").chmod(stat.S_IRUSR)
+
+    # Set up PermissionError on Unix, where we can't delete files in read-only directories.
+    td.chmod(stat.S_IRUSR | stat.S_IXUSR)
+
     yield td
 
 
@@ -113,7 +118,7 @@ class TestRmtree:
         # git.index.util "replaces" git.util and is what "import git.util" gives us.
         mocker.patch.object(sys.modules["git.util"], "HIDE_WINDOWS_KNOWN_ERRORS", True)
 
-        # Disable common chmod functions so the callback can't fix the problem.
+        # Disable common chmod functions so the callback can never fix the problem.
         mocker.patch.object(os, "chmod")
         mocker.patch.object(pathlib.Path, "chmod")
 
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -59,12 +59,6 @@ def permission_error_tmpdir(tmp_path):
     yield td
 
 
-@pytest.fixture
-def file_not_found_tmpdir(tmp_path):
-    """Fixture to test errors deleting a directory that are not due to permissions."""
-    yield tmp_path / "testdir"  # It is deliberately never created.
-
-
 class TestRmtree:
     """Tests for :func:`git.util.rmtree`."""
 
@@ -142,7 +136,9 @@ class TestRmtree:
                 pytest.fail(f"rmtree unexpectedly attempts skip: {ex!r}")
 
     @pytest.mark.parametrize("hide_windows_known_errors", [False, True])
-    def test_does_not_wrap_other_errors(self, mocker, file_not_found_tmpdir, hide_windows_known_errors):
+    def test_does_not_wrap_other_errors(self, tmp_path, mocker, hide_windows_known_errors):
+        file_not_found_tmpdir = tmp_path / "testdir"  # It is deliberately never created.
+
         # See comments in test_wraps_perm_error_if_enabled for details about patching.
         mocker.patch.object(sys.modules["git.util"], "HIDE_WINDOWS_KNOWN_ERRORS", hide_windows_known_errors)
         mocker.patch.object(os, "chmod")
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -216,37 +216,59 @@ def _xfail_param(*values, **xfail_kwargs):
     return pytest.param(*values, marks=pytest.mark.xfail(**xfail_kwargs))
 
 
+_norm_cygpath_pairs = (
+    (R"foo\bar", "foo/bar"),
+    (R"foo/bar", "foo/bar"),
+    (R"C:\Users", "/cygdrive/c/Users"),
+    (R"C:\d/e", "/cygdrive/c/d/e"),
+    ("C:\\", "/cygdrive/c/"),
+    (R"\\server\C$\Users", "//server/C$/Users"),
+    (R"\\server\C$", "//server/C$"),
+    ("\\\\server\\c$\\", "//server/c$/"),
+    (R"\\server\BAR/", "//server/BAR/"),
+    (R"D:/Apps", "/cygdrive/d/Apps"),
+    (R"D:/Apps\fOO", "/cygdrive/d/Apps/fOO"),
+    (R"D:\Apps/123", "/cygdrive/d/Apps/123"),
+)
+
+_unc_cygpath_pairs = (
+    (R"\\?\a:\com", "/cygdrive/a/com"),
+    (R"\\?\a:/com", "/cygdrive/a/com"),
+    (R"\\?\UNC\server\D$\Apps", "//server/D$/Apps"),
+)
+
+# Mapping of expected failures for the test_cygpath_ok test.
+_cygpath_ok_xfails = {
+    # From _norm_cygpath_pairs:
+    (R"C:\Users", "/cygdrive/c/Users"): "/proc/cygdrive/c/Users",
+    (R"C:\d/e", "/cygdrive/c/d/e"): "/proc/cygdrive/c/d/e",
+    ("C:\\", "/cygdrive/c/"): "/proc/cygdrive/c/",
+    (R"\\server\BAR/", "//server/BAR/"): "//server/BAR",
+    (R"D:/Apps", "/cygdrive/d/Apps"): "/proc/cygdrive/d/Apps",
+    (R"D:/Apps\fOO", "/cygdrive/d/Apps/fOO"): "/proc/cygdrive/d/Apps/fOO",
+    (R"D:\Apps/123", "/cygdrive/d/Apps/123"): "/proc/cygdrive/d/Apps/123",
+    # From _unc_cygpath_pairs:
+    (R"\\?\a:\com", "/cygdrive/a/com"): "/proc/cygdrive/a/com",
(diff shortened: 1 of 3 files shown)


test/test_util.py
```diff
@@ -44,140 +43,142 @@
 from test.lib import TestBase, with_rw_repo
 
 
-class _Member:
-    """A member of an IterableList."""
+@pytest.fixture
+def permission_error_tmpdir(tmp_path):
+    """Fixture to test permissions errors in situations where they are not overcome."""
+    td = tmp_path / "testdir"
```
Byron, line 49: I keep choking on this as it seems more fragile than it has to be given that one could also create a tempdir with a unique name as well.

Byron (approved):
This looks so much cleaner, thanks so much!

Please note that my one comment merely is a note and nothing more, as I understand this code as always been like that and worked fine in single-threaded test-runners, which is all it has to be able to do.

test/test_util.py
EliahKagan: Can you tell me more about this concern? I'll be working on `test_util.py` again soon, so I'll have a good opportunity to make further improvements. However, [`tmp_path`](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) is already always unique: it's a `Path` object for a path like `/tmp/hw9aphg`, not `/tmp` itself.

Byron: Ah, perfect, then there is no concern, I just wasn't aware and too overwhelmed/not very focussed at the time.

EliahKagan: > then there is no concern

Ah, excellent.

Some of the code this PR modifies are the tests I introduced it in #1700. Both before and after the changes here, these tests may give the mistaken impression that variables like `td` refer to objects representing paths like `/tmp/tmpdir` that are always the same, rather than paths like `/tmp/hw9aphg/tmpdir` (or, with `pytest`, a more systematic scheme but still one that makes sure to use an initially nonexistent location).

After all, if `tmp_path` really corresponds to a unique and newly created directory, why not use *it* rather than creating a `tmp_path / "testdir"` subdirectory?

I do this for several reasons:

- I think of `tmp_path` as a resource the testing framework and/or custom fixtures manage: it is created before the test and (where appropriate and feasible) removed after the test. So for testing the behavior of `util.rmtree` in deleting a directory, I make a subdirectory for that, even if I wouldn't have to.
- Currently, I have configured `pytest` not to attempt to delete the temporary directory it creates afterwards, if the test failed. This allows the directory to be examined to troubleshoot a regression that causes a failure. But this would be confusing to actually do, if there is confusion between the `pytest`-managed temporary directory that may be deleted if the test succeeds, and the directory that the code under test attempts to delete.
- In at least test, conceptually zero directories are created, and the purpose of the temporary directory is to ensure that it *doesn't* have an entry of a particular name in it, so that this nonexistent entry can itself be used, to test what happens when the code under test raises a `FileNotFoundError`.
- In some forthcoming tests I plan to propose soon (but that I've had in mind for a short while, including when working on this PR), two directories will be created. Making them subdirectories of the `tmp_path` directory will make the test code a bit simpler, as I won't have to use a more versatile but complex fixture to create multiple `pytest`-managed temporary directories. More importantly, it will allow the two directories being affected by `util.rmtree` to be distinguished from each other when examining the tests and their results, rather than having them both have meaningless names that change each time.

This may be insufficiently clear in reading the test code, and if so, I would be interested in improving the situation, perhaps by changing some variable names, adding/expanding some docstrings or comments, or both. But maybe that wouldn't help; I'm not sure.

(Regarding the broader issue of test-runner parallelism, I'll post a separate top-level comment below.)

EliahKagan:
I'm curious if there are any areas where the test suite is written in a way that prevent parallelism that would otherwise be feasible and desirable. If so, I would be interested in trying to remedy that.

Some of the use of temporary files is done with insecure deprecated functions, which would best be replaced, as noted in https://github.com/gitpython-developers/smmap/issues/41. That issue is in smmap, but GitPython and gitdb also contain such code. Although these are subject to a race condition, it is *mostly* an issue between the process that uses the insecure function and other processes running other code, rather than in the affected code performing multithreading or multiprocessing. The same temporary paths are not automatically reused, except by strange chance or deliberate arrangement; new pseudorandom names for temporary files and directories are still generated. The biggest reason it can be a problem is security-related.

The facility that was previously used in the specific test code changed in this PR was [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) from the Python standard library. Like pytest's `tmp_path`, this is free of that security weakness, and like all the techniques for getting usable temporary file and directory paths discussed here, it gives a new path and can be used in parallel (including, but not limited to, across threads of the same process).

I'm aware of two factors that limit *GitPython itself* in multithreaded use. The first is that it changes directory temporarily to do some operations, as reported in #1052. The second is that the fix I made for #1635 (in #1636, and improved in #1650 to fix #1646) creates a race condition on `os.environ`, as pointed out in https://github.com/gitpython-developers/GitPython/pull/1650#issuecomment-1710089136 (see also https://github.com/gitpython-developers/GitPython/pull/1650#issuecomment-1710139306). So that the tests *also* change directory and patch `os.environ` doesn't seem like a problem.

The test suite does also do various monkey-patching that would cause problems if tests were run using multithreading, and this goes beyond any problem that I am aware of already existing in the code under test. It is possible to attempt to run tests using multithreading, and I think there are plugins for doing that, though I'm not familiar with the details of any of them. What I think is more common, and more useful for Python tests, is to run them using *multiprocessing*, such as with [`pytest-xdist`](https://pypi.org/project/pytest-xdist/). This retains isolation across tests, so it should work fine even when tests change directory, mutate `os.environ`, and monkey-patch GitPython and standard library facilities. Multiprocessing is also typically faster than multithreading in Python, except for code that is IO-bound or that delegates CPU-bound operations to native-code extension modules that release the GIL, though this situation will [eventually improve](https://peps.python.org/pep-0703/).

If the same temporary directory/files are reused in a way that can clash, or if code is taking what amounts to a lock on `rorepo`'s shared backing repository (currently the GitPython repository), etc., then this would get in the way of multiprocessing just as much as multithreading, thereby preventing test parallelism from being achieved through multiprocessing using `pytest-xdist` or other such tools. This is the sort of thing I'd be especially interested in becoming aware of and improving, though if there are reasons to specifically want multi*threaded* test runners to work, then I am of course interested.

(I've posted [in the review thread above](https://github.com/gitpython-developers/GitPython/pull/1729#discussion_r1382441631) about the narrower and separate issue of why I use paths like `tmp_path / "testdir"` instead of just `tmp_path`.)

test/test_util.py
Byron: Thanks for the explanation! Personally I think that it's OK to leave it as is but refactor the next time one stumbles over it. Please note that I didn't read the top-level comment yet, to which I will probably reply separately.

Byron:
Thanks for the analysis!

The reason I keep *thinking* about *multi-threading* is that it would force one to properly sandbox tests. Rust doest that by default as well.
This wasn't to say it's a goal for GitPython to be thread-safe, or to work in a *multiprocressing* powered environment nor that a test-runner which uses *multiprocessing* should be introduced - I'd leave everything as is as long as it works well enough.

I think what I try to say is that software that can deal with these situations typically correctly handles its state, which is always desirable. However, in Python I'd consider that out of scope (*whereas in Rust, it's naturally happening*). Overall, I think my relationship to Rust changes the way I think about everything else, even though achieving such a level of quality an diligence outside of Rust is usually close to impossible or at least incredibly frustrating.

PR #1730, opened 2023-11-03
Clarify license and make module top comments more consistent

EliahKagan:
While working on #1725, I noticed that comments at the top of modules might be made clearer and more consistent, but judged that such a change would be easier to review if proposed in a separate PR. So I've proposed those changes here instead, and included closely logically related changes as well.

This does not attempt to change the terms under which the project is licensed, nor does it change the license file. However, the changes here do include a change to some of the metadata used to specify the license, which will result in some license information being displayed differently on PyPI. I believe this change is for the better, of course, or I wouldn't be proposing it--it more specifically identifies the exact BSD license this project uses, the 3-Clause BSD License--but I recognize that this may merit special scrutiny in review.

Changes are presented in more detail in the three commit messages. In brief summary (the first commit covering the first *three* of these points):

- Filenames, which were given only in some modules, are omitted from all top comments, because GitPython contains many modules that have the same filename as others of its modules, so the filenames are in practice less clarifying while working on the code than they would be in some other projects.
- Top comments are now present in all modules one might expect or want them to be, indicating that the module is part of GitPython, and what the license is.
- Both existing and new license information in top comments is made more specific, specifying the 3-Clause BSD License rather than using the ambiguous phrase "BSD License". This change should not be confused with the change in [#1662](https://github.com/gitpython-developers/GitPython/pull/1662), which updated links that had come to point to the wrong license. This doesn't modify any URLs, instead modifying the text near them.
- The README is updated to give both names, emphasizing the more modern license name.
- `setup.py` is modified to pass `"BSD-3-Clause"` for the `license=` keyword argument to `setup`, instead of `"BSD"`. The license-related *classifier* is not changed, since no more specific classifier is available. (The commit message has significant further details about this.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,7 +289,7 @@ gpg --edit-key 4C08421980C9
 
 ### LICENSE
 
-[New BSD License](https://opensource.org/license/bsd-3-clause/). See the [LICENSE file][license].
+[3-Clause BSD License](https://opensource.org/license/bsd-3-clause/), also known as the New BSD License. See the [LICENSE file][license].
 
 [contributing]: https://github.com/gitpython-developers/GitPython/blob/main/CONTRIBUTING.md
 [license]: https://github.com/gitpython-developers/GitPython/blob/main/LICENSE
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -1,8 +1,7 @@
-# __init__.py
 # Copyright (C) 2008, 2009 Michael Trier (mtrier@gmail.com) and contributors
 #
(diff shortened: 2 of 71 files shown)


Byron (approved):
Thanks a lot for (what feels like) finalizing the improvement to the license which should now have arrived in its final state :). Great work!

> (The commit message has significant further details about this.)

Thanks for mentioning this specifically, that does help to make me take a look where otherwise I might not have done that.

PR #1735, opened 2023-11-06
Revise and restore some module docstrings

### LINKED ISSUE #1733: Unclear why fun.py (and others?) shouldn't have a module docstring
While working on #1725 I noticed the `git.index.fun` module has this comment at the top:

https://github.com/gitpython-developers/GitPython/blob/d5fc6e5337011d54fd2f3ff5aac71dbb38e43ee6/git/index/fun.py#L1-L2

Actually, that exact comment is due to my revision in that PR, but it's revised from:

https://github.com/gitpython-developers/GitPython/blob/add46d9daa7664628ba3512297a2401981694c98/git/index/fun.py#L1-L3

This was introduced in 9ccd777, at which time an existing module docstring was converted to a comment instead. Various other module docstrings were also converted to comments. But some modules today do have docstrings. Running `make -C doc html` seems to work fine; I ran it a number of times on Ubuntu and Windows while working on that PR and since.

I didn't want to assume it was okay to turn that back into a docstring. But either it (and any other top of module comments) should be turned into docstrings, *or* the reason they should not be should be made clearer (and any module docstrings causing problems fixed or converted back to comments).

EliahKagan:
Fixes #1733

In 9ccd777, several module docstrings were converted into comments or outright deleted, apparently due to at least one of them (in `git/index/fun.py`) and possibly others having broken Sphinx autodoc.

As discussed later in #1733, this no longer seems to be a problem, not even in `git/index/fun.py` where the comment included a note indicating it had.

So this converts them back to docstrings, re-adding those that were removed in 9ccd777. This also revises them, partly to avoid making outdated claims, but mostly for style and clarity. It also revises some already-present module docstrings. Other than modules 9ccd777 affected, this does not add new docstrings to modules without them.

One of the revisions made to some module docstrings here, to improve readability, is to remove language like "Module implementing" or "Module for", which was already used only in some places. However, this language is retained in the specific cases where it is clarifying, for example by avoiding documenting a module as if it the same thing as a single class implemented in it (which would be distracting because it would read as a mistake, and which could be confusing in cases where a developer uses an import with an `as` clause and needs to debug it because it was intended to be an `import` or `from` statement but was accidentally written as the other of those).

---

I have tested this on Ubuntu and Windows, with Python 3.12, with `make -C doc html` and inspecting some of the generated documentation (the documentation built from `git/index/fun.py`). Everything *seems* to work fine even after this change. The change causes text fo the module docstring to be added to the rendered documentation, as expected, and does not appear to break or interfere with generation of any of anything else, including documentation generated from other docstrings in the module. No errors or warnings are shown when generating the documentation.

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -3,8 +3,7 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
-"""Module containing module parser implementation able to properly read and write
-configuration files."""
+"""Parser for reading and writing configuration files."""
 
 import abc
 import configparser as cp
--- a/git/exc.py
+++ b/git/exc.py
@@ -3,7 +3,7 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
-"""Module containing all exceptions thrown throughout the git package."""
+"""Exceptions thrown throughout the git package."""
 
 __all__ = [
     # Defined in gitdb.exc:
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -3,6 +3,9 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
+"""Module containing IndexFile, an Index implementation facilitating all kinds of index
+manipulations such as querying and merging."""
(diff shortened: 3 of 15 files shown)


Byron (approved):
Thanks a lot for these continuous improvements, they are much appreciated.

Issue #1333, opened 2021-09-02
Library leaves dangling processes after use


zhiltsov-max:
Hi, I've faced a problem on Windows with file/directory removal with `shutil.rmtree()`. The problem, by itself, is not new and already has known solutions, including `git.util.rmtree()`. However, just using this function was not enough in my case and I started to dig deeper - there was an error about an open file handle. I discovered there were extra child `git` processes hanging in the process tree, but, unexpectedly, it was happening after all the library objects were removed:
![image](https://user-images.githubusercontent.com/13832349/131838938-b2c962ef-ca7c-4621-b4f5-71c9601f1fee.png)

Then I tried to catch and trace `Popen` calls from the library and found out this function:
https://github.com/gitpython-developers/GitPython/blob/5b3669e24a8ce7f3f482de86fcf95620db643467/git/cmd.py#L1190-L1199

From comments and names, it looks like such persistent behavior is intended. There is also the `AutoInterrupt` class here:
https://github.com/gitpython-developers/GitPython/blob/5b3669e24a8ce7f3f482de86fcf95620db643467/git/cmd.py#L367-L373

returned from:
https://github.com/gitpython-developers/GitPython/blob/5b3669e24a8ce7f3f482de86fcf95620db643467/git/cmd.py#L841

The class comment indicates that the process should be killed when the object goes out of scope. To me, it looks like an attempt to imitate [the RAII C++ idiom](https://en.cppreference.com/w/cpp/language/raii). Unfortunately, in Python it does not work this way because of garbage collector and the scoping rules. That is, an object that has no references can be removed in any time after it lost its last reference. The "pythonic" way to deal with this behavior is using a context manager (which this class is *not*). 

My suggestion is to collect these processes and manage them on a library/repo level, or wrap this in a context manager (which is better than just calling `wait()` manually in case of exceptions). BTW, probably, the process created is not reused after the first call.

``` python
import git
r = git.Repo.init('test_repo') 
r.index.commit('aa')   
del r
```
Upd: I see there is an undocumented `repo.close()`, which can help, but can have undesired side-effects because of a forced gc call.

Byron:
Thanks for all the research done to investigate, here are my thoughts.

> The class comment indicates that the process should be killed when the object goes out of scope. To me, it looks like an attempt to imitate the RAII C++ idiom. Unfortunately, in Python it does not work this way because of garbage collector and the scoping rules

[...]

PR #1745, opened 2023-11-28
Test native Windows on CI

### LINKED ISSUE #1630: Repo.index.reset fails on windows
## How to reproduce
1. clone any repo
2. edit a tracked file
3. in python, try to reset with
   ```python
   Repo().index.reset(working_tree=True)
   ```
   You will get a stack trace.

## Step by step recipe:
```
$ git clone https://github.com/gitpython-developers/GitPython git-python
$ cd git-python
$ python -m venv --prompt git venv
$ .\venv\Scirpts\Activate.ps1  # or whatever script is appropriate for you
$ rm pyproject.toml  # make any edit
$ git status -sb
## main...origin/main
 D pyproject.toml
$ python
>>> from git import Repo
>>> repo.index.reset(working_tree=True)
  File "C:\Users\<user>\code\git-python\git\index\util.py", line 103, in check_default_index
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<user>\code\git-python\git\index\base.py", line 1320, in reset
    new_inst = type(self).from_tree(self.repo, commit)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\<user>\code\git-python\git\index\base.py", line 367, in from_tree
    repo.git.read_tree(*arg_list, **kwargs)
  File "C:\Users\<user>\code\git-python\git\cmd.py", line 739, in <lambda>
  File "C:\Users\user>\code\git-python\git\cmd.py", line 1312, in _call_process
  File "C:\Users\<user>\code\git-python\git\cmd.py", line 1107, in execute
  cmdline: git read-tree --index-output=C:\Users\<user>\code\git-python\.git\tmpxnmxtmmu HEAD
  stderr: 'fatal: unable to write new index file'
```

## Expectation
I was expecting the result would be the equivalent of `git reset --hard HEAD`. Am I doing something incorrectly here? I'm unfamiliar with Windows, so it's quite possible.

EliahKagan:
This expands the CI test matrix in the main test workflow, `pythonpackage.yml`, so that it is parameterized not only by Python version but also by operating system. Now it tests on both Ubuntu and Windows. It also adds conditional `xfail` markings with precise conditions and brief descriptions of the expected failures.

Most of the detailed information about specific changes in this pull request is in the commit messages. I give general information here, as well as covering, in some detail, a few areas where I think it is not readily apparent (and maybe I am wrong) what the best approach is, or what changes should be included.

### Approach to adding `xfail` markings

A number of tests were known always to fail on Windows, with some others known to fail on Windows on particular versions or other conditions. As discussed in https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717475862, I have not attempted to actually fix the failures. Instead, I have marked each failing test case with a conditional `xfail` marking with a description of the failure and the exception with which the test fails. I have sought to make the conditions precise. Most of them are simply that a test fails on native Windows, so the condition is `os.name == "nt"`, but some of the conditions are more specific. In some cases there are really multiple independent expected causes of failure, so some have multiple `xfail` markings. (`pytest` supports stacking them in this way.)

I have tried to document what is currently known about the failures, including known or suspected information about their causes, in the descriptions, and sometimes in more detail in the commit messages that added them. Sometimes this takes the form of a reference to an existing issue, and for some others there may be an existing issue I did not manage to identify, but it is likely that a few bugs are currently only "reported" in commit messages in this PR. In some cases I anticipate I can fix a bug soon; others might benefit from having an issue opened for them.

(An exception is #1630, for which I added `xfail` marks in 6e477e3, which I did not document as thoroughly as I could have. I hope to open a PR to remedy #1630 as soon as Windows has CI test jobs--which if this PR is approved would be at that point--so I figured the commit message didn't need much information.)

### The `WinBashStatus` class in `test_index`

Some more detailed information about this appears in commit messages.

The original motivation for more precise detection of when `bash.exe` is the WSL wrapper was to fix CI. On the Windows CI runners, running `bash.exe` in a shell (any kind of shell) runs Git Bash, which is not related to WSL. But `Popen` finds the WSL-wrapper `bash.exe` in `System32` instead, because the way searching for executables works on Windows differs significantly between the shell and non-shell case. This is an intentional design choice in Windows, not GitPython nor even Python itself. But I think it's accurate to say GitPython must contend with it and, more importantly, some users of GitPython may be relying on it, perhaps even without knowing they are.

When a program is run without using a shell and a search must be performed because just the name of the executable is given, a few directories, including `System32`, are expected to be (and typically are) searched first, before the `PATH` environment variable is even used. This keeps a `shutil.which`-based technique from working. There are other important differences in search behavior of `shutil.which` and `Popen`, but that's the difference that's relevant here.

There are three related issues:

- One of the tests currently fails on WSL `bash`, while passing on native `bash` (like Git Bash).
- When `bash.exe` is found in `System32`, *and* it is the WSL wrapper, that does not imply the existence of any installed distributions (i.e., GNU/Linux systems to run in WSL). This is a common situation. Even when some Windows components needed for WSL to work are not yet installed, `bash.exe` can exist there, yet it does not always. (This is also the situation on CI when WSL is not set up.) Because the tests are `xfail` when `bash.exe` is altogether absent, I think it makes sense also to `xfail` them in the common situation that `bash.exe` can't run `bash` in a WSL-hosted distribution because no such distribution is present.
- Distinguishing expected and unexpected WSL failures, and failures not related to WSL, is complicated by how, while determining the status of WSL on a machine is usually fairly easy interactively, it seems there is no trivial way to do so programmatically in way that covers all cases, or even all common cases.

But the advantage of doing it seems significant to me: not only does this make it more efficient to interpret and document test failures, it also should be helpful in investigating and debugging possible future design changes that seek to build on [#1399](https://github.com/gitpython-developers/GitPython/pull/1399), in a sufficiently backward-compatible way, to retain its benefits (see [#703](https://github.com/gitpython-developers/GitPython/issues/703), [#971](https://github.com/gitpython-developers/GitPython/issues/971)), while avoiding unexpected behavior. (If it can be done in a way that is non-breaking for current users of hooks on Windows, perhaps in the future Git Bash can be used as the default way to run hooks, with other approaches, including WSL, available as backup strategies or if specified.)

I wrote `WinBashStatus` with the intention that, if necessary, it could be retained for an extended time. However, my hope for it is that it will actually disappear (perhaps with some insights from the process making their way into other code). If, in the future, hooks are run on Windows in a way that works on more systems, and has fewer strange edge cases, then there may be no more need for it or anything like it.

I have tried to make it work properly locally as well as on CI. To work locally, it needs to avoid language and locale related limitations other than those those already present. Because some aspects of the approach I used are not obvious, I've included comments about it. They were originally longer, with parenthesized examples of locale-related situations, but that seemed more detailed than necessary, so I removed them in [e00fffc](https://github.com/gitpython-developers/GitPython/pull/1745/commits/e00fffc918da5cd6c3c749d1d2e59d8ae6835189). (But I can bring those back if desired.)

### Which versions should we test?

This PR currently deviates from the preferences expressed in https://github.com/gitpython-developers/GitPython/pull/1654#issuecomment-1717475862 in one significant way, *which can be changed if you wish.* At the time of that discussion, we both held the view that it would be good to test only the lower and upper bounds of the supported version range for Windows. The process of adding these tests has led me to the view that it may be better, at least for now, to test all six versions on Windows (as is done on Ubuntu).

Initially I had them all enabled for the purpose of developing the changes here and planned to remove some at the end, but the reason I now advocate keeping them all for the time being, are, from most to least significant:

- The native Windows tests are much faster than I had feared, running *much* closer to the speed of the Ubuntu jobs than of the Cygwin job. (This is especially remarkable considering that it holds up even with them installing Debian via WSL in the Windows test runner to run the hook tests in `test_index.py`. I think there is value in running those tests, but their `xfail` marks do consider `bash.exe` being the WSL wrapper, yet no WSL systems being installed, as an expected failure condition. So if necessary, the `setup-wsl` step can be dropped to gain that minute. As detailed below, I don't know if caching works for the runs in an open PR, so you may observe further delay here.)
- The conditions under which tests fail on Windows are weirder than I had expected, in at least one way that specifically benefits from testing more versions: `TestSubmodule.test_rename` newly fails in 3.12 with a "being used by another process" `PermissionError`, possibly due to [GC changes](https://github.com/python/cpython/issues/97922). Because of the role of such errors in [#790](https://github.com/gitpython-developers/GitPython/issues/790) and [#1333](https://github.com/gitpython-developers/GitPython/issues/1333), it seems to me that easily distinguishing what versions have a failure, before and after code changes that potentially affect the test, might be helpful in the near future. To do this would at least require a 3.11 job.
- The lower bound GitPython currently supports is 3.7. There are a number of important changes in 3.8, which so far seem not to have been an issue; most special-casing of 3.7 seems to have been in test suite only. To ensure that changes (including security fixes) are effective on 3.7, it should be tested on Windows. To decide what to do about it when that breaks, 3.8 should be tested as well.
- The latest Python build the Cygwin project currently packages is 3.9; for this reason, it is probably the most used on Cygwin, and it makes sense that it is the version tested in the Cygwin job. I had not until recently considered the benefits of being able to compare 3.9 results between native Windows and Cygwin. I think this would avoid situations where one might worry the problem is with 3.9, though that may only be a small boon. However, the stark difference in performance between native Windows and Cygwin makes me think the Cygwin job could be sped up. (For example, it may be possible to use the GitHub Actions caching feature to do I/O in a faster way for installing Cygwin and/or GitPython's dependencies in the Cygwin system.) This is another area where the ability to compare could be helpful.

Most of the benefit could be gotten by testing 3.7, 3.8, 3.11, and 3.12 (the last point above is less compelling than the others, I think). However, this is still much more than the original idea of testing just a couple versions. It seems to me that the additional resource usage of testing two more versions is worthwhile for the benefit of comprehensively testing all the versions GitPython officially supports. I also think other approaches for decreasing undesired load might be preferable, such as removing `fail-fast: false` (the Ubuntu jobs, if desired, could still be set to continue even if another job has failed), skipping the documentation-building step on some or all Windows jobs, and/or either not installing WSL for the hook tests or switching the distribution to Alpine Linux.

With many pushes over a short time--as for example in this pull request, where I pushed the commits separately with ten second delays in between, to make it easy to inspect intermediate CI results--there can be more jobs queued than immediately available runners. (I believe the number of runners is limited per organization.) It is in that case that the number of tests can be most of an issue. Even *if* you decide to accept this PR without the number of Windows jobs being decreased, it might turn out later that this is annoying. I accept that you may know now that you want fewer Windows test jobs, but also, I understand that if you accept this as it is now, that doesn't mean all the jobs will be kept forever.

### `setup-wsl` and caching

On my fork, the `setup-wsl` step seems usually to take between 30 and 90 seconds, and most often near the low end.

I did not disable GitHub Actions caching on `setup-wsl`. Based on how it works in my fork, I expect it to take up 82 MB. I think this is probably fine, and it can be turned off if desired, but I wanted to mention it just in case. I believe GitHub Actions caching quotas are per-organization. I expect that turning off caching (while still installing WSL) would make things somewhat slower.

I am unsure if caching will actually work on it in the CI jobs run on this repository due to the `pull-request` trigger. My vague recollection is that caching does not happen for open PRs that introduce it. If that is the case, then the amount by which installing WSL slows down each Windows job, *as observed here*, may be more than I have observed.

### What is *not* done here

It seemed to me that the `bash.exe` status classification logic in `test_index` (described above) was easier to do in this PR than separately, because it facilitates precise `xfail` markings. However, other substantial changes that could in principle be part of this PR but can be omitted are omitted. Specifically:

- I have not merged `cygwin-test.yml` into `pythonpackage.yml`. I had hoped to do that while adding native Windows tests, but I now think it is best deferred to a later PR. One reason is for more effective reviewing: if merging them produces conditional logic whose complexity is undesirable, it will be easier to notice the problem after the native Windows tests already exist for comparison. Another reason, though, is that I don't know what it will look like after changes have been made to it to speed it up (see above). If this involves adding caching, including for installing `GitPython`'s dependencies, then that might cause it to diverge sufficiently from `pythonpackage.yml` that they should no longer be merged.
- I have not changed the distribution used for WSL in the tests from Debian to Alpine Linux, even though I suspect this would speed things up (perhaps especially if caching is going to be turned off in the `setup-wsl` step). This requires either a fix for [setup-wsl#50](https://github.com/Vampire/setup-wsl/issues/50) or a workaround.
- I have not added macOS jobs. I think this may be worth doing. I don't currently have the ability to test on Apple hardware. I can virtualize systems I don't regularly use in development, such as OpenIndiana or the Simplified Chinese localized build of Windows Server, on Ubuntu or Windows hosts. But virtualizing macOS is not similarly straightforward. Furthermore, some subtleties of Unix can be revealed by testing regularly on systems that are not GNU/Linux, of which macOS has GitHub Actions runners. (For [#1738](https://github.com/gitpython-developers/GitPython/issues/1738), the way I tested on macOS was to run a GitHub Actions job on a macOS runner with a [tmate debugging](https://github.com/marketplace/actions/debugging-with-tmate) step and SSH into it.) However, omitting that here limits scope and allows it to be evaluated separately. Limited testing indicates that, as on Windows, some different expected timings may have to be given in at least one of the performance tests in order to make macOS jobs pass.
- I have not reduced the complexity of `rmtree` and `HIDE_WINDOWS_KNOWN_ERRORS` tests in `test_util`. With native Windows CI jobs, *some* of the behavior that is specific to Windows (especially since [#1739](https://github.com/gitpython-developers/GitPython/pull/1739)) can probably be tested only on Windows, and skipped otherwise. I think there should be a way to make this change in a way that simplifies the tests. (As one example, every `os.name == "nt"` in a `@pytest.mark.parametrize` parameter set would just become `True`.) But it's not at all necessary to do that to achieve the aims here, and it would delay this PR.
- I have not converted tests that skip due to `unittest.SkipTest` being raised in `git.util.rmtree` into `xfail` tests, made `HIDE_WINDOWS_KNOWN_ERRORS` default to `False`, or decreased its use (though I have made sure not to increase its use). It seems to me that such steps would be good progress toward a solution for [#790](https://github.com/gitpython-developers/GitPython/issues/790), but I think they will be easier to do, as well as to review, if done after, and separately from, the addition of native Windows CI jobs.

### DIFF
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -73,7 +73,6 @@ jobs:
         python -c 'import sys; print(sys.platform)'
         python -c 'import os; print(os.name)'
         python -c 'import git; print(git.compat.is_win)'  # NOTE: Deprecated. Use os.name directly.
-        printenv PATH | tr ':' '\n'
 
     - name: Test with pytest
       run: |
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -73,6 +73,7 @@ jobs:
         python -c 'import sys; print(sys.platform)'
         python -c 'import os; print(os.name)'
         python -c 'import git; print(git.compat.is_win)'  # NOTE: Deprecated. Use os.name directly.
+        printenv PATH | tr ':' '\n'
 
     - name: Test with pytest
       run: |
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -70,9 +70,7 @@ jobs:
         command -v git python
         git version
         python --version
-        python -c 'import sys; print(sys.platform)'
-        python -c 'import os; print(os.name)'
-        python -c 'import git; print(git.compat.is_win)'  # NOTE: Deprecated. Use os.name directly.
+        python -c 'import os, sys; print(f"sys.platform={sys.platform!r}, os.name={os.name!r}")'
 
     - name: Test with pytest
       run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,6 +35,12 @@ jobs:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: ${{ matrix.experimental }}
 
+    - name: Set up WSL (Windows)
+      if: startsWith(matrix.os, 'windows')
+      uses: Vampire/setup-wsl@v2.0.2
+      with:
+        distribution: Debian
+
     - name: Prepare this repo for tests
       run: |
         ./init-tests-after-clone.sh
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,6 +35,12 @@ jobs:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: ${{ matrix.experimental }}
 
+    - name: Set up WSL
+      if: startsWith(matrix.os, 'windows')
+      uses: Vampire/setup-wsl@v2.0.2
+      with:
+        distribution: Debian
+
     - name: Prepare this repo for tests
       run: |
         ./init-tests-after-clone.sh
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -65,7 +65,27 @@ jobs:
         python -c 'import sys; print(sys.platform)'
         python -c 'import os; print(os.name)'
         python -c 'import git; print(git.compat.is_win)'  # NOTE: Deprecated. Use os.name directly.
-        printenv PATH | tr ':' '\n'
+
+    # For debugging hook tests on native Windows systems that may have WSL.
+    - name: Show where bash.exe may be found
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        set +e
+        type -a bash.exe
+        python -c 'import shutil; print(shutil.which("bash.exe"))'
+        bash.exe --version
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "--version"]); print(f"result: {p!r}")'
+        bash.exe -c 'echo "$BASH"'
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "-c", """echo "$BASH" """]); print(f"result: {p!r}")'
+        bash.exe -c 'echo "$BASH_VERSION"'
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "-c", """echo "$BASH_VERSION" """]); print(f"result: {p!r}")'
+        bash.exe -c 'printenv WSL_DISTRO_NAME'
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "-c", "printenv WSL_DISTRO_NAME"]); print(f"result: {p!r}")'
+        bash.exe -c 'ls -l /proc/sys/fs/binfmt_misc/WSLInterop'
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "-c", "ls -l /proc/sys/fs/binfmt_misc/WSLInterop"]); print(f"result: {p!r}")'
+        bash.exe -c 'uname -a'
+        python -c 'import subprocess; p = subprocess.run(["bash.exe", "-c", "uname -a"]); print(f"result: {p!r}")'
+      continue-on-error: true
 
     - name: Check types with mypy
       run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,12 +35,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: ${{ matrix.experimental }}
 
-    - name: Set up WSL
-      if: startsWith(matrix.os, 'windows')
-      uses: Vampire/setup-wsl@v2.0.2
-      with:
-        distribution: Debian
-
     - name: Prepare this repo for tests
       run: |
         ./init-tests-after-clone.sh
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -65,6 +65,7 @@ jobs:
         python -c 'import sys; print(sys.platform)'
         python -c 'import os; print(os.name)'
         python -c 'import git; print(git.compat.is_win)'  # NOTE: Deprecated. Use os.name directly.
+        printenv PATH | tr ':' '\n'
 
     - name: Check types with mypy
       run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -10,18 +10,19 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
     strategy:
       fail-fast: false
       matrix:
+        os: ["ubuntu-latest", "windows-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         include:
         - experimental: false
 
+    runs-on: ${{ matrix.os }}
+
     defaults:
       run:
-        shell: /bin/bash --noprofile --norc -exo pipefail {0}
+        shell: bash --noprofile --norc -exo pipefail {0}
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,6 +35,12 @@ jobs:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: ${{ matrix.experimental }}
 
+    - name: Set up WSL (Windows)
+      if: startsWith(matrix.os, 'windows')
+      uses: Vampire/setup-wsl@v2.0.2
+      with:
+        distribution: Debian
+
     - name: Prepare this repo for tests
       run: |
         ./init-tests-after-clone.sh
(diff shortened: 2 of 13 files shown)


EliahKagan:
> Just as a note in case it's ever relevant: I have been mimicking gits way of launching programs at least to the extend that I understand it with the [gix-command](https://github.com/Byron/gitoxide/blob/115c993f7919b4f7edcf04e92bb8293957c028a5/gix-command/src/lib.rs) crate, and it has some special handling for sh actually for Windows as well.

Cool! I'll take a look at this soon--and reply further.

> From what I can tell, `git` doesn't even try to run a shell on Windows, or it's so well hidden that I didn't see it.

I haven't looked into this for hooks. But custom `git` commands implemented as shell scripts work on Windows, and the `PATH` search is as if it were done in the Git Bash MSYS-like environment, even when the `git` command is not run from that environment. I'm not clear on how it achieves that.

At the time the `SHELL_PATH` constant, which I believe varies by platform but is baked into the build, was introduced, it was said [not to apply to Windows](https://github.com/git/git/blob/v1.7.11/Documentation/RelNotes/1.7.11.txt#L116-L118). But I believe that is not currently the case; code in [git-for-windows/git](https://github.com/git-for-windows/git) seems to make significant use of it, including in code that is enabled on Windows.

There is also [a current proposal](https://github.com/git-for-windows/git/pull/3301) in Git for Windows to make what shell it uses configurable at runtime.

> test caching and the effect of using Alpine as DSL instead of Debian. This would also be a first for using a MUSL linux system.

I am also looking forward to this. Of course, this is just for the WSL system used to run shell scripts on Windows (in the cases that WSL is what's used to do that).

Although this does not diminish the likely value in replacing Debian with Alpine for WSL, I think the even more interesting and potentially illuminating thing to do would be to run GitPython itself, and its whole test suite, on Alpine Linux. I don't think GitHub Actions has Alpine Linux runners, but this could be done in a Docker container on an Ubuntu runner.

Issue #1616, opened 2023-07-17
Needs better error message when repository is rejected by `git` (related to `safe.directory`)


marioaag:
**Initial state:**
Executions using the latest version of Git (version 2.39.3), in a repository owned by other user.

**Current behavior:**
Executing: 
```
repo = git.Repo(dir_path)
last_commit = repo.head.commit
```
Ambiguos error is reported:
```
  File "/usr/local/lib/python3.9/site-packages/git/refs/symbolic.py", line 226, in _get_commit
    obj = self._get_object()
  File "/usr/local/lib/python3.9/site-packages/git/refs/symbolic.py", line 219, in _get_object
    return Object.new_from_sha(self.repo, hex_to_bin(self.dereference_recursive(self.repo, self.path)))
  File "/usr/local/lib/python3.9/site-packages/git/objects/base.py", line 94, in new_from_sha
    oinfo = repo.odb.info(sha1)
  File "/usr/local/lib/python3.9/site-packages/git/db.py", line 40, in info
    hexsha, typename, size = self._git.get_object_header(bin_to_hex(binsha))
  File "/usr/local/lib/python3.9/site-packages/git/cmd.py", line 1383, in get_object_header
    return self.__get_object_header(cmd, ref)
  File "/usr/local/lib/python3.9/site-packages/git/cmd.py", line 1370, in __get_object_header
    return self._parse_object_header(cmd.stdout.readline())
  File "/usr/local/lib/python3.9/site-packages/git/cmd.py", line 1329, in _parse_object_header
    raise ValueError("SHA could not be resolved, git returned: %r" % (header_line.strip()))
ValueError: SHA could not be resolved, git returned: b''
```

**Expected behavior:**
More clear an specific error message or even better a parameter in git.repo() to handle the dubious ownership

**Notes:**
Doing some debug found that the new version of Git has enable by default a check for dubious ownership that is not handle gracefully by GitPython.

A work around for this is add an exception in Git to consider the file safe `git config --global --add safe.directory <dir path>`

Byron:
Do I understand correctly that the repository isn't trusted by `git` and `GitPython` fails with a dubious, seemingly unrelated error?

marioaag:
Correct!

Byron:
[...]

PR #1746, opened 2023-11-28
Adding dubious ownership handling

marioaag:
This is a fix for the ticket -> https://github.com/gitpython-developers/GitPython/issues/1616

This was already validated and catch the dubious ownership from the repo object generation to avoid confusing errors forward.

Validations results:

Catching dubious ownership
```
[root@ilabinfra-dev BlueGen]# python3
Python 3.9.16 (main, Sep 12 2023, 00:00:00) 
[GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] on linux
>>> import git
>>> repo = git .Repo('/tmp/BlueGen')
  File "/usr/local/lib/python3.9/site-packages/git/repo/base.py", line 303, in __init__
    raise InvalidGitRepositoryError(err_msg) from None
git.exc.InvalidGitRepositoryError: Detected dubious ownership in repository at: /tmp/BlueGen
```

No dubious ownership
```
[mario@ilabinfra-dev BlueGen]$ python3
Python 3.9.16 (main, Sep 12 2023, 00:00:00) 
[GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] on linux
>>> import git
>>> repo = git .Repo('/tmp/BlueGen')
```

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1322,7 +1322,12 @@ class Git(LazyMixin):
         tokens = header_line.split()
         if len(tokens) != 3:
             if not tokens:
-                raise ValueError("SHA could not be resolved, git returned: %r" % (header_line.strip()))
+                err_msg = (
+                    f"SHA is empty, possible dubious ownership in the repository "
+                    f"""at {self._working_dir}.\n            If this is unintended run:\n\n         """
+                    f"""             "git config --global --add safe.directory {self._working_dir}" """
+                )
+                raise ValueError(err_msg)
             else:
                 raise ValueError("SHA %s could not be resolved, git returned: %r" % (tokens[0], header_line.strip()))
             # END handle actual return value


EliahKagan:
I've noticed this PR contains a rebased version of 359116b as 07f3641 with the same change and message, which seems like it may be unintentional. Depending how this PR is merged, I would expect the duplicate commit to make it into the upstream GitPython history. I don't know that this is really a problem. But if you want to avoid it, running `git rebase main` on the `GitPython-1616` branch should remove it, showing a message like this (after which the rebased `GitPython-1616` branch could be force-pushed):

```text
➜ /workspaces/GitPython (GitPython-1616) $ git rebase main
warning: skipped previously applied commit 07f36411
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
Successfully rebased and updated refs/heads/GitPython-1616.
```

Sorry if there's a reason for this that I missed. I just figured I'd mention it in case it is not intended.

git/repo/base.py
```diff
@@ -296,6 +296,16 @@ def __init__(
         else:
             self.odb = odbt(rootpath)
 
+        # Check dubious ownership
+        try:
+            if not (self.bare or Git.is_cygwin()):
+                _ = self.git.status()
```
Byron, line 302: `git status` can be a very expensive operation, and even when it's not it's still expensive and can't be happening when creating a new Repository instance. Interestingly, it will only be cheap if it fails due to dubious ownership.

Can you find another invocation that ideally just checks for permissions and nothing else?

git/repo/base.py
```diff
@@ -296,6 +296,16 @@ def __init__(
...
+        except GitCommandError as err:
+            if "detected dubious ownership" in err.stderr:
+                err_msg = f"Detected dubious ownership in repository at: {epath}"
```
Byron, line 305: I feel that for many, this message will still not help them understand what's going on, despite it being official `git` lingo.

It think what would help is if it would mention a remedy, like `If this is unintended, run "git config --global --add safe.directory {epath}"`.

Byron (changes requested):
Thanks a lot for tackling this.

A huge problem I see with the approach, is that it imposes the latency of one additional git invocation directly onto the `Repo` instantiation code-path, which runs a lot. The slowdown will be substantial, and a high price to pay for a better error message.

I believe the price is too high, but there might be remedies.

This is because `GitPython` already calls `git` a couple of times, here is what I recall.

* In `cmd.py` it obtains `_version_info` by calling `git` - this happens lazily, and I wonder if it could fail.
* When accessing objects, it spawns two `git` commands to stream objects.

Can a nice error message be generated there instead, if these commands are failing? Maybe more generally, is it possible to 'supervise' the stderr of any invoked command to see if the execution failed because of it? In terms of compute power, I presume this is a better deal as it only costs once commands fail, which usually is an error condition anyway.

What do you think?

Byron:
> I've noticed this PR contains a rebased version of [359116b](https://github.com/gitpython-developers/GitPython/commit/359116b3b9b7011eb7893fa584a43522f1b9bb57) as [07f3641](https://github.com/gitpython-developers/GitPython/commit/07f364111aa2f51e3931aab9086b9790a03bb030) with the same change and message, which seems like it may be unintentional. Depending how this PR is merged, I would expect the duplicate commit to make it into the upstream GitPython history. I don't know that this is really a problem. But if you want to avoid it, running `git rebase main` on the `GitPython-1616` branch should remove it, showing a message like this (after which the rebased `GitPython-1616` branch could be force-pushed):
> 
> ```
> ➜ /workspaces/GitPython (GitPython-1616) $ git rebase main
> warning: skipped previously applied commit 07f36411
> hint: use --reapply-cherry-picks to include skipped commits
> hint: Disable this message with "git config advice.skippedCherryPicks false"
> Successfully rebased and updated refs/heads/GitPython-1616.
> ```
> 
> Sorry if there's a reason for this that I missed. I just figured I'd mention it in case it is not intended.

Thanks, @EliahKagan, for the suggestion. That's exactly what I did and it resolved the commit-duplication like anticipated.

git/repo/base.py
marioaag: ok, I tried to be minimalist but agree with you the remedy instruction could be helpful

git/repo/base.py
marioaag: You came to the same dilemma that I get when I was implementing this...
I thought to just keep it as message enhancement when the git command gets executed but then I thought that it could be little late for the user that already generated the repo instance and maybe did some other stuff but I didn't consider that it could be very expensive operation, not sure if do a manual check could be less expensive just like git is doing it, check if the ownership is different to the actual user but then a lot of extra logic will need to be implemented to consider the git config file exceptions, etc. Kind of duplicated logic and more conde to maintain.

Is it an easy way to check how expensive it is to execute this command? it should be just one time at the repo instantiation, maybe change it for another operation less expensive like check the latest commit?

Byron: I thought about this and realized that adding something into the instantiation-path would still be unacceptable even if it could happen at no cost. The reason for this is that right now, some operations can be performed without `git` and safely so, so users shouldn't be held back by something that `git` *would* say. Further, doing so would be a breaking change.

Personally I think `git` was 'adamant' in their implementation as they put defence of a particular issue before usability. It's understandable, but also far from the best possible implementation of such mitigation.

`gitoxide` was lucky enough as second-mover to have a fresh take on this, as it will simply ignore configuration data from untrusted sources. GitPython should do the same, as much as possible, and prefer to provide better error messages if `git` fails.

marioaag: get it! now I see what is the philosophy of GitPython, let me reformulate this implementation and go with an enhancement of the message in the cases where it apply.

marioaag:
Hi @Byron, the new approach is implemented! I found that all the errors related with dubious ownership comes to the missing SHA so the steep where it is used is the best place to report the issue and all the rest of the actions that can run without this info will be available.

Byron (approved):
Great, that's it!

Thanks so much for going the last mile with this :)!

PR #1748, opened 2023-11-28
Always read metadata files as UTF-8 in setup.py

### LINKED ISSUE #1747: Source and editable installs fail in some locales
Methods of installing GitPython that run code from `setup.py` fail in some locales. This does not affect installing from a wheel, but it does affect installing from an sdist, or installing from a local directory, including the editable install procedure recommended for development in the readme. The same problem happens when building GitPython. Building or installing using the old method of running `setup.py` directly is also affected. The error is of this form, though the codec will not always be `gbk`:

```text
UnicodeDecodeError: 'gbk' codec can't decode byte 0xa6 in position 473: illegal multibyte sequence
```

I am unsure if this ever happens in practice on Unix-like systems, whose locales are usually UTF-8. However, it happens on Windows systems in which `README.md` cannot be decoded using the system's active ANSI code page. This is a rarely-changed systemwide setting, so changing the user preferred languages, display language, or input method are not workarounds. I discovered this on a Simplified Chinese (zh-CN) build of Windows Server 2022 while using it to test some WSL-related test helper logic in #1745. Such a system uses ANSI code page 936. `README.md` is UTF-8, but it currently happens that it can be decoded with code page 1252, which Windows builds for Western European languages use as their ANSI code page. I expect encodings other than cp936 to fail as well.

With the PyPI sdist for GitPython 3.1.40 (on Python 3.12 x86-64, though I expect all supported versions to be affected):

```text
(.venv) C:\Users\Administrator\gptest> pip install --no-binary GitPython GitPython
Collecting GitPython
  Using cached GitPython-3.1.40.tar.gz (200 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [18 lines of output]
        File "C:\Users\Administrator\gptest\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
          main()
        File "C:\Users\Administrator\gptest\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Administrator\gptest\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Administrator\AppData\Local\Temp\pip-build-env-8j1uxmuv\overlay\Lib\site-packages\setuptools\build_meta.py", line 325, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Administrator\AppData\Local\Temp\pip-build-env-8j1uxmuv\overlay\Lib\site-packages\setuptools\build_meta.py", line 295, in _get_build_requires
          self.run_setup()
        File "C:\Users\Administrator\AppData\Local\Temp\pip-build-env-8j1uxmuv\overlay\Lib\site-packages\setuptools\build_meta.py", line 311, in run_setup
          exec(code, locals())
        File "<string>", line 20, in <module>
      UnicodeDecodeError: 'gbk' codec can't decode byte 0xa6 in position 473: illegal multibyte sequence
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

A workaround is to pass `-X utf8` to `python`. For example, `python -X utf8 -m pip install ...` can be used for installation.

The fix should be straightforward. Importing `setup.py` confirms that the specific cause is reading `README.md`:

```text
(.venv) C:\Users\Administrator\repos\GitPython [main ≡]> python -c 'import setup'
  File "C:\Users\Administrator\repos\GitPython\setup.py", line 20, in <module>
    long_description = rm_file.read()
                       ^^^^^^^^^^^^^^
UnicodeDecodeError: 'gbk' codec can't decode byte 0xa6 in position 477: illegal multibyte sequence
```

It can be fixed by passing `encoding="utf-8"`. **I have proposed this change in #1748.**

gitdb and smmap are unaffected, because [gitdb does not open files to read those data in `setup.py`](https://github.com/gitpython-developers/gitdb/blob/ec58b7e55c8fec2479290a22faa293799f6805fc/setup.py#L24), while [smmap does open `README.md` but passes `encoding="utf-8"`](https://github.com/gitpython-developers/smmap/blob/f1ace75be355fdec927793e462b9b12bf6ec9520/setup.py#L14).

EliahKagan:
Fixes #1747

This passes `encoding="utf-8"` to the `open` calls in `setup.py`, so the readme, version, and requirements files are always read as UTF-8, even on systems whose locale is not UTF-8. This fixes the bug described in #1747 where installation other than from a pre-built wheel would fail on many Windows systems using non-European languages.

The specific problem occurred with the `README.md` file. The requirements files are less likely to contain characters not in the ASCII subset, though maybe they could come to contain them, perhaps in comments. The `VERSION` file is even less likely to ever contain such characters. Nonetheless, for consistency, because it is a best practice, and because it appears to be the intent of the existing code, `encoding="utf=8"` is added for opening all of them.

This change is tested on a system whose locale uses Windows code page 936. Editable installation, as well as the other affected ways of installing (and building) described in #1747, are now working. (Installing from a pre-built wheel was never affected.)

I lean toward the position that it's not worth attempting a regression test for this, because although we do have `test_installation`, a truly reliable regression test for #1747 would require the presence of a locale in which the existing logic breaks, but in practice the systems where this happens are Windows, where a single locale (in the sense of "locale" relevant here) is systemwide. Furthermore, the fix is straightforward, following the current best practice of passing `encoding="utf-8"` explicitly where there is not reason to do otherwise. However, there may be a way to add a reasonably reliable test using mocking or other techniques, and I'd be pleased to pursue that if you consider it warranted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,16 +7,16 @@ from setuptools.command.sdist import sdist as _sdist
 import os
 import sys
 
-with open(os.path.join(os.path.dirname(__file__), "VERSION")) as ver_file:
+with open(os.path.join(os.path.dirname(__file__), "VERSION"), encoding="utf-8") as ver_file:
     VERSION = ver_file.readline().strip()
 
-with open("requirements.txt") as reqs_file:
+with open("requirements.txt", encoding="utf-8") as reqs_file:
     requirements = reqs_file.read().splitlines()
 
-with open("test-requirements.txt") as reqs_file:
+with open("test-requirements.txt", encoding="utf-8") as reqs_file:
     test_requirements = reqs_file.read().splitlines()
 
-with open("README.md") as rm_file:
+with open("README.md", encoding="utf-8") as rm_file:
     long_description = rm_file.read()
 
 


Byron:
Thanks so much for tackling this!

The handling of encodings always have been a weak spot for the project (and I think, are weak spots for python in general), but this small change is an important win for sure.

I wouldn't want to see your time ~~wasted~~ spent on trying to test this though, as I think it's reasonable to assume encoding related issues won't occur with that as the encoding of these files is under control of the repository.

Something I find interesting though if it would present an attack-vector. Could some installation-paths be made to fail by silently changing the encoding of `README.md`? I'd assume such an act shows up clearly in a diff, so I'd think the answer is "no" here. And with that, it should be fine to keep it as is without further regression-protection.

PR #1760, opened 2023-12-09
Overhaul noqa directives

EliahKagan:
This includes a few small style improvements I had missed in previous PRs, but consist mostly of improvements to how and where `noqa` directives are used, and code changes (intended to be non-breaking and uncontroversial) to go along with that.

The directives are made more precise. They always give specific errors/warnings to suppress, both where that had been intended but idiosyncrasies of the syntax caused them to be broader, and where nothing had been given before. All `noqa` directives are now statement-level rather than file-level. Some of those that were no longer needed are removed with no other associated change. Some others are removed together with minimal changes to the code they complained about or to other `noqa` directives that had been ineffective (such as due to their placement).

The commit messages have some more information about the details of the changes.

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -3,28 +3,27 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
-# flake8: noqa
 # @PydevCodeAnalysisIgnore
-
-from git.exc import *  # @NoMove @IgnorePep8
-from typing import List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
-from git.types import PathLike
 
 __version__ = "git"
 
+from typing import List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
+
 from gitdb.util import to_hex_sha
+from git.exc import *  # noqa: F403  # @NoMove @IgnorePep8
+from git.types import PathLike
 
 try:
     from git.compat import safe_decode  # @NoMove @IgnorePep8
     from git.config import GitConfigParser  # @NoMove @IgnorePep8
-    from git.objects import *  # @NoMove @IgnorePep8
-    from git.refs import *  # @NoMove @IgnorePep8
-    from git.diff import *  # @NoMove @IgnorePep8
-    from git.db import *  # @NoMove @IgnorePep8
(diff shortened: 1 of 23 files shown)


Byron (approved):
Thanks so much for your continued contributions, and for (nearly) single-handedly making GitPython more maintainable!

EliahKagan:
> Thanks so much for your continued contributions, and for (nearly) single-handedly making GitPython more maintainable!

Well, a couple of the broken line-level suppressions had also been *put in code* by me. But thanks!

More broadly, this is actually a case where `ruff` has already helped, even though I don't really have it working for this project and I'm not sure when I will.

More challenges than I expected have arisen in trying to migrate `flake8` to `ruff` for GitPython that I think have to be solved to preserve some of the checks GitPython has been benefiting from, including that some of those checks are still considered to be in preview for Ruff and I want to make I understand the implications of `preview = true`, that `ruff` has to be compiled if installed on Cygwin but this always fails with the `test-cygwin.yml` runner is set up (and also it would be bad to make that workflow take even longer to accommodate a compilation), and `Literal["foo"]` is only special-cased to recognize `"foo"` as a literal string rather than a type name when `Literal` is imported from `typing` or `typing_extensions`, whereas GitPython imports it from one of its own modules, which conditionally imports it from one or the other of those modules based on the Python version (I'd rather not suppress the warning about accessing a name that was never defined).

But when experimenting with it, I did find that it warned about not recognizing some of the file-level suppressions that `flake8` recognized but as something broader and quite different from what they intended. (I consider it a good thing about `ruff` that it did not recognize them!) I considered fixing them, but I decided that instead they could be removed, leading to this PR.

Byron:
Thanks for sharing the details on how this PR came to be, it's exciting to hear some R~~ust~~uff will be happening.

I wonder if `cygwin` has to use `ruff` (or `flake` for that matter) at all. Will it make a difference given that the linter will probably work the same on any platform? But probably there is a misconception somewhere on my side.

EliahKagan:
Cygwin doesn't have to use `ruff`. This is a minor issue, just one that has to be figured out. `ruff` can be omitted as a test dependency altogether and just used via `pre-commit`, which is the situation with `flake8`. Or maybe it could be listed but omitted for `platform == "cygwin"`. Or a `test-minimal` group could be created. There are probably other options.

Issue #1756, opened 2023-12-03
Git.execute's kill_after_timeout callback assumes procps


EliahKagan:
### Background

Calling `Git.execute`—whether directly, or indirectly by calling the dynamic attributes of a `Git` instance—and passing `kill_after_timeout` with a non-`None` value, creates a timer on a separate thread that calls the local [`kill_process`](https://github.com/gitpython-developers/GitPython/blob/fe082ad5e297119a3073dab74fa604eaf547ecc7/git/cmd.py#L1008-L1036) function. This callback function uses `os.kill` to kill the process. Before killing the process, it enumerates the process's *direct children*. If sending the first signal succeeds (basically, if the parent process still existed), it also attempts to kill the child processes.

The children are enumerated with `ps --ppid`:

https://github.com/gitpython-developers/GitPython/blob/fe082ad5e297119a3073dab74fa604eaf547ecc7/git/cmd.py#L1010-L1014

### The problem

The `--ppid` option is [not POSIX](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/ps.html). Most GNU/Linux systems have [procps](https://gitlab.com/procps-ng/procps), whose `ps` implementation supports `--ppid`. I am unsure if *any* other implementations of `ps` support it. The procps tools generally run only on Linux-based systems, because they use the `/proc` filesystem (and assume it is laid out as in Linux). Although they can run on any such system, Alpine Linux and some minimal GNU/Linux environments do not ship them, defaulting to `ps` from [busybox](https://busybox.net/) instead.

[...]

PR #1761, opened 2023-12-10
Clarify some Git.execute kill_after_timeout limitations

EliahKagan:
This makes the two preliminary changes discussed as possibly useful in https://github.com/gitpython-developers/GitPython/issues/1756#issuecomment-1848350451:

- **The `Git.execute` docstring is modified to include additional portability information for `kill_after_timeout`.** I've tried to do this in a way that avoids going into too much detail, but still gives enough information to allow developers of software using GitPython to infer what systems may be reasonable to use it on, or what the effects of using it on systems where it is less effective might be. I considered linking in the docstring to #1756 or discussing the race condition issue, but omitted both at this time. I'd be pleased to add such information if it is wanted.

- **The implementation of the local `kill_process` function, which is used as the callback for `kill_after_timeout`, is modified to avoid giving the impression that it could be safely or effectively used in similar form on Windows.** This mostly consists of removing code that was present with the intention of supporting Windows. But one of the changes that is part of this is raising `AssertionError` in the event that it somehow were called on Windows. This should not currently be able to happen because Windows is checked for, and an exception raised, [prior to registering](https://github.com/gitpython-developers/GitPython/blob/a58a6be043f62e6552ef1638ec497f3b805c5480/git/cmd.py#L960-L964) the callback in the first place. But this is an area of the code that is somewhat complicated, and I think this may be justified to express how that code is not portable or safe for Windows as written. The changes to `kill_process` are done in their own commit (and just one), so that they are easier to identify if in the future some of the attempts for it to support Windows are revisited (though this could not be done by calling the `ps` command).

See discussion in #1756. However, I do not assum based on what has been discussed there that these specific changes are necessarily ideal, and I'd be pleased to make changes on request.

I considered adding a test but decided not to, based in part on the point about tests in https://github.com/gitpython-developers/GitPython/issues/1756#issuecomment-1838101270, and in part on the limited nature of the changes. However, I did do the following to check that it still seems to work:

```text
>>> git.Git().execute(['sleep', '10'], kill_after_timeout=5)
  File "/workspaces/GitPython/git/cmd.py", line 1111, in execute
git.exc.GitCommandError: Cmd('sleep') failed due to: exit code(-9)
  cmdline: sleep 10
  stderr: 'Timeout: the command "sleep 10" did not complete in 5 secs.'
>>> git.Git().execute(['sleep', '4'], kill_after_timeout=5)
''
```

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1015,11 +1015,9 @@ class Git(LazyMixin):
 
         def kill_process(pid: int) -> None:
             """Callback to kill a process."""
-            p = Popen(
-                ["ps", "--ppid", str(pid)],
-                stdout=PIPE,
-                creationflags=PROC_CREATIONFLAGS,
-            )
+            if os.name == "nt":
+                raise AssertionError("Bug: This callback would be ineffective and unsafe on Windows, stopping.")
+            p = Popen(["ps", "--ppid", str(pid)], stdout=PIPE)
             child_pids = []
             if p.stdout is not None:
                 for line in p.stdout:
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot for these improvements, they are such a step up, while laying the ground work for additional cleanup (if and when it happens).

PR #1765, opened 2023-12-12
Extract all "import gc" to module level

EliahKagan:
The `gc` module [was already imported](https://github.com/gitpython-developers/GitPython/blob/68272aa0028c35bd4bf36ebfd0fb55ac292b163d/git/repo/base.py#L8) at module scope in `git/repo/base.py`. Although it was moved recently, it is not a new addition. It has been present for about seven years, since f1a82e4 (#555). Importing the top-level git module or any submodule of it runs that import statement:

```python
>>> import sys
>>> "gc" in sys.modules
False
>>> import git
>>> "gc" in sys.modules
True
```

Because the `gc` module is already imported, reimporting it is fast.

Imports that there is no specific reason to do locally should be at module scope. Having them local decreased readability, in part because of how black inserts a black line between them and `gc.collect()` calls they are imported to allow.

An alternative to this change would be to remove the preexisting top-level `import gc` (there is also another one in the test suite) and replace it with a local import as well. I am unsure if that would affect performance and, if so, whether the effect would be good or bad, since the small delay of the import might potentially be less desirable to an application if it occurs while the work of the application is already in progress.

If a `gc.collect()` call runs as a consequence of a finally block or `__del__` method being called during interpreter shutdown, then in (very) rare cases the variable may have been set to None. But this does not appear to have been the intent behind making the imports local. More importantly, a local import should not be expected to succeed, or the imported module usable, in such a situation.

---

Conceptually unrelated to this, it looks like the newly released version of `black` fails to install on Cygwin in the `test-cygwin.yml` workflow. I expect failures from that workflow for this reason. Although unrelated, there is of course no rush on this PR, and if that does happen, it would prevent any Cygwin tests from running until a fix or workaround is applied. If that is fixed first, this PR could be rebased. If this problem doesn't happen, then it is resolved or somehow specific to the setup in my fork and on my local machine, in which case probably nothing has to be done about it.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -1,6 +1,7 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
+import gc
 from io import BytesIO
 import logging
 import os
@@ -1079,8 +1080,6 @@ class Submodule(IndexObject, TraversableIterableObj):
                     self._clear_cache()
                     wtd = mod.working_tree_dir
                     del mod  # Release file-handles (Windows).
-                    import gc
-
                     gc.collect()
                     rmtree(str(wtd))
                 # END delete tree if possible
--- a/test/performance/test_commit.py
+++ b/test/performance/test_commit.py
@@ -5,6 +5,7 @@
 
 """Performance tests for commits (iteration, traversal, and serialization)."""
 
+import gc
 from io import BytesIO
 from time import time
 import sys
@@ -17,8 +18,6 @@ from test.test_commit import TestCommitSerialization
 
 class TestPerformance(TestBigRepoRW, TestCommitSerialization):
     def tearDown(self):
-        import gc
-
         gc.collect()
 
     # ref with about 100 commits in its history.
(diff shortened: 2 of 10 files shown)


Byron (approved):
Thanks a lot for the cleanup!

It looks like cygwin is now broken, which is a little strange as I would expect them to test on cygwin. Maybe that implies it's something to be fixed here?

In any case, it seems like no PR can be merged if this cygwin failure is now a common occurrence, but maybe they will fix it soon enough and sitting it out is the best way of handling this.

EliahKagan:
> In any case, it seems like no PR can be merged if this cygwin failure is now a common occurrence, but maybe they will fix it soon enough and sitting it out is the best way of handling this.

I have no objection to waiting. However, my guess is that the cause is sufficiently weird that a fix may take a while. I also somewhat doubt much testing is done on Cygwin. Since, as is expected to be the case for `ruff`, it's probably not necessarily to install `black` on Cygwin at all. I've proposed this in #1766, with a more detailed rationale and information on the problem.

Byron:
Thanks a lot! With the `black`-fix PR merged and in `main`, I think merging this PR will be fine.

PR #1766, opened 2023-12-12
Don't install black on Cygwin

EliahKagan:
This makes the `black` test dependency apply only to systems *other* than Cygwin. It does not block `black` from being separately installed. It does not reconfigure or otherwise affect pre-commit hooks. This is to work around a problem where the newest stable version of `black` fails to install on Cygwin. Pinning `black` to the previous stable version specifically on Cygwin would also work, but this is simpler and it is not obvious that there is much of a benefit to installing `black` (automatically) on Cygwin.

As of black 23.12.0 and [as seen](https://github.com/gitpython-developers/GitPython/actions/runs/7181278013/job/19555344996) in #1765 ([d70ba69](https://github.com/gitpython-developers/GitPython/pull/1765/commits/d70ba6941cf0f005f3f247c70233f51a0cc35969)), installing `black` builds a wheel on Cygwin, downloading dependencies for the "d" extra (i.e., `black[d]`), which provides dependencies for running the `blackd` daemon. This includes dependencies, such as `asyncio`, whose compilation fails on CI as it is currently set up, fails under some common Cygwin setups (since it needs a C compiler, the Python headers, and possibly other dependencies), and even if successful would cause the Cygwin test workflow to take significantly longer to run. This consistently occurs on this project's CI and on my local Cygwin setup.

A better solution may be preferred, especially if some users are actually running black on Cygwin (though it can be installed at a previous version, if acceptable, in the virtual environment). So this may be viewed as a stopgap measure until a better solution is put in place.

I do not know if the problem is due to a bug in Black. Although [some changes have been made](https://github.com/psf/black/compare/23.11.0...23.12.0) in Black's `pyproject.toml` since the previous release, including to its dependencies in the "d" group, it does not look like these changes should cause this to happen. Inspection of the available wheels on PyPI for both versions ([23.11.0](https://pypi.org/project/black/23.11.0/#files), [23.12.0](https://pypi.org/project/black/23.12.0/#files)) also does not point to an obvious explanation, though I could be missing something.

It seems no similar situation occurs on other systems, even with Python 3.9 (the version this project tests with on Cygwin).

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,4 +1,4 @@
-black
+black ; sys_platform != "cygwin"
 coverage[toml]
 ddt >= 1.1.1, != 1.4.3
 mock ; python_version < "3.8"


Byron (approved):
Yes, let's start by getting CI working again, and see if eventually it makes sense to re-enable it or its substitute.

Thanks again for your help!

EliahKagan:
Sounds good. To clarify, although `ruff` recently [does offer auto-formatting](https://docs.astral.sh/ruff/formatter/), which is mostly `black`-compatible, and I'd love to replace `black` with `ruff` eventually, I don't propose to do so immediately even once `ruff` is working: my focus for introducing `ruff` to GitPython is with the idea of doing linting replacing `flake8` plus, perhaps, various further checks (that could be added over time).

EliahKagan:
Good news: The bug has been reported and analyzed in https://github.com/psf/black/issues/4107. It seems the developers of `black` understand the cause (which appears to be a bug in a dependency management tool) and know how to fix it.

Once the bug is fixed, I'm thinking this PR could be reverted, since having fewer special cases decreases complexity, and because maybe someone is running `black` on GitPython code on Cygwin and would benefit from the convenience of it being installed. I should be able to open a PR to revert this one, when the time comes. But please let me know if this is not the approach you would prefer.

Another possibility might be to continue in this direction for other platforms, and have `black` automatically installed only via `pre-commit`, as `flake8` and its plugins currently are. However, this would require some changes to the readme, and also might make things a bit less convenient for users who use `black` integration in editors/IDEs but don't have `black` installed globally. It would have the benefit of making the test dependencies closer to the request in https://github.com/gitpython-developers/GitPython/issues/1744#issuecomment-1827863090, which I am guessing some other downstream maintainers might also prefer. (But it would not be feasible to go all the way, because GitPython's own CI uses one of the `pytest` plugins mentioned there.)

Byron:
Thanks for sharing, good news indeed!

> I should be able to open a PR to revert this one, when the time comes.

Yes, let's bring `black` back to cygwin when the fix has landed.

> Another possibility might be to continue in this direction for other platforms,

Here it seems easier to keep everything as is, but to consider changes if more feedback points in other directions.

EliahKagan:
A new version of `black` with the fix is released, so I've opened #1783 to revert this PR.

Issue #1762, opened 2023-12-10
Git.execute output_stream is not documented to suppress kill_after_timeout, but does


EliahKagan:
When calling `Git.execute`, including indirectly through the dynamic methods of a `Git` object, passing a non-`None` value of `output_stream` suppresses the effect of `kill_after_timeout`. [The `threading.Timer` object is created](https://github.com/gitpython-developers/GitPython/blob/a58a6be043f62e6552ef1638ec497f3b805c5480/git/cmd.py#L1040-L1042) with the local `kill_process` function as its callback, but `start` is never called on it.

https://github.com/gitpython-developers/GitPython/blob/a58a6be043f62e6552ef1638ec497f3b805c5480/git/cmd.py#L1050-L1079

This situation appears not entirely intentional, because:

- The "watchdog" `Timer` is created, even though it is never used.
- The docstring, even as it documents [that `as_process` suppresses `kill_after_timeout`](https://github.com/gitpython-developers/GitPython/blob/a58a6be043f62e6552ef1638ec497f3b805c5480/git/cmd.py#L880) (and that [`as_process` suppresses `output_stream`](https://github.com/gitpython-developers/GitPython/blob/a58a6be043f62e6552ef1638ec497f3b805c5480/git/cmd.py#L867)), does not mention any effect of `output_stream` on `kill_after_timeout`.

So either `kill_after_timeout` should be honored even when `output_stream` is used, *or* the docstring should be updated to mention that it does not (and the code possibly modified to not create the `Timer` object in the case that it is unused).

I am not sure which change should be made, so I'm opening this rather than a PR to propose one of them.

This is effectively separate from #1756. (One quality this shares with #1756 is that it is relevant only to the behavior of `Git.execute` and functions that use it. Functions that use `handle_process_output` are unaffected; its `kill_after_timeout` is implemented differently with nonequivalent behavior and separate code. `handle_process_output` also does not take an `output_stream` argument.)

Byron:
Thanks for bringing this to my attention - I definitely wasn't aware.

Generally, if the user obtains a running process, then they are responsible for handling it no auto-killing should happen. Thus I think the current behaviour, also in relation to `output_stream` should be documented. The code could also be cleaned up so that the timer isn't created in the first place in that case.

Maybe we can consider to emit a warning if incompatible flags are specified to make people aware that their `kill_after_timeout` setting isn't effective.

[...]

PR #1767, opened 2023-12-12
Replace xfail with gc.collect in TestSubmodule.test_rename

EliahKagan:
In various places in the code, a conditional `gc.collect()` is used to allow a test to finish its work (and pass). This adds one more such place. It accordingly removes an xfail marking. `TestSubmodule.test_rename` now passes on all combinations of platform and Python version.

Like the xfail condition, running `gc.collect()` here is conditional, being done only in the specific situation the `PermissionError` occurs. Besides that it does not always run even on Windows (only in 3.12 and later), this resembles various other conditional and non-conditional gc.collect calls.

The situation is a bit odd, but I think that's--albeit by a fairly slight margin--more of a reason to convert the xfail into a `gc.collect()` than to refrain from doing so.

#### Context

#1745 (under "Which versions should we test?"), and also 82c361e, 0b7ee17, and various other less interesting commits, provide further context about this test and the `PermissionError` on Windows that does not occur prior to 3.12.

Because `PermissionError` error message is about another process having the file open, #790 seems relevant. But this does not happen in a call to `git.util.rmtree`.

#### One call is enough, now

It had previously appeared to me that two calls to `gc.collect` were required, but I am unable to reproduce that. It may have been specific to how I was running it on my system at that time. The need for only one call may have been brought about by changes to the code in the mean time, but I have tested that only one call appears required even without the changes in #1765. (Perhaps #1753 contributed? I think most likely not, though.)

### DIFF
--- a/test/test_submodule.py
+++ b/test/test_submodule.py
@@ -948,18 +948,6 @@ class TestSubmodule(TestBase):
         sm.remove()
         assert not sm.exists()
 
-    @pytest.mark.xfail(
-        os.name == "nt" and sys.version_info >= (3, 12),
-        reason=(
-            "The sm.move call fails. Submodule.move calls os.renames, which raises:\n"
-            "PermissionError: [WinError 32] "
-            "The process cannot access the file because it is being used by another process: "
-            R"'C:\Users\ek\AppData\Local\Temp\test_renamekkbznwjp\parent\mymodules\myname' "
-            R"-> 'C:\Users\ek\AppData\Local\Temp\test_renamekkbznwjp\parent\renamed\myname'"
-            "\nThis resembles other Windows errors, but only occurs starting in Python 3.12."
-        ),
-        raises=PermissionError,
-    )
     @with_rw_directory
     def test_rename(self, rwdir):
         parent = git.Repo.init(osp.join(rwdir, "parent"))
--- a/test/test_submodule.py
+++ b/test/test_submodule.py
@@ -958,6 +958,12 @@ class TestSubmodule(TestBase):
         assert sm.rename(sm_name) is sm and sm.name == sm_name
         assert not sm.repo.is_dirty(index=True, working_tree=False, untracked_files=False)
 
+        # This is needed to work around a PermissionError on Windows, resembling others,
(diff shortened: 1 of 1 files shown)


Byron (approved):
That's a great catch! Who would have thought that `gc.collect()` is also applicable here.

This probably also hints at the need to eventually fix the way GitPython handles systems resources. After all, it really seems like one would rather want to use…

```python
with git.open(path) as repo:
   pass
```

…to control the lifetime of all opened resources precisely. Alternatively, GitPython would not be allowed to keep anything open past the lifetime of a method call, but that's not feasible at all. Maybe there is alternatives that I am not seeing.

It's interesting how it's obvious that GitPython was designed in a time when there was no GC, but rather, reference counting with deterministic and immediate calls of destructors.

EliahKagan:
This sounds like it might be part of making the classes that currently have `__del__` methods also be context managers, as discussed in comments in #1669.

> It's interesting how it's obvious that GitPython was designed in a time when there was no GC, but rather, reference counting with deterministic and immediate calls of destructors.

I have some thoughts about how the effects of this might be mitigated, to some extent, even for existing code that would not use newly added context managers (or perhaps even before, or separately from, making more classes context managers). But right now these are half-formed ideas that might turn out to be meritless and (as with #1762) there's some other GitPython-related stuff I'd probably want to finish dealing with first. I mention this mainly so I don't forget or, if I do, so that it is more likely to come up again.

Byron:
> This sounds like it might be part of making the classes that currently have `__del__` methods also be context managers, as discussed in comments in #1669. Though I suppose this usage need not wait for that, with `Repo` having been a context manager since #555. Should the tests be modified to make greater use of that? I'm not sure it would address the sort of thing that happened here and it's not quite what your code example shows.
> 
> >

I left some more ideas in [this comment](https://github.com/gitpython-developers/GitPython/issues/1669#issuecomment-1853531749), and based on it I'd say the tests should be kept as is. Should a full-on context-manager approach happen, the tests can be a great indicator to show how hard the migration would be for existing users, and also guide towards making it as painless as possible.

PR #1768, opened 2023-12-12
Extract remaining local "import gc" to module level

EliahKagan:
In #1765, I missed one, somehow! This fixes that.


Byron:
Thank you!

PR #1770, opened 2023-12-13
Replace some uses of the deprecated mktemp function

EliahKagan:
`tempfile.mktemp` is deprecated for security reasons and the race condition inherent to its use can have an impact on robustness even in situations where the security impact is small or nonexistent. I do not believe the changes made here are fixing security vulnerabilities; I was not able, at least so far, to think of likely ways they could be exploited. Nonetheless I believe these changes are justified, as detailed in the commit messages.

Only one of the uses of `mktemp` was in the `git` module itself. Others are in the test suite. There are a few more in the test suite where I am unsure what the best approach is to replacing them, though I hope to propose a change to eliminate them in the future. All of these considerations, as well as the general issues surrounding `mktemp` and its deprecation, are detailed in the two commit messages.

- 41fac85 makes changes in the test suite.
- 9e86053 changes code in the `git` module.

See also:

- https://github.com/gitpython-developers/smmap/issues/41, which is also about `mktemp`. (I've [proposed a fix](https://github.com/gitpython-developers/smmap/pull/56).) That looks like the only use of `mktemp` in smmap. There are some in gitdb as well, and (as noted) some in GitPython that this PR does not cover.
- #1769 - If you decide both to enable CodeQL and to merge this PR, and they are done in that order, then the effects of this PR will be reflected as changes to the "Code scanning" alerts.

### DIFF
--- a/git/index/util.py
+++ b/git/index/util.py
@@ -3,6 +3,7 @@
 
 """Index utilities."""
 
+import contextlib
 from functools import wraps
 import os
 import os.path as osp
@@ -40,12 +41,10 @@ class TemporaryFileSwap:
 
     def __init__(self, file_path: PathLike) -> None:
         self.file_path = file_path
-        self.tmp_file_path = str(self.file_path) + tempfile.mktemp("", "", "")
-        # It may be that the source does not exist.
-        try:
-            os.rename(self.file_path, self.tmp_file_path)
-        except OSError:
-            pass
+        fd, self.tmp_file_path = tempfile.mkstemp(prefix=self.file_path, dir="")
+        os.close(fd)
+        with contextlib.suppress(OSError):  # It may be that the source does not exist.
+            os.replace(self.file_path, self.tmp_file_path)
 
     def __enter__(self) -> "TemporaryFileSwap":
         return self
@@ -57,10 +56,7 @@ class TemporaryFileSwap:
         exc_tb: Optional[TracebackType],
     ) -> bool:
         if osp.isfile(self.tmp_file_path):
-            if os.name == "nt" and osp.exists(self.file_path):
-                os.remove(self.file_path)
-            os.rename(self.tmp_file_path, self.file_path)
-
+            os.replace(self.tmp_file_path, self.file_path)
         return False
 
 
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -89,8 +89,7 @@ def with_rw_directory(func):
 
     @wraps(func)
     def wrapper(self):
-        path = tempfile.mktemp(prefix=func.__name__)
-        os.mkdir(path)
+        path = tempfile.mkdtemp(prefix=func.__name__)
         keep = False
         try:
             return func(self, path)
--- a/test/performance/lib.py
+++ b/test/performance/lib.py
@@ -65,8 +65,7 @@ class TestBigRepoRW(TestBigRepoR):
     def setUp(self):
         self.gitrwrepo = None
         super().setUp()
-        dirname = tempfile.mktemp()
-        os.mkdir(dirname)
+        dirname = tempfile.mkdtemp()
         self.gitrwrepo = self.gitrorepo.clone(dirname, shared=True, bare=True, odbt=GitCmdObjectDB)
         self.puregitrwrepo = Repo(dirname, odbt=GitDB)
 
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -68,12 +68,11 @@ class TestBase(_TestBase):
             data = data_stream.read()
             assert data
 
-            tmpfilename = tempfile.mktemp(suffix="test-stream")
-            with open(tmpfilename, "wb+") as tmpfile:
+            with tempfile.NamedTemporaryFile(suffix="test-stream", delete=False) as tmpfile:
                 self.assertEqual(item, item.stream_data(tmpfile))
                 tmpfile.seek(0)
                 self.assertEqual(tmpfile.read(), data)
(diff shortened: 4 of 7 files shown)


git/index/util.py
```diff
@@ -40,12 +41,10 @@ class TemporaryFileSwap:
 
     def __init__(self, file_path: PathLike) -> None:
         self.file_path = file_path
-        self.tmp_file_path = str(self.file_path) + tempfile.mktemp("", "", "")
-        # It may be that the source does not exist.
-        try:
-            os.rename(self.file_path, self.tmp_file_path)
-        except OSError:
-            pass
+        fd, self.tmp_file_path = tempfile.mkstemp(prefix=self.file_path, dir="")
```
Byron, line 44: Just as a note, I am of course note thrilled that this adds at least two additional system calls, one for opening the tempfile, and one for closing it.

Using `mkstemp` this way is most certainly an anti-pattern, but I also believe that this is just a symptom of the way temporary files are used in the codebase.

In `gitoxide`, for instance, temporary files are created safely just like this, and typically kept open to be written to. Then, when ready, they will be renamed into place. I don't really understand why GitPython evolved this kind of usage, but it's good to know there are alternatives.

The reason I also don't consider this a big deal after all is that the immaturity of this `index` implementation hopefully deters most from using it in the first place.

Byron (approved):
Thanks again for making these improvements, and I will be curious how this impacts the newly added CodeQL analysis and the related issues.

I left a note, which I hope is somewhat useful for context at least.

git/index/util.py
EliahKagan: I would say there are two things bad about that specific part of the code I introduced here:

#### The extra system calls

I believe that creating a new empty file always involves overhead of at least two system calls in that the file is opened and then closed. (The file could be closed by the system as a result of cleaning up after a process that opens it without closing it, but using a separate process would introduce far more overhead.)

Here, the file is created to reserve its name. When the file is opened to create it, and then needs to be separately opened and written, by this or another process, I believe the overhead cannot be avoided, regardless of language or framework. Instead, to avoid it, we would need to avoid creating the file ahead of time, while still avoiding the original race condition, by moving the logic that generates possible names and tries them into the code that uses the file.

That should be feasible. It certainly is feasible on Windows, where `os.rename` fails (and is documented to do so) when its target already exists. I believe there should be a reasonable way to do it on POSIX systems as well, but this hinges on whether a way of renaming a file that *atomically* succeeds or fails depending on the target's nonexistence is available. (In some other areas, where a subprocess must reopen the file, I believe it is not feasible to do better. That applies to the related [changes to `git/index/base.py`](https://github.com/gitpython-developers/GitPython/pull/1751/files#diff-56d167dc0ca4d51682e56b5e602c096f3f264fd1ebb61cc03bbb6c409c866eeb) in #1751, but I believe does not apply here.)

The issue is that, even if it can be done correctly, this would require writing more code. Generating names and retrying, as `mkstemp` does, would have to be implemented in GitPython. However, if you consider this valuable, I can open an issue about it. Then perhaps someone will implement it and, if not, there is a substantial chance I may eventually get to it.

#### Not supporting `Path` for the prefix

Intuitively it feels like `os.mkstemp` should support both `str` and a `str`-based `Path` for `prefix`, but it actually only supports `str`. That's a problem because:

https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/index/util.py#L42

I am unsure when this regression would cause a problem in GitPython's own use of `TemporaryFileSwap`. Its only use in GitPython is in `git.index.base.IndexFile.from_tree`, which is used in several places. It looks like it may be able to pass a `Path` in some cases, but I'm not sure. In addition to any some such circumstances in GitPython, the `TemporaryFileSwap` class is public in the `git.index.util` module, and projects that use GitPython may be using it directly.

I'll try to open a pull request today to fix this. When doing so, if I can, I'll try and make the `mkstemp` call slightly less unidiomatic and weird, by using `dir=` for the directory part that *can* be a `Path`, and `prefix=` just for the filename prefix part. If I can't propose a fix soon, then I'll extract this section of the comment to a new issue.

EliahKagan: Sorry, the way I put that was very unclear. "When the file is opened to create it, and then needs to be separately opened and written" is the condition where I was saying, *if* that condition holds, then the extra work cannot be avoided. "This" and "another process" did not really need to be distinguished; in my mind, they were different cases, because it is only in the "other process" case that this unfortunate condition may not be able to be avoided.

The "other process" case that came up in GitPython was #1751. But I should acknowledge that even that might sometimes be avoided, for example if `git` were to replace the original whose name is known beforehand, rather than the copy whose name can only safely be decided by creating it. Perhaps this is part of what you were referring to (though my guess is that gitoxide doesn't have to deal with this at all because it doesn't use a `git` subprocess for index operations). I don't know if such a change would be feasible in GitPython, and in case I will assume it's not worthwhile to pursue it based on what you've said about `index`.

---

In terms of the general idea, what had I meant was that *if* we need to decide an exact filename we are going to write to *before* it is to be opened for write, but it is a temporary file and we cannot know an available name before creating the file, then we will suffer at least one of the following:

- A race condition on the name in the filesystem (and maybe a system call for the immediately stale initial check, as `mktemp` does).
- Two system calls to create the file, unless the system provides a way to atomically attempt to create a file that fails when it already exists, other than by opening the file.

As far as I am aware, getting around this always involves weakening the premise (the condition I had very unclearly sought to describe earlier) through design, though in some designs one achieves the goal without having to think about it.

PR #1774, opened 2023-12-19
Run some Windows environment variable tests only on Windows

EliahKagan:
These are a couple changes to simplify and clarify the tests in `test_util.py`.

- Use `Path.touch` instead of `Path.write_bytes(b"")` in the tests. I guess I had formerly had the idea that `pathlib` gained `Path.touch` recently, but that is not at all the case. (This applies to all operating systems, but it is the more minor of the two changes.)
- Run some tests of Windows-specific environment variables on Windows only, and simplify the tests accordingly. The reason I think this makes sense now, even though it wouldn't have been the best idea when those tests were first introduced, is detailed in the [6a8ed70](https://github.com/gitpython-developers/GitPython/pull/1774/commits/6a8ed70a6003c13800d908ab055038eb61ba4ce9) message.

The latter is one of the changes listed in the "What is *not* done here?" section of #1745. Or it is part of that--perhaps some of the other tests, more directly concerning `rmtree`, can also be simplified, I'm not sure. (Right now I'm thinking that will become clearer if it is considered together with further progress toward #790 rather than beforehand.)

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -46,7 +46,7 @@ def permission_error_tmpdir(tmp_path):
     """Fixture to test permissions errors in situations where they are not overcome."""
     td = tmp_path / "testdir"
     td.mkdir()
-    (td / "x").write_bytes(b"")
+    (td / "x").touch()
 
     # Set up PermissionError on Windows, where we can't delete read-only files.
     (td / "x").chmod(stat.S_IRUSR)
@@ -73,7 +73,7 @@ class TestRmtree:
             td / "s" / "y",
             td / "s" / "z",
         ):
-            f.write_bytes(b"")
+            f.touch()
 
         try:
             rmtree(td)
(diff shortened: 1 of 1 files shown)


Byron (commented):
Thank you! Definitely an improvement!

I also wonder why `touch()` wasn't used before, and don't really have an explanation for it 😁.

EliahKagan:
Well, the code that should've used it but didn't was also code written by me, For whatever reason, I had not thought to use it, or had not thought I could use it. I think I had gotten the wrong idea that it wasn't supported on all versions (it is, though).

Byron:
Oh, how embarrassing 🤦‍♂️, I actually thought this lack of *touch* was still in the pre-Eliah codebase 😅. Despite reading your comment about it in the PR text, I thought that was the reason you didn't *change* it yet, even though you were the original *creator*. I definitely didn't mean to question you, but rather, my probably 15 years younger self 😁.

Issue #1631, opened 2023-08-22
Error when handling non-ascii filename


lubomir:
Steps to reproduce:

```
$ mkdir foo
$ cd foo
$ git init
$ touch 'bokm'$'\345''l.alias'
$ python -c 'import git; git.Repo(".").untracked_files'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/git/repo/base.py", line 839, in untracked_files
    return self._get_untracked_files()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/git/repo/base.py", line 856, in _get_untracked_files
    filename = filename.encode("ascii").decode("unicode_escape").encode("latin1").decode(defenc)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe5 in position 4: invalid continuation byte
```

I can reproduce with 3.1.30 and 3.1.32.

A file with such name is present in ftp://ftp.gnu.org/gnu/aspell/dict/nb/aspell-nb-0.50.1-0.tar.bz2
I'm not sure what encoding it's supposed to be. Firefox detects it as windows-1252 (and displays it correctly).

Byron:
The handling of filepaths is completely broken in `GitPython` as it tries to handle it like it's strings even though these imply a known encoding.

This could definitely use an overhaul, these issues are in many places.

pushfoo:
> The handling of filepaths is completely broken

1. How much of this is Python 2 legacy?
2. Are there more detailed notes on how / why?
3. Is there an external dependency policy of any sort restricting what can be used beyond license compatibility?

Byron:
Probably most of it is the way it is due to a lack of understanding of how to handle paths in general and in git specifically. Most problems occour merely by trying to decode bytes into something that can be used in python even though one is dealing with either OS paths or bundles of bytes produced by the `git` binary itself.

I don't think there is a limit of what could be used in terms of external dependency if it has a compatible license *and* is truly required. I'd hope python can handle OS paths properly out of the box by now.

Issue #1362, opened 2021-10-19
Error using Chinese path or Chinese branch name


xunmi-git:
# Environmental Information
Windows10
Python3.8
GitPython 3.1.24

# Error 1
## Chinese path: 
`git.Repo.clone_from(url="git@gitee.com:xun-mi-git/test-x.git", to_path="C:\\Users\\asus\\3D Objects\\个人项目\\test")`
## Error message
```python
Exception in thread Thread-6:
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 1370, in _readerthread
    buffer.append(fh.read())
UnicodeDecodeError: 'gbk' codec can't decode byte 0xaa in position 41: illegal multibyte sequence
```
## screenshot
![寻觅_20210619020652](https://user-images.githubusercontent.com/43723838/137853118-b819e5ad-bdeb-46f0-adbf-ec9589d56f82.png)

# Error 2
## Chinese branch: 
```python
    git.Repo.clone_from(url="git@gitee.com:xun-mi-git/test-x.git", to_path="C:\\test")
    a = Repo("C:\\test").remotes[0]
    for i in a.refs:
        print(i.name)
    a = a.pull()
```
## Error message
```python
origin/HEAD
origin/默认分支
Exception in thread Thread-7:
Traceback (most recent call last):
  File "C:\Users\asus\3D Objects\Python环境\常用通用库\lib\site-packages\git\cmd.py", line 109, in pump_stream
    for line in stream:
UnicodeDecodeError: 'gbk' codec can't decode byte 0xae in position 58: illegal multibyte sequence

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #1776, opened 2023-12-21
Fix TemporaryFileSwap regression where file_path could not be Path

EliahKagan:
As noted in https://github.com/gitpython-developers/GitPython/pull/1770#discussion_r1433438311, in 9e86053 (#1770) I accidentally introduced a regression in `git.index.util.TemporaryFileSwap` where the `file_path` parameter could no longer be a `Path`, even though it was (and remained) annotated in a way compatible with that.

Passing a `Path` object instead of `str` failed with `TypeError`, with a message like this (with `WindowsPath` instead of `PosixPath` on Windows, but otherwise the same):

```text
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

Although much code that results in `TemporaryFileSwap` being used is under test, which would reveal some kinds of regressions, `TemporaryFileSwap` itself did not have any tests. It is public, so even if a `Path` were never passed to it in GitPython's own use of it, applications that use GitPython may be making direct use of it and are justified in assuming it accepts both a `str` and a `str`-based `Path`, as annotated.

This pull request:

- **Adds a general test of `TemporaryFileSwap`.** It does not test every code path that is supposed to be supported, such as code paths involving the file not existing in the first place, or the temporary file being deleted before it would be swapped back (those are deliberately covered in the code and may thus be valuable to test, and perhaps I, or someone else, could add tests for that later). However, I parameterized the test by the type of the `file_path` argument passed to `TemporaryFileSwap`, and the test shows both that it works when the type is `str` and that it fails, as described above, when the type is `Path`.
- **Fixes the bug**, and also changes how `mkstemp` is called to make it less unidiomatic. (However, it is still opening the file to reserve its name, and then immediately closing it, so this does not address the original misgivings noted in [#1770 (review)](https://github.com/gitpython-developers/GitPython/pull/1770#pullrequestreview-1779282078). See my comment in that review thread.)

There are some further details, which I'd consider less essential but potentially of value, in the commit messages.

### DIFF
--- a/git/index/util.py
+++ b/git/index/util.py
@@ -41,7 +41,7 @@ class TemporaryFileSwap:
 
     def __init__(self, file_path: PathLike) -> None:
         self.file_path = file_path
-        fd, self.tmp_file_path = tempfile.mkstemp(prefix=self.file_path, dir="")
+        fd, self.tmp_file_path = tempfile.mkstemp(prefix=str(self.file_path), dir="")
         os.close(fd)
         with contextlib.suppress(OSError):  # It may be that the source does not exist.
             os.replace(self.file_path, self.tmp_file_path)
--- a/git/index/util.py
+++ b/git/index/util.py
@@ -41,7 +41,8 @@ class TemporaryFileSwap:
 
     def __init__(self, file_path: PathLike) -> None:
         self.file_path = file_path
-        fd, self.tmp_file_path = tempfile.mkstemp(prefix=str(self.file_path), dir="")
+        dirname, basename = osp.split(file_path)
+        fd, self.tmp_file_path = tempfile.mkstemp(prefix=basename, dir=dirname)
         os.close(fd)
         with contextlib.suppress(OSError):  # It may be that the source does not exist.
             os.replace(self.file_path, self.tmp_file_path)
--- a/test/test_blob_filter.py
+++ b/test/test_blob_filter.py
@@ -14,14 +14,15 @@ from git.objects import Blob
 from git.types import PathLike
 
 
-# fmt: off
-@pytest.mark.parametrize('paths, path, expected_result', [
-    ((Path("foo"),), Path("foo"), True),
-    ((Path("foo"),), Path("foo/bar"), True),
-    ((Path("foo/bar"),), Path("foo"), False),
-    ((Path("foo"), Path("bar")), Path("foo"), True),
-])
-# fmt: on
+@pytest.mark.parametrize(
+    "paths, path, expected_result",
+    [
+        ((Path("foo"),), Path("foo"), True),
+        ((Path("foo"),), Path("foo/bar"), True),
+        ((Path("foo/bar"),), Path("foo"), False),
+        ((Path("foo"), Path("bar")), Path("foo"), True),
+    ],
+)
 def test_blob_filter(paths: Sequence[PathLike], path: PathLike, expected_result: bool) -> None:
     """Test the blob filter."""
     blob_filter = BlobFilter(paths)
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -34,10 +34,11 @@ from git.exc import (
 )
 from git.index.fun import hook_path
 from git.index.typ import BaseIndexEntry, IndexEntry
+from git.index.util import TemporaryFileSwap
 from git.objects import Blob
-from test.lib import TestBase, fixture, fixture_path, with_rw_directory, with_rw_repo
 from git.util import Actor, hex_to_bin, rmtree
 from gitdb.base import IStream
+from test.lib import TestBase, fixture, fixture_path, with_rw_directory, with_rw_repo
 
 HOOKS_SHEBANG = "#!/usr/bin/env sh\n"
 
@@ -1087,3 +1088,25 @@ class TestIndex(TestBase):
         file.touch()
 
         rw_repo.index.add(file)
+
+
+class TestIndexUtils:
+    @pytest.mark.parametrize("file_path_type", [str, Path])
+    def test_temporary_file_swap(self, tmp_path, file_path_type):
+        file_path = tmp_path / "foo"
+        file_path.write_bytes(b"some data")
+
+        with TemporaryFileSwap(file_path_type(file_path)) as ctx:
+            assert Path(ctx.file_path) == file_path
+            assert not file_path.exists()
+
+            # Recreate it with new data, so we can observe that they're really separate.
+            file_path.write_bytes(b"other data")
+
+            temp_file_path = Path(ctx.tmp_file_path)
+            assert temp_file_path.parent == file_path.parent
+            assert temp_file_path.name.startswith(file_path.name)
+            assert temp_file_path.read_bytes() == b"some data"
+
+        assert not temp_file_path.exists()
+        assert file_path.read_bytes() == b"some data"  # Not b"other data".


Byron:
Thanks a lot for the fix and its discovery in the fist place!

Of course one would expect that each standard library item that is handling items in the file system uses `PathLike` to not enforce an encoding on the provided input, but maybe some time in the future that will all be fixed in Python itself.

EliahKagan:
> Of course one would expect that each standard library item that is handling items in the file system uses `PathLike` to not enforce an encoding on the provided input, but maybe some time in the future that will all be fixed in Python itself.

Somewhat related to this topic, I've recently posted https://github.com/gitpython-developers/GitPython/discussions/1374#discussioncomment-7928044. If my understanding as expresed there is mistaken or otherwise confused, I'd be pleased to make corrections.

I don't fully understand why `PathLike` would avoid enforcing an encoding, or why an encoding would have to be enforced when using `str`. Part of my confusion is that, in the context of GitPython, there are two things `PathLike` could refer to.

- I think you mean the `PathLike` this PR concerned, which is `git.types.PathLike` and is [defined](https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/types.py#L41) (though using an older notation for backward compatibility) as **`str | os.PathLike[str]`**.
- The other possibility is [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) itself, which covers any type that provides a `__fspath__` method that can be called with no arguments (besides `self`) and returns a `str` or `bytes` object. In annotations, when it is guaranteed to return `str`, `os.PathLike[str]` can be used, and when it is guaranteed to return `bytes`, `os.PathLike[bytes]` can be used.

(`str | os.PathLike[str]` is not redundant, because `str`, like `bytes`, does not itself have a `__fspath__` method.)

My understanding of the current recommendation in Python for filesystem operations is to use `str`-based types for paths whenever possible, because this is more portable and avoids assuming any encoding. I infer this from [these official tips](https://docs.python.org/3/howto/unicode.html#tips-for-writing-unicode-aware-programs) taken together with some specific points about paths such as the note [in the `pathlib.Path` docs](https://docs.python.org/3/library/pathlib.html#operators) that "Calling `bytes` is only recommended under Unix. Under Windows, the unicode form is the canonical representation of filesystem paths."

Both `str` itself and types that type-check as `os.PathLike[str]`, such as a `Path[str]`, are `str`-based. As far as I am aware, there are no encoding differences between what `str` and `Path[str]` store and express.

On Windows, there is occasionally the further issue that a "Unicode string" in the Windows API, though intended always to be valid UTF16-LE--which Python converts to and from `str` behind the scenes, I believe in practice performing a conversion to and from UTF-8 at least in CPython--can contain any sequence of 16-bit `wchart_t` values (`wchar_t` is 2 bytes on Windows, unlike the usual 4), including invalid surrogate pairs. Then conversion to UTF-8 may fail or produce something incorrect.

I think this ties in with some of what was [discussed there](https://github.com/gitpython-developers/GitPython/pull/1672#discussion_r1333061045).

One thing I am unsure about is if `git` can be reliably be made to disregard Windows's active code page settings and output UTF-8. I expect this to be relevant whether `git` is run with its standard streams treated as text or as binary data, since with text the encoding might not be what is expected; and with binary data the encoding, once the data are decoded, which they have to be at least some of the time, might not be what is expected.

Another thing I am unsure of, which represents a bigger gap in my understanding, is how much setting `LC_ALL` and related environment variables to `C` is guaranteed to achieve for `git`, even on POSIX systems. Can data that cannot be represented in ASCII be mangled in this scenario? Certainly this makes the *messages* parseable, in the sense of preventing translated strings from being used, but I am unsure about its effects on filenames, branch and tag names, etc. I know that some systems have added a non-equivalent `C.UTF-8` locale, while others have not.

When I first saw [some code like this](https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/repo/base.py#L913-L914), which I think is is the sort of thing you were [alluding to there](https://github.com/gitpython-developers/GitPython/pull/1672#discussion_r1334083074)--and which I have found by testing on my zh-CN localized Windows Server 2022 virtual machine to be at issue in #1362--my first thought was that this could be solved using the usual approach of both *(a)* using `str` for everything that is conceptually text, everywhere, and *(b)* passing `encoding="utf-8"` (as in #1748) to every function that is being called in a way that may entail the encoding or decoding of text and that accepts an `encoding` argument.

But because I'm not sure this can always in practice be done for interacting with `git` subprocesses, I am getting the idea that fixing things may be much more involved than that. (Perhaps #641 is related to this. We don't have to use the `codecs` module for much anymore, but that seems to be the idea, though I think more code would have to be changed.)

Byron:
Thanks for reminding me that `PathLike` is a GitPython invention which boils down to `str` (it seems). I always hoped PathLike would be something that doesn't boil down to some Python controlled encoding, but instead to an unspecified OS-encoding.

*(More about that to follow in a future paragraph)*

> On Windows, there is occasionally the further issue that a "Unicode string" in the Windows API, though intended always to be valid UTF-16LE--which Python converts to and from `str` behind the scenes, I _believe_ in practice performing a conversion to and from UTF-8 at least in CPython, but possibly using a UTF-32 encoding under some circumstances--can contain any sequence of 16-bit `wchart_t` values (`wchar_t` is 2 bytes on Windows, unlike the usual 4), including invalid surrogate pairs. Then conversion to UTF-8 may fail or produce something incorrect.

It's amazing that you are aware of this as I only learned it recently when digging into the details of [the Rust `OsString`](https://doc.rust-lang.org/std/ffi/struct.OsString.html) type, which led to the very curious finding of [the WTF encoding](https://simonsapin.github.io/wtf-8/) that Rust uses specifically on Windows to allow fastest-possible conversions to `String` if possible.

> Another thing I am unsure of, which represents a bigger gap in my understanding, is how much setting `LC_ALL` and related environment variables to `C` is guaranteed to achieve for `git`, even on POSIX systems. Can data that cannot be represented in ASCII be mangled in this scenario? Certainly this makes the _messages_ parseable, in the sense of preventing translated strings from being used, but I am unsure about its effects on filenames, branch and tag names, etc. I know that some systems have added a non-equivalent `C.UTF-8` locale, while others have not.

For all I know, that variable is used to control the `git` CLI interface language, as I never tried to alter how `git` handles encodings of data. As far as I know, this is pre-determined everywhere, with the addition that paths are treated specifically on Unix (a bunch of bytes, no encoding), MacOS (deal with precomposed unicode) and Windows (convert paths to UTF8, and [ignore failures(!!)](https://github.com/git/git/blob/564d0252ca632e0264ed670534a51d18a689ef5d/compat/win32/dirent.c#L12)).


> When I first saw [some code like this](https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/repo/base.py#L913-L914), which I think is is the sort of thing you were [alluding to there](https://github.com/gitpython-developers/GitPython/pull/1672#discussion_r1334083074)--which is at issue in #1631 and, based on testing on a zh-CN localized Windows Server 2022 virtual machine, seems to be at issue in #1362--my first thought was that this could be solved using the usual approach of both _(a)_ using `str` for everything that is conceptually text, everywhere, and _(b)_ passing `encoding="utf-8"` (as in #1748) to every function that is being called in a way that may entail the encoding or decoding of text and that accepts an `encoding` argument.

Using `str` everywhere would certainly be nice (at least if the encoding is forced/assured to be UTF-8), and maybe at least for handling paths it would be correct most of the time, and more often than now. For all I know, `str` has an encoding though, the python interpreter global encoding, which makes it hard to use in practice. Maybe this changed, or maybe there are details, and all these won't matter as long as GitPython can control the encoding.

Practically, GitPython still won't work correctly though.

What `gitoxide` does is the following:

* If the encoding is unknown or dynamic (like with commit messages), use [`BString/bstr`](https://docs.rs/bstr/latest/bstr/), which makes the string easily printable and processable if it is indeed ascii/UTF-8, but won't enforce any encoding nonetheless. It's a bunch of bytes.
* When dealing with filesystem paths, standard [`Path`-types](file:///Users/byron/.rustup/toolchains/1.69-aarch64-apple-darwin/share/doc/rust/html/std/path/index.html) are used. When going from operating system paths to Git, these will have to be converted to `BString`, which is always fine on Unix, with mild conversion on MacOS and with major, fallible conversion on Windows. It's fallible only because of illformed UTF-8 which possibly still exists, but admittedly `gitoxide` will just panic as it is implemented now (which is probably better than ignoring the issue like `git` seems to do).
* The above conversion will be performed in reverse as needed to get back from `git`-paths to OS paths.

For all I can gather, the best, most unassuming way to deal with paths in GitPython would be to use `bytes`, which is probably terrible to handle. At some point, one will have to assume some encoding to work with these, and in fact `git` does assume ASCII, which is why it's so fine with UTF-8 as well.

That was interesting to me to write all this down here, and I hope it makes enough sense to you to help with mapping any/some/all of it into the Python world. My conclusion would be that we'd want `bstr` in Python.

PR #1778, opened 2023-12-21
Fix if items of Index is of type PathLike

stegm:
The `move' and `remove' index operations accept `PathLike` items, but throw a `TypeError' if any of the items are of that type.
Example:

```python
index = repo.index
index.remove(Path("foo")) # should work and not throw TypeError
```


Byron (approved):
Thanks a lot for the fix :).

The test seems to validate that `Path`-like types are now not failing anymore, and probably it also makes sure that it is generally is working.

PR #1780, opened 2023-12-22
Better document IterableObj.iter_items and improve some subclasses

### LINKED ISSUE #1775: Unclear why iter_items should be favored over list_items
In [`git.util.IterableObj`](https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1237-L1271), the [`list_items`](https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1248-L1261) method recommends that [`iter_items`](https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1263-L1271) be used instead, and begins to explain why, but it appears the actual explanation was never written:

https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1255-L1256

The deprecated [`git.util.Iterable`](https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1201-L1233) class has the same note in its [`list_items`](https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1209-L1224) docstring:

https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1218-L1219

In both cases, the subsequent non-blank line in the docstring documents what the `list_items` method itself returns, so the continuation truly is missing. The intended continuation does not seem to me to be something that can be inferred from the docstring as a whole. For example, here's the whole `IterableObj.list_items` docstring:

https://github.com/gitpython-developers/GitPython/blob/4023f28660e3a78ca9a83f03ca10b39780d32b0e/git/util.py#L1250-L1258

It may seem odd that, unlike #1712, I did not notice this when working on #1725. But I *think* the reason is not that the docstrings had made sense to me at that time, but instead that I had noticed the more minor (really, almost trivial) issue that perhaps the first paragraph should be split so that only its first sentence would be its summary line, decided to return to it later to consider that further, and then forgot about it.

The text "Favor the iter_items method as it will" appears to have been present, and its continuation absent, for as long as the surrounding code has been in GitPython. It was introduced in f4fa1cb along with the `Iterable` class itself.

In general, it may sometimes be preferable to obtain an iterator rather than a list or other sequence because it may not be necessary to materialize a collection just to iterate over it, and because unnecessary materialization can sometimes increase space usage. On the other hand, materialization guards against mutation of the original collection during iteration. But these are *completely general* ideas, not informed by the `list_items` docstrings nor even by any consideration specific to GitPython.

My guess is that the docstring intended to say something more specific, or at least to identify which general benefit of `iter_items` serves to recommend it. So I don't think I could propose a specific improvement to that documentation without insight into what was originally intended.

### LINKED ISSUE #1779: Submodule.iter_items sometimes attaches second empty iterator to its StopIteration
*This is minor bug, and the main benefit of fixing it may be an improvement in code clarity rather than the behavioral difference. I think it makes sense to fix this in the same pull request that fixes #1775, and I've opened #1780 for that, but I think it's clearer to document it here than in the pull request.*

`Submodule.iter_items` is a generator function as intended, because `yield` appears in its scope:

https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/objects/submodule/base.py#L1445-L1446

Sometimes when one writes a function that returns a generator object, it is a regular function rather than a generator function, because the function returns a generator expression or because it calls another function that returns a generator object. That technique is sometimes used to fail fast on errors, since calling a generator function does not immediately run its code, but only returns a generator object, and the code runs when that generator object is iterated. However, that is not done here. That it is not appears intentional and reasonable, and even if not, changing it would probably break backwards compatibility.

However, it contains this code, at the top:

https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/objects/submodule/base.py#L1400-L1405

The intention there is to make it so that calling `iter_items` returns an empty iterator when either of those two exceptions is raised. It succeeds at doing that, but its success is unrelated to using `iter([])` as the operand of `return`. Returning from a generator function--like falling off the end, which is more common--causes there to be no more items to iterate through, so a `next` call in which this happens raises `StopIteration`.

The operand of `return` in a generator function is neither returned to the caller--a generator object is returned--nor returned when `next` is called on that generator object. Instead, it becomes *the `value` attribute* of the `StopIteration` exception raised on the `next` call to indicate there are no more values. This feature of generators is rarely used, and the common ways of using generators, such as a `for` loop, ignore it. (When `yield from` is used as an expression, it evaluates to that value, which allows generators to get status information from their subgenerators, in the rare case their subgenerators return a meaningful value. It can also be accessed directly by explicitly catching `StopIteration`, binding the exception object to a variable, and accessing its `value` attribute. That's about it.)

I suppose it might make sense for `Submodule.iter_items` to have its returned generator object signal, in some way, that it is empty due to `OSError` (`IOError` and `OSError` are the same in Python 3) or `BadName`. But I think that having it attach a second empty iterator object to the `StopIteration` exception is clearly not intended to do so.

The statement was originally a bare `return`. The `iter([])` operand was added in 82b131c (#1282). It came in along with numerous additions and improvements in type hinting, and it looks like it was mistakenly added with the idea it might be needed for make type-checking succeed. However, I don't believe the major type checkers were as stable back then as they are today, so perhaps it was instead added as a workaround for a bug in a static type checker.

Interestingly, this is conceptually related to #1755, and I would've fixed it there, had I been aware of it at that time. The logic in the body `Submodule.iter_items` usually implicitly "returns" `None` by falling off the end (and `None` is bound as the `StopIteration` exception's `value` attribute), which is its intended behavior, but sometimes it inadvertently "returns" a `list_iterator` object instead.

EliahKagan:
Fixes #1775
Fixes #1779

This includes improvements to docstrings, and also other clarity improvements such as reordering methods, in both `git.util.IterableObj` and the deprecated `git.util.Iterable` class and its associated metaclass.

For #1775, it fills in the missing information. It also rewords the surrounding information, has the less recommended `list_items` methods' docstrings refer to the more recommended `iter_items` methods (instead of the other way around), and makes some further changes along those lines detailed in dfee31f.

I think that and the other changes in this pull request, documented in commit messages and the above-linked issues this closes, make sense to do together. For #1779, I added tests for some surrounding behavior that did not appear to be under test, as well as extending one of them to be a regression test for that minor bug, before fixing it.

I opened #1779 as its own issue rather than documenting here so that, if it turns out that both it and the bigger changes related to #1775 need to be discussed, the discussions are less likely to clash. (But I have no preference as to where such discussion should take place.)

---

There are some possible further changes to `git.util.IterableObj` and `git.util.Iterable` that I did not include here because I am not sure if they are wanted, but which I'd be pleased to open another pull request for (or to add to this pull request if preferred):

- I think it would make sense to put the class that is actually used in GitPython, and that other people should use, first. That is, I think `IterableObj` should be defined before `Iterable`. I can see why it's not done that way now: defining `Iterable` first allows its associated metaclass to be defined next to it, and also for `Iterable` and `IterableObj` to be defined next to each other, which may be beneficial since it may make it less likely for their similar docstrings to go out of sync.

- In `IterableObj` and `Iterable`, I think it would be okay for only the methods that are most recommended to give full details about their use, and others to refer to those methods. I made this change partially, moving information that was only in the `list_items` methods to the `iter_items` methods. But I propose going further and having the docstrings in the deprecated `Iterable` class refer to those in `IterableObj` for their detailed information, at least regarding the effect of omitting or passing extra arguments. I did not do this here because I don't know why the docstrings were near-duplicated in the first place, and perhaps there is a good reason for them to be that way. There would also be the question of whether these comments should be made different or one of them removed, given that the return types they're referring to are not the same:

  https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/util.py#L1227-L1228

  https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/util.py#L1265-L1266

If the second of those suggestions it to be done, then the first can be done too with little to no disadvantage, since it would no longer be necessary to keep their docstrings in sync.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -1401,7 +1401,7 @@ class Submodule(IndexObject, TraversableIterableObj):
             pc = repo.commit(parent_commit)  # Parent commit instance
             parser = cls._config_parser(repo, pc, read_only=True)
         except (IOError, BadName):
-            return iter([])
+            return
         # END handle empty iterator
 
         for sms in parser.sections():
--- a/git/remote.py
+++ b/git/remote.py
@@ -130,7 +130,7 @@ def to_progress_instance(
     return progress
 
 
-class PushInfo(IterableObj, object):
+class PushInfo(IterableObj):
     """
     Carries information about the result of a push operation of a single head::
 
@@ -300,7 +300,7 @@ class PushInfoList(IterableList[PushInfo]):
             raise self.error
 
 
-class FetchInfo(IterableObj, object):
+class FetchInfo(IterableObj):
     """
     Carries information about the results of a fetch operation of a single head::
 
--- a/git/util.py
+++ b/git/util.py
@@ -1183,7 +1183,8 @@ class IterableList(List[T_IterableObj]):
 
 
 class IterableClassWatcher(type):
-    """Metaclass that watches."""
+    """Metaclass that issues :class:`DeprecationWarning` when :class:`git.util.Iterable`
+    is subclassed."""
 
     def __init__(cls, name: str, bases: Tuple, clsdict: Dict) -> None:
         for base in bases:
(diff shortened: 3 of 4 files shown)


Byron (approved):
Thanks a lot for these documentation improvements, it's amazing to see how you keep chipping away from this huge task of modernising GitPython!

Regarding the suggestions, please do go ahead and apply them as they make perfect sense to me. Thanks again!

EliahKagan:
> Regarding the suggestions, please do go ahead and apply them as they make perfect sense to me. Thanks again!

Thanks, I've gone ahead and proposed them in #1785.

PR #1782, opened 2023-12-22
Pre-deprecate setting Git.USE_SHELL

### LINKED ISSUE #1781: `USE_SHELL = True` not needed since 2.0.8 but still suggested for Windows
The `Git.USE_SHELL` attribute is currently presented this way:

https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/cmd.py#L281-L285

I believe this encourages developers writing graphical Windows applications that use GitPython to set `USE_SHELL = True`, or to retain this in their code from when it may have been added in the past, at a time that it worked around a problem.

However, this is no longer necessary, even in that specific use case on Windows. **The problem of console windows being created for GitPython's `git` subprocesses in graphical Windows applications was solved back in 2016**, in 0d93908 (#469), by passing the `CREATE_NO_WINDOW` [process creation flag](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags). How and where that is expressed has changed over time, but the current code defines `PROC_CREATIONFLAGS` (which it passes to `creationflags` in calls to `Popen`):

https://github.com/gitpython-developers/GitPython/blob/d986a59467e586afb815d71845f29b605b756cdf/git/cmd.py#L231-L236

That change to passing `CREATE_NO_WINDOW` was not the main motivation behind #469, and it appears to some extent to have escaped notice, not having made it into the manually written changelogs of any version of GitPython. The recommendation to set `USE_SHELL` to `True` had previously been entered into the changelog in 1c2dd54 for 0.3.7, when it was made no longer the default on Windows, as it had been since 3f277ba (#126). In contrast, it was no longer necessary (nor useful) to do this since [2.0.8](https://github.com/gitpython-developers/GitPython/blob/main/doc/source/changes.rst#208---features-and-bugfixes), the first release after #469 was merged, but that change was never noted.

Using `shell=True` as a default is undesirable at least in the typical case that `Git.execute` is called indirectly through dynamic methods, because it is not typically feasible to account for the effect of shell expansions. The `Git` class is used in GitPython, and documented for use, in ways that do not account for characters with special meaning to the shell being present in text supplied as paths, branch and tag names, and so forth.

For that reason, combined with the better approach to #126 that came in with #469, comments or associated documentation for `Git.USE_SHELL` should no longer recommend its use in any likely scenarios, and should even caution against its use in the specific scenario where it was previously recommended.

---

I had suspected in #1685 (see "The `pythonw` use case") that calling `Popen` with `shell=True` might no longer be needed, but I did not realize that a change to GitPython, rather than Python or Windows, was responsible, and I was not sure. I now feel sure, due to a combination of the following factors:

- Official Microsoft documentation of the effect of the `CREATE_NO_WINDOW` [process creation flag](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags).
- Addition of a symbolic constant in Python 3.7 [documented accordingly](https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NO_WINDOW). Note that the technique worked before this, it was just necessary to define the constant or otherwise pass the correct value manually. These days, GitPython only supports Python 3.7 and higher, so the constant can be (and is) used.
- The explanation in 0d93908 (from [#469](https://github.com/gitpython-developers/GitPython/pull/469)) of the improvement of passing the flag.
- **Manual testing, done today.** I improved on the testing procedure I had used in [#1685](https://github.com/gitpython-developers/GitPython/issues/1685) by testing with `pythonw -m idlelib` on Python 3.7.9 and 3.12.1 virtual environments on Windows 10, both with `git.cmd.PROC_CREATIONFLAGS` unmodified, and separately by binding it to `subprocess.CREATE_NEW_PROCESS_GROUP` (i.e., it value without the `CREATE_NO_WINDOW` flag included). Removing the `CREATE_NO_WINDOW` flag, which I had not tried when working on [#1685](https://github.com/gitpython-developers/GitPython/issues/1685), reliably produced a console window that lasted the duration of a `git` command. I tested with the long-running command `git.Repo.clone_from("https://github.com/huggingface/transformers.git", "transformers")`.
- Verification that the specific example in the `USE_SHELL` comment, to the extent to which it may have intended other situations where `USE_SHELL = True` was needed, is out of date. That comment points to `test_untracked_files` as a case that requires `USE_SHELL = True` on Windows. CI reveals that test case passes with no special action and with the default of `False`, including on Windows, on all Python versions GitPython supports: [3.7](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886104405#step:12:477), [3.8](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886104650#step:12:477), [3.9](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886104817#step:12:477), [3.10](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886105001#step:12:477), [3.11](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886105114#step:12:477), [3.12](https://github.com/gitpython-developers/GitPython/actions/runs/7297204582/job/19886105203#step:12:477).

It may make sense to deprecate `Git.USE_SHELL`, or to deprecate setting it to `True`. Even before that is determined, however, I think it is worthwhile to change the comment. The main reason I'm opening this as an issue, rather than detailing it in a PR, is so that if the particular wording or other aspects of my proposed change (#1782) end up needing to be rejected or deferred, the issue itself won't be lost track of.

EliahKagan:
Fixes #1781

This does two things:

- Converts some comments on constants and other attributes to docstrings of the kind discussed in [#1734](https://github.com/gitpython-developers/GitPython/discussions/1734), throughout the codebase (not limited to the `git.cmd` module).
- Rewrites the comment, now in the form of a docstring, for `Git.USE_SHELL`, to discourage setting it to `True`, briefly explain why that can be unsafe, and explain the Windows console suppression use case for which it had previously been recommended and that it is no longer necessary or useful for that case ([#1781](https://github.com/gitpython-developers/GitPython/issues/1781)).

I have held off on calling it "deprecated," which I think is a decision that could be made later. There may be ways to limit the downsides of `shell=True` that arise with dynamic methods that call `Git.execute` (including those that happen indirectly through other GitPython code). If so, and a change that improves the safety of most existing client code that uses `shell=True` or `Git.USE_SHELL = True` would be a breaking change, then that would be a further reason to officially deprecate setting `Git.USE_SHELL` to `True`. If so, but it could be done in a way that is not a breaking change, then that might be a reason *not* to deprecate it. In any case, issuing a `DeprecationWarning` when the attribute is set on the class will probably require a metaclass or new custom descriptor to be written, which is more complicated than changing the message and may benefit from being reviewed separately.

The reason I've converted attribute/constant comments to non-reified docstrings and "pre-deprecated" `USE_SHELL` together in one pull request, rather than separately, is:

- Turning them into something that editors such as VS Code recognize and render as docstrings makes it so the revised "docstring" on `USE_SHELL` is more salient, increasing the benefit of the change.
- Making the changes separately, in the event that a release for an unrelated reason were to end up being done between them, would--depending on the order--either forgo some of the benefit, or risk the opposite effect of making the old and currently incorrect message more salient.

However, I would be pleased to sever this into two separate PRs or otherwise change this.

My commit message for 106bbe6 notes some of the above-mentioned considerations that may affect deprecation. If you'd prefer those thoughts not go in a commit message, I'd be pleased to amend it (or, if you prefer, I think GitHub will let you force push to my branch for the PR).

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -273,23 +273,30 @@ class Git(LazyMixin):
 
     # CONFIGURATION
 
-    git_exec_name = "git"  # Default that should work on Linux and Windows.
+    git_exec_name = "git"
+    """Default git command that should work on Linux, Windows, and other systems."""
 
-    # Enables debugging of GitPython's git commands.
     GIT_PYTHON_TRACE = os.environ.get("GIT_PYTHON_TRACE", False)
+    """Enables debugging of GitPython's git commands."""
 
-    # If True, a shell will be used when executing git commands.
-    # This should only be desirable on Windows, see https://github.com/gitpython-developers/GitPython/pull/126
-    # and check `git/test_repo.py:TestRepo.test_untracked_files()` TC for an example where it is required.
-    # Override this value using `Git.USE_SHELL = True`.
     USE_SHELL = False
+    """If True, a shell will be used when executing git commands.
 
-    # Provide the full path to the git executable. Otherwise it assumes git is in the path.
+    This should only be desirable on Windows, see https://github.com/gitpython-developers/GitPython/pull/126
+    and check `git/test_repo.py:TestRepo.test_untracked_files()` TC for an example where it is required.
+
+    Override this value using ``Git.USE_SHELL = True``.
+    """
+
     _git_exec_env_var = "GIT_PYTHON_GIT_EXECUTABLE"
     _refresh_env_var = "GIT_PYTHON_REFRESH"
+
     GIT_PYTHON_GIT_EXECUTABLE = None
-    # Note that the git executable is actually found during the refresh step in
-    # the top level __init__.
+    """Provide the full path to the git executable. Otherwise it assumes git is in the path.
+
+    Note that the git executable is actually found during the refresh step in
+    the top level ``__init__``.
+    """
 
     @classmethod
     def refresh(cls, path: Union[None, PathLike] = None) -> bool:
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -282,10 +282,20 @@ class Git(LazyMixin):
     USE_SHELL = False
     """If True, a shell will be used when executing git commands.
 
-    This should only be desirable on Windows, see https://github.com/gitpython-developers/GitPython/pull/126
-    and check `git/test_repo.py:TestRepo.test_untracked_files()` TC for an example where it is required.
+    This exists to avoid breaking old code that may access it, but it is no longer
+    needed and should rarely if ever be used. Prior to GitPython 2.0.8, it had a narrow
+    purpose in suppressing console windows in graphical Windows applications. In 2.0.8
+    and higher, it provides no benefit, as GitPython solves that problem more robustly
+    and safely by using the ``CREATE_NO_WINDOW`` process creation flag on Windows.
 
-    Override this value using ``Git.USE_SHELL = True``.
+    Code that uses ``USE_SHELL = True`` or that passes ``shell=True`` to any GitPython
+    functions should be updated to use the default value of ``False`` instead. ``True``
+    is unsafe unless the effect of shell expansions is fully considered and accounted
+    for, which is not possible under most circumstances.
+
+    See:
+    - https://github.com/gitpython-developers/GitPython/commit/0d9390866f9ce42870d3116094cd49e0019a970a
+    - https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags
     """
 
     _git_exec_env_var = "GIT_PYTHON_GIT_EXECUTABLE"
--- a/git/config.py
+++ b/git/config.py
@@ -65,12 +65,14 @@ log = logging.getLogger("git.config")
 log.addHandler(logging.NullHandler())
 
 
-# The configuration level of a configuration file.
 CONFIG_LEVELS: ConfigLevels_Tup = ("system", "user", "global", "repository")
+"""The configuration level of a configuration file."""
 
-# Section pattern to detect conditional includes.
-# https://git-scm.com/docs/git-config#_conditional_includes
 CONDITIONAL_INCLUDE_REGEXP = re.compile(r"(?<=includeIf )\"(gitdir|gitdir/i|onbranch):(.+)\"")
+"""Section pattern to detect conditional includes.
+
+See: https://git-scm.com/docs/git-config#_conditional_includes
+"""
 
 
 class MetaParserBuilder(abc.ABCMeta):  # noqa: B024
@@ -283,12 +285,14 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
     """
 
     # { Configuration
-    # The lock type determines the type of lock to use in new configuration readers.
-    # They must be compatible to the LockFile interface.
-    # A suitable alternative would be the BlockingLockFile
     t_lock = LockFile
-    re_comment = re.compile(r"^\s*[#;]")
+    """The lock type determines the type of lock to use in new configuration readers.
 
+    They must be compatible to the LockFile interface.
+    A suitable alternative would be the :class:`~git.util.BlockingLockFile`.
+    """
+
+    re_comment = re.compile(r"^\s*[#;]")
     # } END configuration
 
     optvalueonly_source = r"\s*(?P<option>[^:=\s][^:=]*)"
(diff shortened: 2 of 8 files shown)


Byron (approved):
Thanks so much for figuring this out and for making the necessary improvements to the USE_SHELL docstring.

As mentioned, using a metaclass for deprecation seems more complicated than it's worth, but I 'd think the calling it deprecated in the doc string would be sufficient.

That way, for when there is a version 4 of GitPython, some or all of the deprecated items can be removed.

EliahKagan:
> As mentioned, using a metaclass for deprecation seems more complicated than it's worth

I agree using a metaclass is not justified here. But I suspect that using a custom descriptor, which avoids interfering with other aspects of the class, would be justified, to make it more likely that developers writing code that sets the attribute will become aware of the deprecation. I say this especially in light of how `DeprecationWarning` is issued for most other deprecated features in GitPython and, even more so, how a docstring on an attribute with this syntax, while more discoverable than a comment because many editors/IDEs show it, is not reified. Because no `__doc__` attribute is added or extended with the text, Sphinx does not generate anything for it (I have just checked this, to be sure).

In any case, I will not add a custom descriptor either in this PR. If I ever propose that, I'll use a separate, subsequent pull request, ensuring it can be easily declined without forgoing any of the changes here.

> but I 'd think the calling it deprecated in the doc string would be sufficient.

Should I add a commit to this pull request for that change, rewording the docstring to explicitly say it is deprecated? Based on your comment, it sounds like you may be calling for that to be done here. However, based on the "approved" status taken together with the presence of an alternative explanation for not yet merging this PR -- the temporarily broken CI on main due to the bug #1784 fixed -- I'm not sure. I'd be pleased to make that change here (and retitle the PR accordingly) if desired.

Byron:
Apologies, communication here was ambiguous as I definitely missed to hit the merge button *somehow*.

I also leave it to your judgement to decide in which way a deprecation notice should be added - to me its main purpose would be to help finding the deprecated item to be removed. Users who upgrade would then hopefully realise that the item is gone, maybe with help of exhaustive change-notes that can easily be searched.

EliahKagan:
I think another purpose of deprecation can be to inform developers of software using GitPython of the risks of setting `Git.USE_SHELL` to `True`. However, even if it were decided that `DeprecationWarning` *should* be raised, there is the question of whether it should be raised on any access to the attribute, or only setting the attribute (which I think would be my preference), or only setting the attribute to `True`. I don't mean we shouldn't deprecate all use of it, only that I'm not sure `DeprecationWarning` should be raised on the less dangerous uses.

Therefore, while the changes I have proposed in #1787 deprecate `USE_SHELL` by documenting it as such, they do not issue `DeprecationWarning`. If I propose a change to issue the warning, I can do so in a separate PR (not just from this one, which has been merged, but also separate from #1787).

Byron:
Yes, I agree that the PR is a good first step and that a deprecation warning *could* be emitted as a follow-up. Since the attribute is bound for removal, in theory all access would ideally trigger a warning to help codebases to be cleared of it. It really depends on the purpose of the warning, as I can also imagine only warning on mutation or when set to true to prevent incorrect usage, which an option to fully deprecate it later and before 4.0 lands. In a way, one could implement a deprecation-ladder if one was inclined, even though I think that this might capture resources that are better spent elsewhere.

EliahKagan:
On further reflection and in view of what you're saying, I think the best thing to do is to have a `DeprecationWarning` issued on all accesses to the attribute, but have two versions of the warning: one for when it is set to `True` with a little extra text saying why that is bad (or at least mentioning *that* it is bad), and the other for all other access scenarios.

I should acknowledge, though, that when I looked through the various places deprecation notices appeared when working on a0fa2bd (#1787), I found that it was more of an even split between issuing a warning and not issuing one than I had thought. As such, the expectation that a warning be issued may not be as high as I had anticipated, at least among developers who have been working with GitPython for a long time.

With that said, I think in most of the other situations where no warning is issued, one could and probably should be. I might look at them all together and open a PR proposing more such additions than just for `Git.USE_SHELL`.

Byron:
> With that said, I think in most of the other situations where no warning is issued, one could and probably should be. I might look at them all together and open a PR proposing more such additions than just for `Git.USE_SHELL`.

Thanks a lot! I like that change in stance as it also serves as a reminder that getting a 4.0 ready *can* be a long-term goal. To me, this would be the GitPython as it should have been. Modern, clean, and with fewer bugs around encodings.

PR #1785, opened 2023-12-23
Shorten Iterable docstrings and put IterableObj first

EliahKagan:
This shortens the `git.util.Iterable` docstrings, removing most of the text duplicated from `git.util.IterableObj` docstrings and instead referring the reader to the specific corresponding methods in `IterableObj`. A comment about return types that appeared in both places but really only documented `IterableObj` is likewise removed from `Iterable`.

This also reorders the classes, placing `IterableObj` before `Iterable` and its associated metaclass. This makes it easier to find, and may help remind users that `IterableObj` is the class they should use.

These changes build on those in dfee31f (#1780).

Like the current tip of the main branch, I expect this to fail the Cygwin workflow until #1784 is merged or the bug it documents is otherwise fixed. It may make sense to rebase this onto main after that to verify that the tests still pass on Cygwin.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -1182,68 +1182,6 @@ class IterableList(List[T_IterableObj]):
         list.__delitem__(self, delindex)
 
 
-class IterableClassWatcher(type):
-    """Metaclass that issues :class:`DeprecationWarning` when :class:`git.util.Iterable`
-    is subclassed."""
-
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, much better!

PR #1786, opened 2023-12-24
Fix incompletely revised Iterable/IterableObj docstrings

EliahKagan:
Three wrong references in docstrings to `list_items` methods had accidentally remained: one that was meant to be removed altogether, and two that were meant to be references to `iter_items`.

This completes those changes, so the docstrings make sense and references to further information are to places that have that information.

---

I don't know how I missed this, which consisted of incomplete changes both in dfee31f (#1780) and in 2b768c7 (#1785). I distinctly remember anticipating this sort of problem specifically, and deliberately checking for it, *twice*, before opening #1780, and then checking for it *twice again* before #1785.

However, examination of the diffs for those PRs reveals that the unintended/inaccurate references are present, and not a result of anything odd happening in the merges. Given that, taken together with how I was fairly tired yesterday, I think this was purely my error, with no tooling or other issues contributing.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -1204,8 +1204,6 @@ class IterableObj(Protocol):
         filtering. However, when the method is called with no additional positional or
         keyword arguments, subclasses are obliged to to yield all items.
 
-        For more information about the arguments, see list_items.
-
         :return: Iterator yielding Items
         """
         raise NotImplementedError("To be implemented by Subclass")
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks so much - the workflow required to catch such a subtle issue is astonishing!

And of course, seeing the changes, I must wonder what I have been doing during the review which could easily have caught this.

I think one major reason for this is my stance towards 'waving through' your PRs after skimming over the changes themselves (I always read all of the PR message, even though I typically miss commit messages). Another contributor might be my lax attitude towards non-code changes. This PR shall be a reminder for myself to be more diligent in future.

PR #1787, opened 2023-12-25
Deprecate Git.USE_SHELL

EliahKagan:
This builds on the changes in #1782, actually deprecating `Git.USE_SHELL` as discussed in comments there, though no `DeprecationWarning` is currently issued on access to it. The `USE_SHELL` docstring is updated to state the deprecation while retaining most of the wording from 106bbe6. The docstring for `Git.execute`, on the `shell` parameter, is also updated to discourage passing `shell=True`, but does not declare that deprecated.

This also revises some existing deprecation notices and updates their predicted removal/breakage version number, converts more comments in docstrings (especially, but not only, those related to existing deprecations), and makes other such changes, building on the changes from 94a85d1.

A few more details are given in the commit messages.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -1233,10 +1233,10 @@ class IterableClassWatcher(type):
         for base in bases:
             if type(base) is IterableClassWatcher:
                 warnings.warn(
-                    f"GitPython Iterable subclassed by {name}. "
-                    "Iterable is deprecated due to naming clash since v3.1.18"
-                    " and will be removed in 3.1.20, "
-                    "Use IterableObj instead \n",
+                    f"GitPython Iterable subclassed by {name}."
+                    " Iterable is deprecated due to naming clash since v3.1.18"
+                    " and will be removed in 4.0.0."
+                    " Use IterableObj instead.",
                     DeprecationWarning,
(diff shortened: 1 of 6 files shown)


git/util.py
```diff
@@ -1233,10 +1233,10 @@ def __init__(cls, name: str, bases: Tuple, clsdict: Dict) -> None:
         for base in bases:
             if type(base) is IterableClassWatcher:
                 warnings.warn(
-                    f"GitPython Iterable subclassed by {name}. "
-                    "Iterable is deprecated due to naming clash since v3.1.18"
-                    " and will be removed in 3.1.20, "
-                    "Use IterableObj instead \n",
+                    f"GitPython Iterable subclassed by {name}."
+                    " Iterable is deprecated due to naming clash since v3.1.18"
+                    " and will be removed in 4.0.0."
```
Byron, line 1238: That's a great catch, I definitely wasn't aware of this 'version-deadline'. Lifting it to an actual major version is what I would have expected here, too.

Byron (approved):
Thanks a lot for what feels like completing the `USE_SHELL` issue - it's much appreciated.

PR #1791, opened 2024-01-09
Override bash executable, defaulting to Git for Windows git bash over WSL bash

emanspeaks:
Provides a mechanism to set a path to bash to use on Windows for executing commit hook commands.  If none is set, the default behavior should be to prefer Git Bash from Git for Windows.

If a user is running in Windows, they probably are using Git for Windows, which includes Git Bash and should be associated with the configured Git command set in `refresh()`.  Search first for Git Bash relative to the configured Git command, and, if it exists, use it.  Otherwise, fail back to the hardcoded default of `"bash.exe"`.

The original "default" bash without a path assumes that bash.exe is on the PATH.  If a user chose not to make MSYS2 tools available on the PATH on Git for Windows install, another method must be used to find that bash if it should be preferred over one found on the PATH.  Newer versions of Windows include a bash.exe in System32 that is then found by default, but this bash.exe is intended for use with WSL.  If a user intended to invoke git in WSL, presumably this package would have been installed there and not registered as "Windows" and reduces the ambiguity.  In most cases where running the python in "native" Windows, then, it is undesirable to invoke the WSL Bash that is not associated with the Git for Windows install.
Furthermore, direct access to bash.exe in System32 is reportedly deprecated in lieu of commands that explicitly invoke WSL.  

This addresses issues where git hooks are intended to run assuming the "native" Windows environment as seen by git.exe rather than inside the git sandbox of WSL, which is likely configured independently of the Windows Git as well.  A noteworthy example are repos with Git LFS, where Git LFS may be installed in Windows but not in WSL, causing commit hooks for LFS to fail despite being installed alongside the configured Git command in this package.


git/cmd.py
```diff
@@ -11,12 +11,13 @@
...
 import signal
-from subprocess import Popen, PIPE, DEVNULL
+from subprocess import Popen, PIPE, DEVNULL, run, CalledProcessError
 import subprocess
 import threading
 from textwrap import dedent
+from pathlib import Path
 
-from git.compat import defenc, force_bytes, safe_decode
+from git.compat import defenc, force_bytes, safe_decode, is_win
```
Byron, line 20: I think `is_win` [is deprecated](https://github.com/gitpython-developers/GitPython/blob/32c02d1c016b99dce9edc8895749fa936db85fab/git/compat.py#L31).

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+        # This addresses issues where git hooks are intended to run assuming
+        # the "native" Windows environment as seen by git.exe rather than
+        # inside the git sandbox of WSL, which is likely configured
+        # independently of the Windows Git.  A noteworthy example are repos
+        # with Git LFS, where Git LFS may be installed in Windows but not
+        # in WSL.
+        if not is_win:
+            return "bash"
+        try:
+            wheregit = run(["where", Git.GIT_PYTHON_GIT_EXECUTABLE], check=True, stdout=PIPE).stdout
```
Byron, line 343: I think there should be an assertion to assure `GIT_PYTHON_GIT_EXECUTABLE` is already set prior to using it.

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+
+    GIT_PYTHON_BASH_EXECUTABLE = None
+    """Provide the full path to the bash executable. Otherwise it assumes bash is in the path.
+
+    Note that the bash executable is actually found during the refresh step in
+    the top level ``__init__``.
+    """
+
+    @classmethod
+    def _get_default_bash_path(cls):
```
Byron, line 322: Isn't there a type annotation missing for the return type?

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+        except CalledProcessError:
+            return "bash.exe"
+        gitpath = Path(wheregit.decode(defenc).splitlines()[0])
+        gitroot = gitpath.parent.parent
+        gitbash = gitroot / "bin" / "bash.exe"
+        return str(gitbash) if gitbash.exists() else "bash.exe"
+
+    @classmethod
+    def refresh_bash(cls, path: Union[None, PathLike] = None) -> bool:
+        """This gets called by the refresh function (see the top level __init__)."""
```
Byron, line 353: A better doc-string would be one that describes what it does. Since it's public, it should be documented in detail.
I think it should also be documented which environment variable this function uses.

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+            wheregit = run(["where", Git.GIT_PYTHON_GIT_EXECUTABLE], check=True, stdout=PIPE).stdout
+        except CalledProcessError:
+            return "bash.exe"
+        gitpath = Path(wheregit.decode(defenc).splitlines()[0])
+        gitroot = gitpath.parent.parent
+        gitbash = gitroot / "bin" / "bash.exe"
+        return str(gitbash) if gitbash.exists() else "bash.exe"
+
+    @classmethod
+    def refresh_bash(cls, path: Union[None, PathLike] = None) -> bool:
```
Byron, line 352: `path` seems unused by `GitPython`, which means that it should be removed or documented as part of the public API.

It looks like this method is copied from `refresh()` with adaptations, so shortcomings are inherited. I don't think they should be allowed to proliferate though.

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+
+                # Determine what the user wants to happen during the initial refresh we
+                # expect GIT_PYTHON_REFRESH to either be unset or be one of the
+                # following values:
+                #
+                #   0|q|quiet|s|silence|n|none
+                #   1|w|warn|warning
+                #   2|r|raise|e|error
+
+                mode = os.environ.get(cls._refresh_env_var, "raise").lower()
```
Byron, line 404: I think the mode extraction itself should be factored into its own function and shared. 
It seems like the message itself can be a parameter, along with the environment variable itself.

Byron (changes requested):
Thanks a lot for tackling this and for explaining the reasoning so very well!

@EliahKagan already prepared me for this case and I hope he can take a look at this PR as well to see if it fits into the big picture.

EliahKagan:
> @EliahKagan already prepared me for this case and I hope he can take a look at this PR as well *[...]*

I will try to look at this in detail today or tomorrow.

EliahKagan (changes requested):
Thank you for bearing with my delay in reviewing this, and I apologize for any inconvenience. I strongly agree with the idea here, and I think this pull request will be suitable for merging, after some changes.

As you say, running hook scripts in WSL is not a good default, and it often fails or causes incorrect behavior. I think the problems with the way GitPython currently does it are severe enough to justify changing the default in a bugfix release. In my opinion, using WSL automatically even when a Git Bash shell is available is a bug, is usually not what users intend, and should be changed even though doing so may break some existing code that relies on the current behavior. (In #1745 and #1777 I had expressed hope that this could somehow be fixed in a way that does not break backward compatibility at all, for anyone, but I realize that is not realistic.) The ability to restore the old behavior by setting `GIT_PYTHON_BASH_EXECUTABLE` to `bash.exe` should be clearly documented.

The case for using the Git for Windows `bash.exe` when available is actually even stronger than it seems. Unlike on other systems, path search on Windows [differs substantially](https://github.com/python/cpython/issues/91558#issuecomment-1100942950) depending on whether a shell is used. That is, running a program in `cmd.exe` resolves commands differently from a direct call to `CreateProcessW`. One effect is that `Popen` (and `subprocess.run`, etc.) on Windows without `shell=True` searches a few directories prior to those listed in the `PATH` environment variable. One such directory is `System32`, even if it is not in `PATH` or no matter what is listed before it. So even when a "native" `bash.exe` is in a `PATH` directory listed before `System32`, the WSL `bash.exe` in `System32` is usually what `Popen` runs. (I say "usually" as there are other subtleties.) Furthermore, the WSL `bash.exe` often exists in `System32` even if no user has intentionally set up WSL and even if no WSL distributions are installed that would let it succeed.

Using the Git for Windows `bash.exe` whenever it is available would solve those problems, as well as the ones you described, and the further problem that using the WSL `bash` can hurt performance as it may need to start a WSL system. This is also much closer to the behavior that users are likely to expect, because Git for Windows itself runs hooks with a `#!/bin/sh` or `#!/bin/bash` shebang (and some others) with its `bash.exe`.

I also agree with much of the strategy this takes. However, there are some changes that should be made, to avoid introducing new [untrusted search path](https://cwe.mitre.org/data/definitions/426.html) vulnerabilities, similar to CVE-2023-40590 or CVE-2024-22190, that result in arbitrary code execution; to cover more of the common ways Git for Windows may be installed and run, while also avoiding using a `bash.exe` associated with a separate `git` from the one GitPython will use; and to ensure no code that uses GitPython without using hooks can break as a result of changes made here. That is all covered among my comments below. There are also two considerations not covered in review comments:

**A fix for CVE-2024-22190 (#1792) was recently merged.** That was after this pull request was initially opened, so this will have to be updated to get that bugfix, and to help prevent that security bug from being inadvertently reintroduced. This is also needed to resolve conflicts, which fortunately are very minor, and to get the changes from #1803 without which subsequent CI runs would fail (#1802). Although the [guidelines](https://github.com/gitpython-developers/GitPython/blob/main/CONTRIBUTING.md) don't express a preference on how PRs should be updated, based on https://github.com/gitpython-developers/GitPython/pull/1659#issuecomment-1756984419 and how minor the conflicts are, I recommend rebasing (maybe @Byron can speak to this more definitively). Updating this PR won't automatically fix *new* untrusted search path vulnerabilities the current code in this PR would add, but I've left review comments on the specific affected code about how that can be done.

**The tests should be modified to ensure these changes work as expected.** I am unsure if any new tests are needed. The main thing I'd suggest doing is to remove all WSL-related `xfail` mark decorators from test functions in `test/test_index.py`, and remove the "Setup WSL" step from the `pythonpackage.yml` workflow. Currently, the WSL `bash.exe` is being used on CI, even with these changes. This is because, as written, the changes here make assumptions Git for Windows does not always satisfy, including as it is used on the Windows runners for GitHub Actions. As commented below, that can be fixed. I suggest removing marks and the "Setup WSL" step, so all hook tests fail on CI due to the WSL `bash.exe` having no distribution to run `bash` in (see 9717b8d), before fixing the bug in how the Git for Windows `bash.exe` is found, and verifying that this makes the tests pass.

git/cmd.py
EliahKagan: I've recommended a change here that would no longer use `GIT_PYTHON_GIT_EXECUTABLE` directly. But maybe that shouldn't affect whatever assertions will go here, since `_call_process` would still be using `GIT_PYTHON_GIT_EXECUTABLE` indirectly.

A related design question, and one I am inclined to think this PR gets right, is when and how often to attempt automatic refresh of the `bash` path. This does it at import time only, and allows code using GitPython to manually refresh. (This design decision here affects whether there's anything to assert here, which is why I'm bringing it up in this thread.)

The opposite approach--which I think is in principle ideal but likely should not be done--would be, when no `bash` path has been set by the user or caller, to do the check before *each* attempt to run a hook with `bash`. Then changes to the filesystem, and (more likely) changes to `Git.GIT_PYTHON_GIT_EXECUTABLE` by directly setting it without calling `refresh` or `refresh_bash`, would be reflected in which `git`'s associated `bash` is looked for, and maybe there would be no need to call `refresh_bash` at import time. But I think there are two problems with that, which make the approach in this PR better, at least as an initial implementation of the feature. First, deferring the check and doing it more often is more complicated, and may thus introduce unexpected bugs. Second, I worry there might be a performance impact of refreshing often, doubling the number of subprocess invocations made to run hooks, which in a minority of use cases might be done very often. 

None of this should be confused with the question of whether the additional `bash --version` check should be included and, if so, how a failure should be reported and how the reporting method should be influenced by environment variables. That is a separate issue, which I've posted other review comments about.

git/cmd.py
Byron: I agree, let's wait until after the review comments were addressed if this is still an issue.

emanspeaks:
Thanks for all the comments!  Been busy with the day job the last week or so, but planning to try to address these things this weekend so we can get another round of reviews going and get it merged.

git/cmd.py
emanspeaks: I think static type checkers since about Python 3.8 generally no longer require return type annotations if they can be trivially inferred, so by habit I left it off.  I will add one for consistency's sake with everything else here :-)

git/cmd.py
emanspeaks: I addressed this argument in my new docstring.  I think it should be kept so there's another route for a user to provide an explicit path to use in Python rather than the roundabout way of setting an environment variable.

git/cmd.py
```diff
@@ -305,6 +306,173 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
...
+        # Assumes that, if user is running in Windows, they probably are using
+        # Git for Windows, which includes Git BASH and should be associated
+        # with the configured Git command set in `refresh()`.  Regardless of
+        # if the Git command assumes it is installed in (root)/cmd/git.exe or
+        # (root)/bin/git.exe, the root is always up two levels from the git
+        # command.  Try going up to levels from the currently configured
+        # git command, then navigate to (root)/bin/bash.exe.  If this exists,
+        # prefer it over the WSL version in System32, direct access to which
+        # is reportedly deprecated.  Fail back to default "bash.exe" if
+        # the Git for Windows lookup doesn't work.
```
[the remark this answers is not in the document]
emanspeaks: I addressed it in the documentation of the first four on your list.  Hopefully that is explicit enough for folks in addition to capturing it in the release notes, presumably.

emanspeaks:
I believe I have addressed all the comments.  I removed the xfails from the unit tests, and had assertion errors in `test_hook_uses_shell_not_from_cwd()` due to the hook *actually* executing in `type(_win_bash_status) in {WinBashStatus.Absent, WinBashStatus.WslNoDistro}` block rather than raising the exception.  I'm admittedly not understanding what this test is trying to test, but it seems like the `else` block is the behavior we now want since we should generally have working hooks regardless of WSL status.  I opted to just comment out the former block and make every case follow the former else branch to get the test to pass.  If that is not the correct result for this test, though, please let me know/enlighten me on what this test is trying to do so I can fix the issue.  Otherwise, with your concurrence, this PR should hopefully be ready :-)

EliahKagan (changes requested):
Thanks for the revisions! I agree that this addresses everything from the first review, and I think this is very close to being ready to merge.

I think the the only tricky remaining issue is what to do in `test_uses_shell_not_from_cwd`. I believe that one of the reasonable approaches is to leave it as it is, possibly removing the commented-out code altogether, and possibly adding a comment about future directions. But I'm also aware of two other reasonable approaches. I've left a review comment there that covers all three, and that also tries to answer the question of what that test is testing. Further information and context on what that test is checking for can be found in the "When hook scripts are run" subsection of CVE-2024-22190.

Aside from that, I recommend deleting the code that this PR currently comments out. You may have been intending to do that anyway, but I've posted review comments where applicable for it. It should not be necessary to install a WSL distribution on CI anymore, due to the improvements you've made in this PR. Furthermore, at least outside of tests related to security or the ability to signal a failed hook run, I don't think there is any need to cover WSL in the tests anymore (thus none of the WSL-related xfail decorations should be needed). If it turns out some such code will be needed again, it can be found in the commit history and restored as needed, with whatever modifications end up being required for it.

I've also noticed and included review comments about a few other small things.

git/cmd.py
EliahKagan: This comment is only in case of interest; it does not unresolve the thread, and it can be ignored.

Although static type checkers have become very capable, which is helpful for things like editor/IDE integration, my understanding is that the *meaning* of a function with no type annotations has remained the same. That is, a function with [no type annotations](https://github.com/gitpython-developers/GitPython/pull/1755#issuecomment-1837419098) still has `Any` as the type of all its parameters and of its return type, and the function is "dynamically typed" (which in practice means that its implementation should not be checked for type errors unless a tool is explicitly told to do so, though some tools have different defaults).

This, or at least `mypy`'s interpretation, can be verified on the [mypy Playground](https://mypy-play.net/?mypy=latest&python=3.12) (which also verifies that the type of a `self` or `cls` argument need not be annotated so long as another annotation is present, i.e., the change that has been made here, though it is valuable for more than just stylistic consistency, is fully sufficient).

Untyped:

```python
class C:
    @classmethod
    def create(cls):
        reveal_type(cls)
        print("x" + 3)
        return cls()

reveal_type(C.create())
```

```text
main.py:4: note: Revealed type is "Any"
main.py:4: note: 'reveal_type' always outputs 'Any' in unchecked functions
main.py:8: note: Revealed type is "Any"
Success: no issues found in 1 source file
```

Typed:

```python
from typing import Self

class C:
    @classmethod
    def create(cls) -> Self:
        reveal_type(cls)
        print("x" + 3)
        return cls()

reveal_type(C.create())
```

```text
main.py:6: note: Revealed type is "Type[Self`0]"
main.py:7: error: Unsupported operand types for + ("str" and "int")  [operator]
main.py:10: note: Revealed type is "__main__.C"
```

EliahKagan:
The new conflict looks to be from #1819 and easily resolvable. #1819 bumped the version on the `setup-wsl` action that this PR comments out.

schaveyt:
Nudge

Byron:
I believe this PR is waiting for the author, there is a lot of pending review comments and conflicts by now.
Maybe at this point it's fair to say the PR is stale.

Byron:
Closing due to age and conflicts. Please feel free to restart this with a new PR.

PR #1792, opened 2024-01-10
Fix two remaining Windows untrusted search path cases

EliahKagan:
Although #1636 caused GitPython to avoid using an untrusted search path on Windows under the most typical circumstances, it turns out that some cases remained. I believe only two remain at this time. This pull request fixes those.

This vulnerability was reported privately, and a full public writeup is not available at this time--I want to get this PR out as coordinated/requested, without extra delays. However, some details are present in commit messages, and I'll try to update this in some way in the future, either to include more information or to link to a relevant advisory or other source.

*Edit:* The CI status on all commits appears to be as I expected and intended.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
   hooks:
   - id: shellcheck
     args: [--color]
-    exclude: ^git/ext/
+    exclude: ^test/fixtures/polyglot$|^git/ext/
 
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.4.0
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -29,8 +29,8 @@ from git.util import (
     cygpath,
     expand_path,
     is_cygwin_git,
+    patch_env,
     remove_password_if_present,
-    safer_popen,
     stream_copy,
 )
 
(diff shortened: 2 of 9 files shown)


Byron (approved):
Thanks so much for figuring this out (!!) and for providing not only the fix, but also the tests that exemplify the issue while assuring they are now fixed.

I will create a new release shortly.

EliahKagan:
I hope it's okay that I've made a small edit to the advisory to fix some confusing and slightly inaccurate wording that I had missed when first writing it. I'd be pleased to make any further changes if desired (or even undo this change if for some reason it is not considered beneficial).

I've applied the change already at https://github.com/gitpython-developers/GitPython/security/advisories/GHSA-2mqj-m65w-jghx for this repository. For the global advisory, I cannot apply the change directly; I've submitted https://github.com/github/advisory-database/pull/3290.

I've noticed that these changes may be hard to spot quickly in the public web-based interface, so I've also made a gist whose [revision history shows](https://gist.github.com/EliahKagan/478e34530f88df77580f436526130a7e/revisions#diff-0ebb4bc00c0cb17af8a1e7dc079aed4cf1a07f35872d40ec0b84cdbedf183142) the proposed changes.

Byron:
Thanks a lot, and since the first time I wasn't able to suggest improvements, now that it is further improved, I am even more out of my ability to do so. I am sure, however, that the one reviewing the advisory PR will come to the same conclusion and merge the improvements.

Thanks again for your relentless pursuit of excellence!

EliahKagan:
The global advisory has been successfully updated.

PR #1798, opened 2024-01-11
Remove test dependency on sumtypes library

### LINKED ISSUE #1797: Testing depends on alpha `sumtypes`
Since commit 7ff3cee63c53520650db14bdf6e55551b8848567 `GitPython`'s tests started to depend on `sumtypes` which does not have any stable release available at PyPI and seems to be dead for more than two years now (see https://github.com/radix/sumtypes/commits/master/).

This causes inconvenience for downstream packagers where tests are run to make sure the constructed package is working properly.  Could you please make the `sumtypes` test dependency optional (especially on non-Windows OSes)?  Or switch off of it completely?

Thank you.

EliahKagan:
Fixes #1797

The `WinBashStatus` class in `test_index.py`, introduced in #1745, had used the `sumtypes` library to express the various cases relevant to xfail markings and troubleshooting and, where applicable, to express some associated debugging information. This includes the `Inapplicable` case, so `sumtypes` was used on all systems, not just Windows, which also usefully allowed the other cases to be inspected dynamically on any system, and allowed xfail markings to be written without excess complexity.

However, that only used a small fragment of `sumtypes` functionality, which can be expressed with only minor inelegance with explicit data classes. This pull request makes that change and removes `sumtypes` as a test dependency.

Because the `sumtypes` dependency was apparently causing problems for downstream maintainers (per #1797), I believe this change is justified. I believe the small inelegance of less compact definitions, and of the case classes not having an inheritance relationship to `WinBashStatus` itself, are the only disadvantages of this changed approach. The actual logic, stored information, and available debugging information is the same, fully useful `repr`s are still produced, no code change is needed in any of the markings that use `WinBashStatus`, and the class is still defined on all systems.

Nonetheless I have retested the cases locally (since only two of them occur on CI), verifying that they still work and that the same debugging information remains available.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -9,4 +9,3 @@ pytest-cov
 pytest-instafail
 pytest-mock
 pytest-sugar
-sumtypes
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -13,7 +13,7 @@ from pathlib import Path
 import re
 import shutil
 from stat import S_ISLNK, ST_MODE
-from subprocess import CompletedProcess, run
+import subprocess
 import tempfile
 
 import ddt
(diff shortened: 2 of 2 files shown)


Byron (approved):
Thanks a lot for your help!

Looking at the change, I'd even think it's preferable over the 'magic' that was introduced  as it's more explicit and on that merit probably more easily maintained.

Issue #1771, opened 2023-12-13
Status of Tree.cache and TreeModifier is unclear


EliahKagan:
In b295c13, `Tree.cache` was documented in `CHANGES.rst` as removed, based on the discovery in #369 that it could not serve its originally intended purpose. But `Tree.cache`, and its supporting code such as the `TreeModifier` class, were not removed. They remain part of the public interface, are not indicated as deprecated, and nothing in any docstrings counsels against their use.

I am unsure, based on this, if there is any reasonable use of them. My guess is that they may have been retained so that code that calls them *without relying on the effect* would not be broken, but I am not sure. If that is the case, it could be documented, `DeprecationWarning`s could be raised, and some of the non-public code may be possible to remove.

For example, even if `TreeModifier.set_done` needs to perform the sort, and do so with exactly the current effect, both the `git_cmp` and `merge_sort` functions could probably be eliminated entirely, since they were introduced to do something that was cumbersome (at least at the time, supporting Python 2 and 3) with the `list.sort` method. But this is no longer the case ever since 4f1110b. If I understand, this change made it no longer sort longer sequences first:

```diff
-    return len_b - len_a
+    return len_a - len_b
```

Byron:
Thanks for bringing this up. 

Since I don't have any knowledge anymore about this implementation except that it is most likely incorrect and should rather not be used for anything serious, what's left to discuss, I presume, is if code should be deprecated so it can eventually be removed.

Maybe it would be worth trying to find it if it is usable for something after all (are there no tests for this?), or one could pre-emptively deprecate it to see if an outcry follows.

In general, I'd always be for removing code, especially if its effectiveness is unclear, but I'd also not want to risk breakage in case we are missing something.

As always, I'd rely on your expertise and analysis, with a strong slant towards following your suggestions.

EliahKagan:
> Maybe it would be worth trying to find it if it is usable for something after all (are there no tests for this?)

Currently there are no tests, as they were removed in b295c13, which possibly was also intended to remove the code under test, I am not sure. Some of that test code could potentially be brought back in some capacity (or just experimented with on a feature branch to try and figure out what, if anything, `Tree.cache` and `TreeModifier` might be used for). I don't know that they were able to pass: they appear to have been expanded as part of a change made to verify that `Tree.cache` and `TreeModifier` *couldn't* achieve their once hoped-for goal.

[...]

Issue #369, opened 2015-12-08
Sorting in TreeModifier.set_done() seems to be inconsistent with git's own sorting -> breaks repository


ddanier:
git enforces tree objects to be sorted by name. GitPython enables users to sort their tree's by calling TreeModifier.set_done(), see:
https://github.com/gitpython-developers/GitPython/blob/master/git/objects/tree.py#L45

This sorting somehow seems to be inconsistent with git's own sorting mechanism, which means adding a tree might break the git repository. This means "git fsck" will throw errors and some remotes (gitlab for example) will no accept the commit. Anyways this bug leaves the git object store in an broken state which cannot be fixed easily, especially if there is much history after the broken commit/tree. I had to delete multiple branches from projects I manage because of their broken state so far.

About the problem itself:

GitPython uses core Python functions to sort the file list. This looks correct, but has one major difference to what git itself does. If you have two files beginning with the same chars (for example "file" and "file.second") their ordering will differ:

What git does:
1) "file.second"
2) "file"

What GitPython does:
1) "file"
2) "file.second"

We found this problem using fabdeploit for deployments using git, which allows us to add/removed files to/from the git repository. This way tree's will be modified which means sorting is essential.

Please fix this. ;-)

Byron:
Thanks for the detailed and informative issue ! This is indeed a critical one, whose fix is on the way.

To the interested audience, this is how git compares tree entries:

```c
int name_compare(const char *name1, size_t len1, const char *name2, size_t len2)
{
	size_t min_len = (len1 < len2) ? len1 : len2;
	int cmp = memcmp(name1, name2, min_len);
	if (cmp)
		return cmp;
	if (len1 < len2)
		return -1;
	if (len1 > len2)
		return 1;
	return 0;
}
```

Now git-python does it similarly. It's worth noting that the most difficult was to make it work in Py3 as well as in Py2, which contributes to most of the added code.

Byron:
I am hoping to hear from you if it does indeed fix the issue on your end :).
Thanks again.

ddanier:
Whoa, hadn't thought this would be this complicated. Big thanks for the quick response!

I will test if the bug is fixed for our repositories, but I'm pretty certain we will have no more issues.

ddanier:
Sadly I still get errors. I took one tree which produces errors and used "git show" to see what is stored in object store. The attached files correspond to the four different ways I used to create the tree:

[...]

PR #1799, opened 2024-01-12
fix: treeNotSorted issue

et-repositories:
https://github.com/gitpython-developers/GitPython/issues/369

### DIFF
--- a/test/test_tree.py
+++ b/test/test_tree.py
@@ -8,7 +8,9 @@ from io import BytesIO
 from git.objects import Tree, Blob
 from test.lib import TestBase
 
+import os
 import os.path as osp
+import subprocess
 
 
 class TestTree(TestBase):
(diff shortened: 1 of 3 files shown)


Byron (changes requested):
Thanks a lot for providing a fix!

I wonder if there was a way to assure the ordering is correct by looking at what git would do.

I assume the `files_in_desired_order` is from a repository, could you show a `git cat-file <tree>`?  Alternatively, maybe there is a way to use `git` to build a (somewhat) similar tree, and have the test compare the outcome of GitPython with the outcome of git itself - it's much more effort though.

Finally, [here is a good tree for testing ordering](https://github.com/Byron/gitoxide/blob/ef8ccd9d16143d37155d063747c69cade80f162d/gix-object/tests/fixtures/make_trees.sh#L14), and you might be able create a test-case with this one as well.

Something I'd appreciate is to re-order the commits so that the test, whatever you come up with, first, and the fix later so it's clear how it fails first.

Thanks a lot for bearing with me, but I'd like to gain confidence that this is now indeed what git would do, and not just different but subtly wrong (I have been bitten by this more than once).

test/test_tree.py
```diff
@@ -40,6 +40,41 @@ def test_serializable(self):
...
+
+        def file_names_in_order():
+            return [t[1] for t in files_in_desired_order]
+
+        def names_in_mod_cache():
+            a = [t[2] for t in mod._cache]
+            here = file_names_in_order()
+            return [e for e in a if e in here]
+
+        assert names_in_mod_cache() == file_names_in_order(), "add() keeps order"
```
Byron, line 73: Could you make this `self.assertEqual()`? That way I'd expect better error message in case of failure, a failure which should show up in CI at least once with the unfixed version of the code.

EliahKagan:
Would this also fix #1771?

test/test_tree.py
EliahKagan: The `pytest` runner [rewrites the bytecode of `assert` statements](https://docs.pytest.org/en/7.1.x/how-to/assert.html) appearing in tests, including in `unittest` tests, to provide the kind of [rich introspection](https://docs.pytest.org/en/7.1.x/example/reportingdemo.html) for `assert lhs == rhs` as for `self.assertEqual(lhs, rhs)` (while still evaluating expressions only once). It is possible to direct `pytest` not to do this, but that is rarely done.

This rich introspection includes element-by-element diffs when `lhs` and `rhs` are lists, as in this test. It is also done even when a custom assertion message is specified, as here: the rich introspection is shown after the message. So I would not expect using `self.assertEqual` would give better information here than `assert`; they should give output of the same quality. Whether or not this change is to be made, *if* it makes a difference in how much information is shown on failure, then that seems like a separate bug that should be investigated.

Experiments seem to bear this out, and suggest this change may not be needed: [a failure with the original test using `assert`](https://github.com/EliahKagan/GitPython/actions/runs/7504730404/job/20432447621#step:12:536) provides as detailed list introspection as [a failure when the test uses `self.assertEqual`](https://github.com/EliahKagan/GitPython/actions/runs/7504776269/job/20432592653#step:12:536). (I've linked to the part of each output that shows the first report, since that's where the test code can be examined along with the results. But the rich introspection can also be seen, in both cases, in the summaries at the end of the runs.)

Of course, that is all with the `pytest` runner. `unittest.TestCase` assertion methods should still be used if `unittest` tests are going to be run with the `unittest` runner rather than the `pytest` runner, and a few of the modules are possible to run fully with the `unittest` runner, including the `test_tree.py` module at issue here. Overall, though, GitPython's test suite is not compatible with the `unittest` runner, because some tests are pure pytest tests, but even more so because various `pytest` features are used throughout the test suite, in `unittest` tests and their decorations. (One of the goals of `pytest` is to allow its features to be gradually introduced in tests written against other frameworks that the `pytest` runner supports.)

Currently the `test_tree.py` module makes little use of `unittest.TestCase` assertion methods, already using `assert` statements instead in most cases, including for all equality assertions. So if this assertion is to be changed from the `assert lhs == rhs, message` form to the `self.assertEqual(lhs, rhs, message)` form, then I think it may be best to change the other assertions throughout that test module as well.

Byron: Thanks for the clarification! Under these circumstances that change request is definitely void, after all, the current assert style is more readable and lightweight.

I think it would be interesting to one day unify the assert-style within the project, as it will then just be about consistency, not about features.

doc/requirements.txt
```diff
@@ -1,3 +1,3 @@
-sphinx==4.3.0
+sphinx==5.0.0
```
Byron, line 1: I think this should be out of scope for this PR, maybe you can elaborate?

et-repositories: Screenshot 2024-01-15 at 2 27 37 PM

I'm trying to fix this CI/CD error , but it seems useless (, anyway I'll remove the commits about this

et-repositories:
> Would this also fix #1771?

i think it's related

doc/requirements.txt
EliahKagan: I have proposed a fix in #1803 for the Sphinx bug discovered here, which as noted is not specific to or caused by any of the changes in this pull request (see also #1802 for details on it). If #1803 is merged to main, then this pull request could be updated by rebasing it onto main to eliminate the issue here too.

Byron (approved):
Thanks a lot for bearing with us for changes and the turbulent CI.

EliahKagan:
> > Would this also fix #1771?
> 
> i think it's related

*If* this change has made the feature usable as originally intended and supersedes the advice in https://github.com/gitpython-developers/GitPython/issues/369#issuecomment-164519241, then I think think also fixes #1771.

If not, then I think #1771 should be modified accordingly, and any non-obvious limitations of the affected code documented in its docstrings (which would close #1771).

Byron:
I'd still lean towards rather removing it as it's deemed unfit, but that might be overly generalising and unnecessarily dismissive of the work of 'young' me. Since @et-repositories has had interest in fixing this, maybe they also have interest in investigating this matter?

Issue #1806, opened 2024-01-23
NullHandler added to all but git.repo.base logger


EliahKagan:
*This issue should not be confused with what the change in #1807 is trying to improve. The outcome of that PR is unrelated to this issue, because GitPython itself uses `logging.NullHandler` not `git.util.NullHandler`.*

Although the [recommended default](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) behavior for libraries is to allow messages of `WARNING` or higher severity to be displayed when the calling code does not configure logging, a library also has the option of causing its messages to be suppressed by default by registering an instance of `NullHandler` as a handler. (Logging is still happening; messages can still be subscribed to by the addition of other handlers.) GitPython takes this approach, though in a bit of a different way the Python documentation suggests for it. In GitPython, `NullHandler`s are registered for the individual modules' loggers rather than for a library-wide "top-level" logger.

This is with one exception. `git.repo.base` does not register a `NullHandler`:

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (main $=)$ git grep -Fn NullHandler -- git/
git/cmd.py:85:log.addHandler(logging.NullHandler())
git/config.py:65:log.addHandler(logging.NullHandler())
git/objects/commit.py:56:log.addHandler(logging.NullHandler())
git/objects/submodule/base.py:60:log.addHandler(logging.NullHandler())
git/objects/submodule/root.py:26:log.addHandler(logging.NullHandler())
git/remote.py:69:log.addHandler(logging.NullHandler())
git/util.py:1297:class NullHandler(logging.Handler):
(.venv) ek@Glub:~/repos-wsl/GitPython (main $=)$ git grep -Fn logging.getLogger -- git/
git/cmd.py:84:log = logging.getLogger(__name__)
git/config.py:64:log = logging.getLogger("git.config")
git/objects/commit.py:55:log = logging.getLogger("git.objects.commit")
git/objects/submodule/base.py:59:log = logging.getLogger("git.objects.submodule.base")
git/objects/submodule/root.py:25:log = logging.getLogger("git.objects.submodule.root")
git/remote.py:68:log = logging.getLogger("git.remote")
git/repo/base.py:92:log = logging.getLogger(__name__)
git/util.py:107:log = logging.getLogger(__name__)
```

It is unclear if this is intentional. It is easy to miss one (see #666) so it may well be an unintended omission.

I believe one of the following changes should be made, but I do not know which:

[...]

PR #979, opened 2020-02-07
Remove Python 2 compatibility shims

Harmon758:
[3.0.0 was supposed to have removed Python 2 support](https://github.com/gitpython-developers/GitPython/blob/master/doc/source/changes.rst#300---remove-python-2-support), but didn't actually remove any of the compatibility shims for Python 2.

In fact, with https://github.com/gitpython-developers/GitPython/commit/3b13c115994461fb6bafe5dd06490aae020568c1 (#860), further shims were added for Python 2 and going so far as to add future as a dependency and check for Python <= 2.3 (when [`sys.getfilesystemencoding`](https://docs.python.org/2/library/sys.html#sys.getfilesystemencoding) was added). I'm not sure why #860 was merged with those shims, especially since it was failing a test due to switching to Unicode strings in Python 2, but this commit that did so improperly merged requirements.txt as well, readding gitdb and ddt as dependencies when they had been removed with https://github.com/gitpython-developers/GitPython/commit/ce21f63f7acba9b82cea22790c773e539a39c158 (#826), so that setup.py could use requirements.txt. gitdb itself had actually already been removed as a dependency over 2 years before requirements.txt was updated, with https://github.com/gitpython-developers/GitPython/commit/2ca7c019a359c64a040e7f836d3b508d6a718e28 in v2.0.9.
The commit was then reverted with https://github.com/gitpython-developers/GitPython/commit/cc664f07535e3b3c1884d0b7f3cbcbadf9adce25, referencing failing Travis CI tests due to future having been added as a dependency in requirements.txt and the Travis CI configuration installing dependencies from test-requirements.txt, rather than requirements.txt and/or setup.py. The changes were then reintroduced with https://github.com/gitpython-developers/GitPython/commit/74a0507f4eb468b842d1f644f0e43196cda290a1, adding installation of requirements.txt (despite what the commit summary says) to the Travis CI configuration. It was then again reverted with https://github.com/gitpython-developers/GitPython/commit/c9dbaab311dbafcba0b68edb6ed89988b476f1dc, referencing the failing Travis CI build for the previous commit due to the same failing test for the PR.
[...]

PR #1807, opened 2024-01-23
Remove git.util.NullHandler

EliahKagan:
*This is not related to the issue of when a noop handler should actually be used, so it is independent of #1806.*

The `NullHandler` class in `git.util` was added when merging #300, to allow a noop handler to be used on Python 2.6, since the standard library `logging.NullHandler` class was added in Python 2.7.

When introduced in d1a9a23, the `git.util.NullHandler` class was also patched into the logging module, but that has no longer been done since 2fced2e (#979), nor does GitPython make other use of it.

`git/util.py` defines `__all__`, which does not list `NullHandler`. As `NullHandler` is not otherwise documented to be public, it is a bug for any code outside GitPython to use it, and removing it is not a breaking change. So there should be no need for a deprecation period or to wait until the next major version of GitPython.

For some reason I had not, at first, thought to check if it was in `__all__`, and assumed it was. The first commit here reflects that mistake, documenting it as deprecated and why the standard library `NullHandler` is preferable, and having it issue `DeprecationWarning` when constructed.

Although I wouldn't have written that if I'd been paying proper attention, I decided to keep it, since the documented difference between it and `logging.NullHandler` may be useful to see in the history. However, I'd be happy to rebase to remove that commit and update the other commit's message accordingly, if desired.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -1295,5 +1295,20 @@ class Iterable(metaclass=IterableClassWatcher):
 
 
 class NullHandler(logging.Handler):
-    def emit(self, record: object) -> None:
+    """Deprecated, use :class:`logging.NullHandler` instead.
+
+    This noop handler is like :class:`~logging.NullHandler` in the standard library,
+    which should be used instead, because it is now always available, and it overrides
+    more logging methods to make them noop. This class only overrides :meth:`emit`.
+    """
+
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for catching this! I'd happily keep all the commits that lead to the removal.

PR #1810, opened 2024-01-24
Clarify why GIT_PYTHON_GIT_EXECUTABLE may be set on failure

### LINKED ISSUE #1804: Explanation makes unclear reference to "first" and "second" imports
The `git.cmd.Git.refresh` class method includes this code:

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L478-L481

It is unclear what "the difference between a first import and a second import" is referring to. (I should have noticed this issue when working on #1725, in which I actually edited that comment to improve its formatting, capitalization, and punctuation, but I apparently did not notice it at that time.)

That code can run multiple times, accessing the same class-wide state each time, because `Git.refresh` can be called multiple times (and, being public, may reasonably be called by code outside of GitPython). **But this is not related to how many imports, or import attempts, occur.** The comment would make sense if Python's import system worked differently, but seems at odds with the actual design. As I understand it:

- **If an import succeeds**, the resulting module object is cached in `sys.modules`. Subsequent imports use the same module object and do not rerun the module's code.

  For example, suppose `hello.py` contains:

  ```python
  print("Hello, world!")
  ```

  Then running `import hello` from a REPL prints the greeting the first time it is run, but running it subsequent times produces no repeated side effect.

- **If an import fails**, the new module object, which may itself be incompletely populated, is removed from `sys.modules`, before control returns via the raised exception propagating up the call stack to the code that attempted the import. Even if that module object turns out still to exist, it will not be reused in subsequent attempts to import the code. A subsequent import attempt will rerun the module's code from scratch, creating a new module object which serves as a new namespace. The state in the old module object is not used; global variables of the module, including classes, and thus also including their attributes, are all new.

  For example, suppose `throw.py` contains:

  ```python
  import sys
  
  me = sys.modules[__name__]
  try:
      me.x += 1
  except AttributeError:
      me.x = 1
  
  try:
      a = sys.a
  except AttributeError:
      a = sys.a = []
  a.append(me)
  
  raise ValueError(x)
  ```

  Then running `import throw` from a REPL raises `ValueError` with a value of `1` each time it is run, and `sys.a` is populated with the abandoned module object from each attempt, all of which are separate objects. (Importing `sys` in the REPL allows it to be inspected, since `import sys` succeeds the first time it occurs, whether that is in the REPL or in `throw.py`, and thus all `import sys` get the same `sys` module object with the same `a` attribute.)

- **If the contents of `sys.modules` are explicitly modified** (rather than implicitly, due to the effect of `import` and `from` statements), one of the above two situations still applies, depending on the nature of the mutation. The exception to this would be if a reference to the old module object is obtained and patched back into `sys.modules`. That would be a very unusual thing to do, except perhaps for purposes of debugging or experimentation, and I'm reasonably sure the intent of that comment--and the code it documents--is not to cover such cases. Furthermore, that would still not rerun top-level code of the module.

- **Python's import system is extensible** and can be made to work differently from the usual rules. Of course, this cannot generally be anticipated by library code, and I likewise doubt this is a factor in the design or documentation.

I suspect that the code is correct and the comment incorrect, but since I'm not sure what was originally intended, I am not entirely sure how to fix it. It could be changed to say something like "discern the difference between the initial refresh at import time and subsequent calls," but this is very general and I worry it would paper over whatever really ought to be said. On the other hand, perhaps this really is all that was meant.

I haven't looked into the history of the Python import system but I don't think it operated in a fundamentally different way at the time this explanation was written. This code, and the original accompanying comment, was added in #640 (in 2b3e769). The reasoning given in that pull request, for the actual code change, continues to makes sense, and does not seem to have relied on any incorrect--nor any now-outdated--assumptions about the import system. As far as I can tell, it is only the comment that should be improved.

EliahKagan:
Fixes #1804

This clarifies the comment that explains the significance of setting the `Git.GIT_PYTHON_GIT_EXECUTABLE` attribute when running Git failed but the attribute wasn't set before.

This happens in the code path in `Git.refresh` where `has_git` is `False` and `old_git` is `None`, provided the refresh mode doesn't call for the initial refresh to raise an exception on failure.

It was previously described in terms of a "first" and "second" import, but as discussed in #1804, the rationale and effect were not altogether clear. This uses wording along the lines of what I suggested there, adjusted a bit for context.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -475,9 +475,10 @@ class Git(LazyMixin):
                     )
                     raise ImportError(err)
 
-                # We get here if this was the init refresh and the refresh mode was not
-                # error. Go ahead and set the GIT_PYTHON_GIT_EXECUTABLE such that we
-                # discern the difference between a first import and a second import.
+                # We get here if this was the initial refresh and the refresh mode was
+                # not error. Go ahead and set the GIT_PYTHON_GIT_EXECUTABLE such that we
+                # discern the the difference between the first refresh at import time
+                # and subsequent calls to refresh.
                 cls.GIT_PYTHON_GIT_EXECUTABLE = cls.git_exec_name
             else:
                 # After the first refresh (when GIT_PYTHON_GIT_EXECUTABLE is no longer
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -477,8 +477,8 @@ class Git(LazyMixin):
 
                 # We get here if this was the initial refresh and the refresh mode was
                 # not error. Go ahead and set the GIT_PYTHON_GIT_EXECUTABLE such that we
-                # discern the the difference between the first refresh at import time
-                # and subsequent calls to refresh.
+                # discern the difference between the first refresh at import time
+                # and subsequent calls to refresh().
                 cls.GIT_PYTHON_GIT_EXECUTABLE = cls.git_exec_name
             else:
                 # After the first refresh (when GIT_PYTHON_GIT_EXECUTABLE is no longer


git/cmd.py
```diff
@@ -475,9 +475,10 @@ def refresh(cls, path: Union[None, PathLike] = None) -> bool:
                     )
                     raise ImportError(err)
 
-                # We get here if this was the init refresh and the refresh mode was not
-                # error. Go ahead and set the GIT_PYTHON_GIT_EXECUTABLE such that we
-                # discern the difference between a first import and a second import.
+                # We get here if this was the initial refresh and the refresh mode was
+                # not error. Go ahead and set the GIT_PYTHON_GIT_EXECUTABLE such that we
+                # discern the the difference between the first refresh at import time
```
Byron, line 480:
```suggestion
                # discern the difference between the first refresh at import time
```

git/cmd.py
```diff
@@ -475,9 +475,10 @@ def refresh(cls, path: Union[None, PathLike] = None) -> bool:
...
+                # and subsequent calls to refresh.
```
Byron, line 481:
```suggestion
                # and subsequent calls to refresh().
```

Byron (approved):
Thanks a lot for not giving up on finding an explanation for the previous (and unclear) wording in order to find the right way to improve it :).

git/cmd.py
EliahKagan: I notice also that I wrote this in a way that it's unclear which `refresh` function is being referred to. I'm not sure it matters; it's `git.cmd.Git.refresh`, but calls to `git.refresh` also call it.

I should really have written "this refresh function". I'm not sure this matters and I don't know that it has to be changed further at this time, but you can let me know if it should.

Byron: That's a very valid point, I wasn't aware.

Since any `refresh()` function will eventually mean "this refresh function" due to one calling the other, it seems alright to keep the wording as is. If this is wrong, please do feel free to submit a quick PR for a fix.

EliahKagan: I don't consider it wrong. Next time I look through comments and docstrings to see if I can make any easy updates or other improvements--this is likely to happen when I add `ruff`, assuming I do manage to do that, if not to happen before then--I'll see if I still think wording like "this refresh function" makes it clearer.

PR #1812, opened 2024-01-25
Report actual attempted Git command when Git.refresh fails

### LINKED ISSUE #1809: Git.refresh GitCommandNotFound indicates "git" even for other commands
### The bug

Even when the Git command that `Git.refresh` runs and checks is not `git`, its command line is reported as `git` in the `GitCommandNotFound` exception, when such an exception is raised:

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L483-L485

### Steps to reproduce

The effect happens when the exception is raised due to a subsequent call to `Git.refresh`, since that is when `Git.refresh` raises `GitCommandNotFound` rather than `ImportError`. The `command` attribute is directly part of the traceback message, presented as `cmdline`. This can be produced by running a shell one-liner:

```text
$ GIT_PYTHON_GIT_EXECUTABLE=git-2.43 GIT_PYTHON_REFRESH=quiet python -c 'import git; git.refresh()'
  File "/home/ek/repos-wsl/GitPython/git/__init__.py", line 127, in refresh
    if not Git.refresh(path=path):
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ek/repos-wsl/GitPython/git/cmd.py", line 485, in refresh
    raise GitCommandNotFound("git", err)
git.exc.GitCommandNotFound: Cmd('git') not found due to: 'Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()
'
  cmdline: git
```

Although often a custom Git command may be a full path whose last component is exactly `git` (or sometimes `git.exe` on Windows), the above shows a realistic scenario in which even that may not be the case. I do not have a command named `git-2.43` installed, so it is correct that I get an error, but the error should not state the name of the command as `git`.

### Verification

Verbose debugging confirms that the command given in `GIT_PYTHON_GIT_EXECUTABLE` really is running, and that the only problem is that the hard-coded command name `git` is printed instead:

```text
$ GIT_PYTHON_GIT_EXECUTABLE=git-2.43 GIT_PYTHON_REFRESH=quiet GIT_PYTHON_TRACE=full python -c 'import logging; logging.basicConfig(level=logging.DEBUG); import git; git.refresh()'
DEBUG:git.cmd:Popen(['git-2.43', 'version'], cwd=/home/ek/repos-wsl/GitPython, stdin=None, shell=False, universal_newlines=False)
DEBUG:git.cmd:Popen(['git-2.43', 'version'], cwd=/home/ek/repos-wsl/GitPython, stdin=None, shell=False, universal_newlines=False)
  ...
  cmdline: git
```

Capturing the `subprocess.Popen` [audit event](https://docs.python.org/3/library/audit_events.html) verifies that no other logging bug in GitPython is confusing the analysis:

```text
$ GIT_PYTHON_GIT_EXECUTABLE=git-2.43 GIT_PYTHON_REFRESH=quiet python -c 'import sys, git; sys.addaudithook(lambda event, args: event == "subprocess.Popen" and print(args[:2])); git.refresh()'
('git-2.43', ['git-2.43', 'version'])
  ...
  cmdline: git
```

(For both the above runs, I replaced most of the traceback with `...`.)

### Why this is a bug

Since it's possible, however unlikely, that someone has come to rely on the exception object's `command` attribute having the exact value `"git"` in this situation, it should *arguably* only be changed if it really is a bug for it to hold that value.

I believe it is a bug, for three reasons:

#### 1. It is presented as the command line

When users see this in the `GitCommandNotFound` exception message, I believe they will assume it is the command that was run:

```text
  cmdline: git
```

#### 2. It differs from any other `GitCommandNotFound`

In other situations that raise `GitCommandNotFound`, it comes from `Git.execute`, and the exception is constructed with the actual command line as `command` (except that redactions are performed, such as for passwords):

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L1065-L1079

#### 3. The superclass documents the `command` argument

`GitCommandNotFound` inherits its `command` attribute from its `CommandError` superclass. Although `CommandError` does not document the `command` attribute directly, it does document the `command` argument from which the value of that attribute is derived:

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/exc.py#L83-L88

This does not really guarantee the specific meaning of the `command` *attribute*. (Likskov substitutability is usually a goal for objects that have already been constructed/initialized; subclass `__new__` and `__init__` are not expected to treat arguments the same ways as the superclass.) But it is a further reason to consider the current behavior in `Git.refresh` unexpected.

### Possible fixes

This could be fixed by having `Git.refresh` construct the `GitCommandNotFoundError` with the same command line that its `cls().version()` call causes `Git.execute` to use, which is given by `GIT_PYTHON_GIT_EXECUTABLE`.

Another approach, though, would be to keep a reference to the `GitCommandNotFound` exception that occurred occurred due to running `cls().version()`:

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L381-L385

Then that same exception object, with its original information, could be reraised instead of constructing and raising a new `GitCommandNotFound` object.

If this is to be done, then some refactoring may be in order, and a decision about whether that `GitCommandNotFound` object should be used as the `cause` of an `ImportError` in the case that is raised should probably also be made:

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L476

Thus, in addition to the other changes, that statement could be changed to the `raise ... from ...` form.

There may be reasons to avoid this approach, in whole or in part, that I am not aware of.

### LINKED ISSUE #1811: TestGit.test_refresh changes Git.GIT_PYTHON_GIT_EXECUTABLE for other tests
In `test_git.py`, `TestGit.test_refresh` tests the behavior of `git.refresh` when called with bad (nonexistent) and good (existing and usable) explicit `path` arguments.

- With the bad path, there could only be a lasting change to `Git.GIT_PYTHON_GIT_EXECUTABLE` or other global state due to a bug in `git.refresh`, or in `git.cmd.Git.refresh` or some other code `git.refresh` uses.
- **But with the good path, it remains set as `Git.GIT_PYTHON_GIT_EXECUTABLE`**, and tests that run after `test_refresh` use that instead of the original value that (when tests are run normally in an environment that supports them) is usually `"git"`.

https://github.com/gitpython-developers/GitPython/blob/307f1987596a99fb7b7192d8a5308a0df736ca7b/test/test_git.py#L309-L316

```text
>>> from git import Git
>>> from test.test_git import TestGit
>>> t = TestGit()
>>> Git.GIT_PYTHON_GIT_EXECUTABLE
'git'
>>> t.test_refresh()
>>> Git.GIT_PYTHON_GIT_EXECUTABLE
'C:\\Users\\ek\\scoop\\shims\\git.exe'
```

(That was run on Windows, but the effect is not specific to Windows.)

Although the good-path code can be simplified in the way it finds the absolute path it tests, that is not the cause of this.

**In the absence of bugs in the code under test**, this change to `Git.GIT_PYTHON_GIT_EXECUTABLE` between some tests and others should make no difference on most working test systems, because both the old and new paths should both work and typically run the same `git` installation. However, if there are bugs related to this attribute that the test suite should reveal or that otherwise interact with tests, then I think this may lead to inaccurate or less useful results, or complicate troubleshooting.

EliahKagan:
Fixes #1809
Fixes #1811

This modifies `Git.refresh` to report the attempted Git command rather than always `git` even if a different command was used. This affects the `command` attribute of the `GitCommandNotFound` exception, as well as (and automatically) the text displayed as `cmdline:` since that is from that attribute.

It also expands and adds more tests of `git.refesh` (which calls `Git.refresh`) to encompass this behavior, verifying that they fail before #1809 is fixed and pass afterwards, as well as verify some behavior that was already established but not yet under test.

Since #1811 would otherwise be exacerbated by these changes, I've also made it so those tests roll back the global changes they make that may affect other test by calling `git.refresh()` with no arguments. I believe this can be considered to fix that issue.

There is a bit more information in the commit messages, including about why I've taken that approach to #1811, at least so far, rather than specifically capturing and restoring the pieces of state that are currently affected.

This PR also includes some minor refactoring and comment revision for readability in some code called by `git.refresh` that should make no behavioral change. This includes an improvement to the wording of the comment in `Git.refresh` discussed in review comments on #1810; the opportunity for this came closer than I'd predicted in https://github.com/gitpython-developers/GitPython/pull/1810#discussion_r1464565019. (It might be possible to improve it further.)

*Edit: Corrected "some behavior that was not established" to "some behavior that was already established but not yet under test".*

### DIFF
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -310,8 +310,7 @@ class TestGit(TestBase):
         self.assertRaises(GitCommandNotFound, refresh, "yada")
 
     def test_refresh_good_git_path(self):
-        which_cmd = "where" if os.name == "nt" else "command -v"
-        path = os.popen("{0} git".format(which_cmd)).read().strip().split("\n")[0]
+        path = shutil.which("git")
         refresh(path)
 
     def test_options_are_passed_to_git(self):
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -306,16 +306,23 @@ class TestGit(TestBase):
         ):
             self.assertRaises(GitCommandNotFound, self.git.version)
 
-    def test_refresh_bad_git_path(self):
-        path = "yada"
-        escaped_abspath = re.escape(str(Path(path).absolute()))
-        expected_pattern = rf"\n[ \t]*cmdline: {escaped_abspath}\Z"
+    def test_refresh_bad_absolute_git_path(self):
+        absolute_path = str(Path("yada").absolute())
+        expected_pattern = rf"\n[ \t]*cmdline: {re.escape(absolute_path)}\Z"
         with self.assertRaisesRegex(GitCommandNotFound, expected_pattern):
-            refresh(path)
+            refresh(absolute_path)
 
-    def test_refresh_good_git_path(self):
-        path = shutil.which("git")
-        refresh(path)
+    def test_refresh_bad_relative_git_path(self):
+        relative_path = "yada"
+        absolute_path = str(Path(relative_path).absolute())
+        expected_pattern = rf"\n[ \t]*cmdline: {re.escape(absolute_path)}\Z"
+        with self.assertRaisesRegex(GitCommandNotFound, expected_pattern):
+            refresh(relative_path)
+
+    def test_refresh_good_absolute_git_path(self):
+        absolute_path = shutil.which("git")
+        refresh(absolute_path)
+        self.assertEqual(self.git.GIT_PYTHON_GIT_EXECUTABLE, absolute_path)
 
     def test_options_are_passed_to_git(self):
         # This works because any command after git --version is ignored.
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -45,12 +45,12 @@ def _patch_out_env(name):
 
 
 @contextlib.contextmanager
-def _restore_git_executable():
+def _rollback_refresh():
     old_git_executable = Git.GIT_PYTHON_GIT_EXECUTABLE
     try:
-        yield old_git_executable  # Let test code run that may rebind the attribute.
+        yield old_git_executable  # Let test code run that may mutate class state.
     finally:
-        Git.GIT_PYTHON_GIT_EXECUTABLE = old_git_executable
+        refresh()
 
 
 @ddt.ddt
@@ -319,7 +319,7 @@ class TestGit(TestBase):
         absolute_path = str(Path("yada").absolute())
         expected_pattern = rf"\n[ \t]*cmdline: {re.escape(absolute_path)}\Z"
 
-        with _restore_git_executable() as old_git_executable:
+        with _rollback_refresh() as old_git_executable:
             with self.assertRaisesRegex(GitCommandNotFound, expected_pattern):
                 refresh(absolute_path)
             self.assertEqual(self.git.GIT_PYTHON_GIT_EXECUTABLE, old_git_executable)
@@ -328,7 +328,7 @@ class TestGit(TestBase):
         absolute_path = str(Path("yada").absolute())
         expected_pattern = rf"\n[ \t]*cmdline: {re.escape(absolute_path)}\Z"
 
-        with _restore_git_executable() as old_git_executable:
+        with _rollback_refresh() as old_git_executable:
             with self.assertRaisesRegex(GitCommandNotFound, expected_pattern):
                 refresh("yada")
             self.assertEqual(self.git.GIT_PYTHON_GIT_EXECUTABLE, old_git_executable)
@@ -336,7 +336,7 @@ class TestGit(TestBase):
     def test_refresh_good_absolute_git_path(self):
         absolute_path = shutil.which("git")
 
-        with _restore_git_executable():
+        with _rollback_refresh():
             refresh(absolute_path)
             self.assertEqual(self.git.GIT_PYTHON_GIT_EXECUTABLE, absolute_path)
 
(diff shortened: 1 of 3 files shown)


test/test_git.py
```diff
@@ -306,14 +314,43 @@ def test_cmd_override(self):
...
-    def test_refresh(self):
-        # Test a bad git path refresh.
-        self.assertRaises(GitCommandNotFound, refresh, "yada")
-
-        # Test a good path refresh.
-        which_cmd = "where" if os.name == "nt" else "command -v"
-        path = os.popen("{0} git".format(which_cmd)).read().strip().split("\n")[0]
-        refresh(path)
+    def test_refresh_bad_absolute_git_path(self):
+        """Bad absolute path arg is reported and not set."""
```
Byron, line 318: It's a great idea to have an additional docstring!

Byron (approved):
Thanks a lot for tackling this, it's much appreciated!

*(On a personal note, your contributions are so numerous, I have to think about ways to express my gratitude in a non-repetitive fashion 😅 - it's a great problem to have though, please keep 'em coming)*

Issue #1808, opened 2024-01-24
Git.refresh warns to stdout and does not log


EliahKagan:
Warning messages are most often written to standard error. Furthermore, warnings from GitPython, when they are displayed, are in most cases displayed that way unless the caller arranges otherwise, because they are displayed by calling loggers' `warning` methods or, for `DeprecationWarning`s, by calling `warnings.warn`.

However, the warning `Git.refresh` displays when running a test command like `git --version` (with the command being tested used for `git`) fails, and `GIT_PYTHON_REFRESH` is set to `1`, `w`, `warn`, or `warning`, is different. It is printed to standard output. It also does not use any logging facility, so attempts to capture it by configuring Python logging will not succeed.

https://github.com/gitpython-developers/GitPython/blob/d28c20b1dbf18dc60a2b1859e818801bd8d742f2/git/cmd.py#L453-L454

If there is a reason for this behavior, then I think it should be commented or otherwise stated. Otherwise, I am not sure what to do. It may still not be feasible to change, because programs that use GitPython may be relying on the warning being printed to standard output, or on it not being printed to standard error. But if he behavior is going to change in the next major version, then that could be commented and/or this issue could be kept open until then. If the reason is unknown and it is undecided whether the behavior should be changed in the next major release, then a comment could be added just saying that it prints to standard output for backward compatibility.

Byron:
Thanks for bringing this up!

To my mind, libraries have no business with the stdout/stderr of a program, which makes this usage (and those like it) as bug. If other programs rely on the output of a library, they rely on an unstable API without any guarantees.

This is an opening for using the logging facility here instead.

Maybe in relation to #1806, if the logging would be configured to show warnings by default as it is suggested to well-behaved users of the logging module, then this important message should still show up even if GitPython is only used in an ad-hoc fashion. I am sure this was also the reason it prints to stdout, it was such a common issue that it should be assured the users can see it even without any prior configuration to the logging module.

EliahKagan:
> This is an opening for using the logging facility here instead.

I agree.

> Maybe in relation to #1806, if the logging would be configured to show warnings by default as it is suggested *[...]*

[...]

PR #1813, opened 2024-01-26
Don't suppress messages when logging is not configured

### LINKED ISSUE #1806: NullHandler added to all but git.repo.base logger

EliahKagan:
Fixes #1806

This stops adding `NullHandler` instances to GitPython's loggers. As noted in #1806, when they were added in #300 this prevented errors when GitPython logged messages and logging was not enabled, but since Python 3.2 there is a logger of last resort providing a nicer default behavior of showing the messages. (They are still shown with better formatting if logging is configured, even if just done with `logging.basicConfig()`, so applications should still typically configure logging.)

This also makes a few more minor related changes, detailed in the commit messages. This includes changing variable names from `log` to `_logger`, as mentioned in https://github.com/gitpython-developers/GitPython/issues/1806#issuecomment-1907690901, as well as consistently passing `__name__` to `logging.getLogger` instead of hard-coding names in a few places, as mentioned in https://github.com/gitpython-developers/GitPython/issues/1808#issuecomment-1911301659.

The major effect of the changes here is that messages of level `WARNING` or higher are not suppressed by default anymore, even when logging is not configured. So I could have included a change here to take care of #1808 as well, such as dc62096. But I plan to introduce that in a future pull request instead, because it would be best to add tests for it, which I think should build on the tests proposed in #1812. If both #1812 and this PR are merged, I can add tests followed by the changes from dc62096 that would make them pass. (If they are not merged, or they are merged with major changes, then that will also clarify how to move forward with making a warning from the initial refresh use the logging framework.)

This shows as resolving several CodeQL alerts and also adding new alerts that correspond exactly to them. This happens as a result of renaming `log` to `_logger`; CodeQL considers the alert resolved because there are no more calls to `log`, and issues new equivalent alerts for the same situations with `_logger`. The new alerts produce a failing *Code scanning results* check for the PR.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -81,8 +81,8 @@ execute_kwargs = {
     "strip_newline_in_stdout",
 }
 
-log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
+_logger = logging.getLogger(__name__)
+_logger.addHandler(logging.NullHandler())
 
 __all__ = ("Git",)
 
@@ -146,7 +146,7 @@ def handle_process_output(
                         handler(line)
 
         except Exception as ex:
-            log.error(f"Pumping {name!r} of cmd({remove_password_if_present(cmdline)}) failed due to: {ex!r}")
+            _logger.error(f"Pumping {name!r} of cmd({remove_password_if_present(cmdline)}) failed due to: {ex!r}")
             if "I/O operation on closed file" not in str(ex):
                 # Only reraise if the error was not due to the stream closing
                 raise CommandError([f"<{name}-pump>"] + remove_password_if_present(cmdline), ex) from ex
@@ -600,7 +600,7 @@ class Git(LazyMixin):
                     self.status = self._status_code_if_terminate or proc.poll()
                     return
             except OSError as ex:
-                log.info("Ignored error after process had died: %r", ex)
+                _logger.info("Ignored error after process had died: %r", ex)
 
             # It can be that nothing really exists anymore...
             if os is None or getattr(os, "kill", None) is None:
@@ -613,7 +613,7 @@ class Git(LazyMixin):
 
                 self.status = self._status_code_if_terminate or status
             except OSError as ex:
-                log.info("Ignored error after process had died: %r", ex)
+                _logger.info("Ignored error after process had died: %r", ex)
             # END exception handling
 
         def __del__(self) -> None:
@@ -654,7 +654,7 @@ class Git(LazyMixin):
 
             if status != 0:
                 errstr = read_all_from_possibly_closed_stream(p_stderr)
-                log.debug("AutoInterrupt wait stderr: %r" % (errstr,))
+                _logger.debug("AutoInterrupt wait stderr: %r" % (errstr,))
                 raise GitCommandError(remove_password_if_present(self.args), status, errstr)
             return status
 
@@ -1018,7 +1018,7 @@ class Git(LazyMixin):
         # Remove password for the command if present.
         redacted_command = remove_password_if_present(command)
         if self.GIT_PYTHON_TRACE and (self.GIT_PYTHON_TRACE != "full" or as_process):
-            log.info(" ".join(redacted_command))
+            _logger.info(" ".join(redacted_command))
 
         # Allow the user to have the command executed in their working dir.
         try:
@@ -1055,7 +1055,7 @@ class Git(LazyMixin):
         stdout_sink = PIPE if with_stdout else getattr(subprocess, "DEVNULL", None) or open(os.devnull, "wb")
         if shell is None:
             shell = self.USE_SHELL
-        log.debug(
+        _logger.debug(
             "Popen(%s, cwd=%s, stdin=%s, shell=%s, universal_newlines=%s)",
             redacted_command,
             cwd,
@@ -1167,7 +1167,7 @@ class Git(LazyMixin):
             # END as_text
 
             if stderr_value:
-                log.info(
+                _logger.info(
                     "%s -> %d; stdout: '%s'; stderr: '%s'",
                     cmdstr,
                     status,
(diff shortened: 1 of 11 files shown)


Byron (approved):
Thanks a lot for your help!

Once CI is green I will merge this PR and hope this can pave the way to improve the `git.refresh()` messaging as well :).

Byron:
It looks like pypi packages fail to refresh in the cygwin job for some reason, the last one timed out after 6h.

Merging anyway.

EliahKagan:
Sorry about leaving this in the draft state without explanation. I had hoped to check if the CodeQL results would cause a "Code scanning results" failure in the merge commit, as well as to improve on, or if not then to describe, the new Cygwin failure where `pip` blocks indefinitely while downloading some packages. But the problem in the Cygwin run appears separate from the changes in this PR. At some point in time between #1812 and this PR, something changed so that `pip` never finishes downloading `wheel`. If the workflow is edited to make it not download `wheel`, then it never finishes downloading `codecov`. Fortunately, my having left it this way [did not prevent](https://github.com/gitpython-developers/GitPython/pull/1813#event-11611071649) you from merging it.

It looks like the current situation on this repository's main branch is difficult to evaluate because many more CI checks on the pushed merge commit have been cancelled than are affected by either of these issues. Would you be willing to rerun the CI workflows on 10cdd03 *other* than the Cygwin one?

Byron:
Apologies for the premature merge, indeed I considered all failures unrelated and decided to just ignore them, particularly after having seen 6h timeouts for Cygwin. After restarting CI on the 10cdd03 , [Cygwin still hangs](https://github.com/gitpython-developers/GitPython/actions/runs/7671078305/job/20915661161). Does that mean that Cygwin based workflows don't work for anyone right now? That would be major breakage and I have no idea how it can stall forever.

EliahKagan:
Thanks for rerunning the checks on 10cdd03. I see everything other than the Cygwin job is okay. (I suspected this, since it is the situation in my fork since fast-forwarding, but I wasn't sure if there was more going on related to their cancellation.)

> Apologies for the premature merge, indeed I considered all failures unrelated and decided to just ignore them

The merge seems fine to me--no need to apologize. The CodeQL failure was due to the changes here, though the new alerts were replacing equivalent previous alerts. But I think you're mainly talking about the Cygwin failure, which is not from the changes here. It happens even when rerunning the Cygwin test job on a commit where it has successfully completed before.

> Does that mean that Cygwin based workflows don't work for anyone right now?

They're probably not broken for everyone, because the problem seems only to happen with `pip`, and only when attempting to download *some* packages from PyPI but not others. I am also unable to reproduce it locally. But it may be that such use is broken for everyone, or that Cygwin-based jobs are otherwise degraded for everyone. Or there could be a subtle bug in the `test-cygwin.yml` workflow that combines with an external change to produce the problem, as in #1784.

> That would be major breakage and I have no idea how it can stall forever.

Even though `pip` has a default socket timeout of 15 seconds as documented in the output `pip --help` for the `--timeout` option, I wonder if some network operation is still stalling, which seems to happen on Windows sometimes (#525, #530, 8364597, https://github.com/gitpython-developers/GitPython/issues/1676#issuecomment-1735052752). Even though it doesn't happen when the native Windows jobs run `pip`, I still wonder if that is involved. Strange low-level-ish network for downloading specific files is unusual but has been known to happen on CI (https://github.com/actions/runner-images/discussions/8791).

My guess is that it's something else, though. I'll look into it further.

EliahKagan:
I've learned enough about the Cygwin problem to find what I believe to be a reasonable workaround, which I've proposed in #1814 along with my findings.

PR #1816, opened 2024-02-06
Omit warning prefix in "Bad git executable" message

EliahKagan:
This builds on the changes in #1815 by omitting the `WARNING: ` prefix from the message itself that is produced when `GIT_PYTHON_REFRESH` is set to `warn` or a synonym (such as `w` or `log`) and the initial refresh performed at import time fails. This is the change suggested in https://github.com/gitpython-developers/GitPython/pull/1815#issuecomment-1925244493.

When logging is not configured, this changes what we see when the initial refresh fails:

```diff
-WARNING: Bad git executable.
+Bad git executable.
```

This may be considered slightly worse than before, but the advantage is that it doesn't print something that looks like a severity (or a redundancy) when logging *is* configured. For example, with common logging configuration such as by `logging.basicConfig()`, we get this improvement:

```diff
-CRITICAL:git.cmd:WARNING: Bad git executable.
+CRITICAL:git.cmd:Bad git executable.
```

This also makes some improvements to the tests. Besides updating them for the changed message (3250313), it:

- Test that the `GitCommandNotFound` exception raised when attempting a `git` operation, when the command in `Git.GIT_PYTHON_GIT_EXECUTABLE` does not exist but was set directly (rather than through the environment variable or an explicit call to one of the `refresh` functions), has an expected `command` attribute value (85ef145). This case has been under test for a long time, but the test had not checked that attribute.
- Reorder switch the order of that and an adjacent test so the tests of behavior that directly involves `GIT_PYTHON_GIT_EXECUTABLE` are all together (also in 85ef145).
- Use `Git` consistently in `test_git.py`, instead of using `Git` in some places and `type(self.git)` in others. See db36174 for the rationale.
- Change a call like `_logger.warning("%s", X)` to `_logger.warning(X)` in `test_index.py`, because GitPython uses the latter in all the other places where they would be equivalent. See 4b86993 for details. (This is relevant to this PR because 5faf621 adds another such call.)

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -450,7 +450,7 @@ class Git(LazyMixin):
                     )
 
                     if mode in warn:
-                        _logger.critical("WARNING: %s", err)
+                        _logger.critical(err)
                     else:
                         raise ImportError(err)
                 else:
--- a/test/test_git.py
+++ b/test/test_git.py
@@ -363,7 +363,7 @@ class TestGit(TestBase):
                     refresh()
                 self.assertEqual(len(ctx.records), 1)
                 message = ctx.records[0].getMessage()
-                self.assertRegex(message, r"\AWARNING: Bad git executable.\n")
+                self.assertRegex(message, r"\ABad git executable.\n")
                 self.assertEqual(self.git.GIT_PYTHON_GIT_EXECUTABLE, "git")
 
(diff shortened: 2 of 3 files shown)


Byron:
Thanks a lot for your continued support!

I particularly like how diffs are used to visualise the effective changes, they really help to make the changes pop.

PR #1832, opened 2024-02-19
Make sure diff always uses the default diff driver when `create_patch=True`

### LINKED ISSUE #1828: If `diff.external` is set, diffing via API fails silently as external tool isn't understood
GitPython version: 3.1.42.

```python
from git import Repo

repo = Repo(".")
print(repo.index.diff("HEAD"))
print(repo.index.diff("HEAD", create_patch=True))

print(repo.index.diff(None))
print(repo.index.diff(None, create_patch=True))
```

prints

```python
[<git.diff.Diff object at 0x10cbb08b0>]
[]
[<git.diff.Diff object at 0x103d49d30>]
[]
```

`R=True` workaround mentioned in #852 does not help either:

```python
print(repo.index.diff("HEAD", create_patch=True, R=True))
# []
```

This also happens when I try to diff tree against index or working tree

```python
print(repo.head.commit.diff(None, create_patch=True))
print(repo.head.commit.diff(None))

print(repo.head.commit.diff())
print(repo.head.commit.diff(create_patch=True))

```

returns

```python
[]
[<git.diff.Diff object at 0x103d69670>, <git.diff.Diff object at 0x103d699d0>]
[]
[<git.diff.Diff object at 0x103d69700>]
```

It looks like using `create_patch=True` when comparison includes index or working tree always returns empty list. So right now only way to reliably use `create_patch=True` is to diff tree against tree.

_Originally posted by @can-taslicukur in https://github.com/gitpython-developers/GitPython/issues/852#issuecomment-1951424734_

can-taslicukur:
Closes #1828

This change forces `diff` to always use the default diff driver when `create_patch=True`. So any user using external diff can reproduce the same result as users that don't.

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -155,7 +155,6 @@ class Diffable:
 
         if create_patch:
             args.append("-p")
-            # we expect the default diff driver
             args.append("--no-ext-diff")
         else:
             args.append("--raw")
--- a/git/diff.py
+++ b/git/diff.py
@@ -155,6 +155,8 @@ class Diffable:
 
         if create_patch:
             args.append("-p")
+            # we expect the default diff driver
+            args.append("--no-ext-diff")
         else:
             args.append("--raw")
             args.append("-z")
--- a/test/test_diff.py
+++ b/test/test_diff.py
@@ -473,3 +473,45 @@ class TestDiff(TestBase):
         self.assertEqual(True, diff.renamed_file)
         self.assertEqual("file_a.txt", diff.rename_from)
         self.assertEqual("file_b.txt", diff.rename_to)
+
+    @with_rw_directory
+    def test_diff_patch_with_external_engine(self, rw_dir):
+        repo = Repo.init(rw_dir)
+        gitignore = osp.join(rw_dir, ".gitignore")
+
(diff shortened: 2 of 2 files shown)


Byron (changes requested):
Thanks a lot! Much better than overriding the configuration like I suggested!

Would you mind digging into the test-suite and try to adjust one of them so that it will, indeed, set the `diff.external` value to some bogus program that makes the test fail unless the `--no-ext-diff` line is present?

It shouldn't be too hard and will assure this won't ever regress. This woudl also mean the comment can then be removed (as removing the statement will cause test failure).

Thanks for your consideration.

can-taslicukur:
Thank you so much for the review @Byron! 

I've included a new test, inspired by @EliahKagan 's [steps to reproduce the problem in the issue](https://github.com/gitpython-developers/GitPython/issues/1828#issuecomment-1951626373) and my own situation. These tests only pass if we set the `--no-ext-diff` flag when we create patch. PTAL!

Byron:
Thanks so much for your continued involvement and the test, it's great and much more thorough than I envisioned it!

PR #1838, opened 2024-02-23
Fix version_info cache invalidation, typing, parsing, and serialization

### LINKED ISSUE #1829: Refreshing doesn't invalidate cached version_info
The `version_info` property of `Git` instances is cached per-instance, while the `git.refresh` function and `Git.refresh` static method modify global state. When `version_info` has been read on a `Git` instance, refreshing affects the behavior of that instance, but does not invalidate the cache, causing the `version_info` property and dynamic `version` method to give inconsistent results even when no changes to environment variables or the filesystem have occurred.

Here's an example produced on a CentOS 7 system where I have both the system `git` and a newer upstream `git` installed:

```text
ek in 🌐 Eald in GitPython on  main via 🐍 v3.12.1 via 🅒 GitPython
❯ type -a git
git is /home/ek/bin/git
git is /usr/bin/git

ek in 🌐 Eald in GitPython on  main via 🐍 v3.12.1 via 🅒 GitPython
❯ git --version
git version 2.43.0

ek in 🌐 Eald in GitPython on  main via 🐍 v3.12.1 via 🅒 GitPython
❯ /usr/bin/git --version
git version 1.8.3.1

ek in 🌐 Eald in GitPython on  main via 🐍 v3.12.1 via 🅒 GitPython
❯ python
Python 3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:03:24) [GCC 12.3.0] on linux
>>> import git
>>> g1 = git.Git()
>>> g2 = git.Git()
>>> g1.version_info
(2, 43, 0)
>>> git.refresh("/usr/bin/git")
>>> g1.version()
'git version 1.8.3.1'
>>> g2.version()
'git version 1.8.3.1'
>>> g1.version_info
(2, 43, 0)
>>> g2.version_info
(1, 8, 3, 1)
```

Because I had accessed `g1.version_info` before refreshing, the stale version information is handed back even after the refresh.

This is admittedly consistent with how `version_info` is documented:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L839

But it seems to me that the documentation, as currently written, does not prevent it from being surprising. I think that either this should be made clear in the `version_info` property docstring, or the behavior should be changed so that refreshing invalidates all `Git` instances' cached `version_info` properties. I am not sure which is better, because I don't know why `version_info` is cached (and cached per instance).

This issue can probably be considered minor. In particular, this does not prevent `FetchInfo.refresh` (which `git.refresh` calls after calling `Git.refresh`) from updating `FetchInfo._flag_map` correctly, because `FetchInfo.refresh` accesses `version_info` on a newly created `Git` instance:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/remote.py#L346

### LINKED ISSUE #1830: Git.version_info type annotation is overly specific in length
The `version_info` property is declared to be a length-4 tuple:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L833-L841

But it often has fewer values, which is intentional:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L814-L826

So the type annotation should be changed, but I am unsure what it should be changed to, whether any additional documentation should be added, and whether having fewer than some number of numeric fields should be treated as an error and cause an exception to be raised.

If the type annotation should reflect that only a few specific lengths are reasonable, then it could be expressed as a `Union` of specific-length `Tuple` types. Otherwise it could be expressed as a variable-length tuple by writing the type as `Tuple[int, ...]`.

I recommend this be solved in such a way that the `cast` can be removed. Currently, the `cast` is working around ambiguity in how the return type is intended to be understood.

This is separate from #1829, though I noticed it while looking into that and they could possibly be fixed (or improved or otherwise closed) together.

### LINKED ISSUE #1833: Git.version_info can drop intermediate fields unexpectedly
The `version_info` attribute of `Git` instances drops non-numeric fields, even if they appear before a field that is kept:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L822

If a non-numeric field is followed by a numeric field, the effect is to make it look like the field that was actually non-numeric has the value of the subsequent field:

```text
(.venv) ek@Glub:~/repos-wsl/GitPython (main $=)$ cat >fake-git <<'EOF'
> #!/bin/sh
> printf 'git version 1.2a.3 (totally fake)\n'
> EOF
(.venv) ek@Glub:~/repos-wsl/GitPython (main $%=)$ chmod +x fake-git
(.venv) ek@Glub:~/repos-wsl/GitPython (main $%=)$ python
Python 3.12.1 (main, Dec 10 2023, 15:07:36) [GCC 11.4.0] on linux
>>> import git
>>> git.refresh('fake-git')
>>> git.Git().version_info
(1, 3)
```

(As shown, partially numeric fields that do not wholly parse to a number are non-numeric for this purpose.)

I think this should either be changed to stop before the first non-numeric field, or kept as it is but with this behavior documented explicitly to reduce confusion. In the latter case, the rationale should be given. The rationale may just be backward compatibility, though I am not sure if that is a sufficient reason to keep this behavior, since it seems unexpected in the cases where it makes a difference. There may be other alternatives that I haven't thought of.

I think this is a minor issue, since such versions may be rare or nonexistent. `git` version numbers that have a non-numeric field followed by a numeric field are common, such as 2.43.0.windows.1 on Windows. However, GitPython only looks at the first four fields, so the trailing 1 in 2.43.0.windows.1 is not wrongly presented as the field that follows the 0.

From the comments, it looks like this situation was specifically envisioned, suggesting that the exact current behavior may be intended:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L816

However, it seems more likely to me that it was done this way to preserve earlier logic while fixing a specific bug. The `if n.isdigit()` filter was added in 6a61110 (#195).

Full current code for context:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L814-L826

This may make sense to fix at the same time as #1829 and/or #1830, I'm not sure.

### LINKED ISSUE #1836: Unpickled Git object has broken version_info
A `Git` instance produced by pickle deserialization always has a `version_info` of `None`. This prevents version information from being accessed through `version_info`. It also causes the `version_info` property not to satisfy its type annotation as a tuple of `int`s (independently of #1830). This is the case whether or not `version_info` has been accessed on the original instance before it was pickled:

```python
>>> import pickle, git
>>> g = git.Git()
>>> g.version_info
(2, 34, 1)
>>> g2 = pickle.loads(pickle.dumps(g))
>>> g2.version_info
>>>
```

```python
>>> import pickle, git
>>> g = git.Git()
>>> g2 = pickle.loads(pickle.dumps(g))
>>> g2.version_info
>>>
```

This is due to an inconsistency in how the state of not yet having computed `version_info` is represented in the backing attribute `_version_info`:

- `Git` uses `LazyMixin` caching for `_version_info`, which treats the attribute as uncomputed when absent and computed when present.
- `Git` defines `__getstate__` and `__setstate__` for pickling and unpickling to represent state pickling does not preserve by setting attributes to `None`.

The `_version_info` attribute is set to `None` on a `Git` object created by unpickling, so when accessing the `version_info` property delegates to the `_version_info` attribute, its value of `None` is returned and the `if attr == "_version_info"` logic in the `_set_cache_` method (inherited from `LazyMixin` and overridden) is never run.

The intention is that an unpickled `Git` object have `version_info` in the uncomputed state. This works for the other two excluded attributes, which use the same `None` convention as in the `__getstate__` and `__setstate__`  implementations, and which do not use any `LazyMixin` functionality for their caching:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L315

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L319-L323

Unlike `_version_info`, which is managed by `_set_cache_`, the `cat_file_all` and `cat_file_header` attributes are initialized to `None` in `Git.__init__`:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L786-L788

This is natural to fix at the same time as #1829, because a fix for that should change the way caching works. `LazyMixin` caching supports attributes that are computed once and never become stale, so `version_info`/`_version_info` should implement caching in a different way to fix #1829, and that can fix this too if consistent with `__getstate__`/`__setstate__`.

EliahKagan:
Fixes #1829 - cache invalidation
Fixes #1830 - typing
Fixes #1833 - parsing
Fixes #1836 - serialization

Much of the design in the fixes for #1830, #1833, and #1836, as well as why I think it makes sense to include them here with #1829, is already largely covered in the discussion/comments in those three issues.

### Design goals for #1829

For #1829, which I would regard to be the biggest of the changes here, and which is the change for which the most additional code is added to the test suite, I made sure to preserve three important properties of the code:

1. Refreshing is global but affects all `Git` instances. Changing the way refreshing itself fundamentally works is beyond the scope of this PR and would also be a breaking change that I think could only be made in a new major release, and there are good reasons for refreshing to be global. Allowing a per-instance overriding refresh might be useful, but would be much more complicated than the current system unless accompanied by other design changes that would likely be breaking.
2. Caching is performed for `version_info` attributes. This has been explicitly promised in the `version_info` docstring for a long time, with the intention that accessing `version_info` be understood as cheap, and it is likely that substantial code exists that assumes it is cheap (or that throttles parts of GitPython that use it and benefit from it being cheap).
3. Caching does not take place across `Git` instances, and accessing `version_info` on one `Git` instance never has any effect on what happens when `version_info` is accessed on another `Git` instance. Since the actual validity of cached version information depends on when the global operation of refreshing was last done, it may seem like the caching should be made global as well. But this can lead to some new problems, such that I suggest it either be put off until later or never done.

#### Why not cache globally?

To expand on why the that third property is something I've taken care to preserve, these are the three problems I anticipate will arise if the caching were to be made global, in increasing order of significance:

- Although GitPython is not in general thread-safe ([#1052](https://github.com/gitpython-developers/GitPython/issues/1052)), caching globally may still make it less capable of concurrency than before, and the interaction with multiprocessing would also have to be considered.
- That the cache is per-instance is a long-standing behavior, which existing code may rely on. It is not explicitly stated in the property docstring, and I've taken care not to start stating it, in case it is to be changed. But I think it is how readers would have understood the docstring, since it is documenting an *instance* property and does not describe the cache as global or shared.
- Not sharing the cache ensures that the code `Git().version_info` always makes a subprocess call to get fresh information. Besides being one of the ways this behavior may have been relied on already, changing it would also, in practice, lead to code that uses GitPython being written to refresh more often, which would have the opposite effect on speed and resource usage than is intended by caching.

### The cache invalidation implementation

Rather than having the refresh operation reach out to all `Git` instances to invalidate their caches, I've added a `_refresh_token` class attribute that is set and reset along with `GIT_PYTHON_GIT_EXECUTABLE` during a refresh, and added a `_version_info_token` instance attribute (that, like `_version_info`, is excluded from pickle serialization).

When `version_info` is accessed, if `_version_info_token` is the same object as `_refresh_token`, then the cache is valid and the cached value in `_version_info` is returned. Otherwise, the cache is invalid, and `git version` is run to obtain the version; the version information is cached in `_version_info`, *`_version_info_token` is updated to `_refresh_token`*, and the newly computed value is returned.

The `_version_info_token` instance attribute starts out as `None` to simplify debugging and unpickling. Its value is always either `None` or a value that has at some point been the value of the `_refresh_token` class attribute. The `_refresh_token` class attribute is never `None`, but always a direct instance of `object`, even before the first refresh, which ensures it never wrongly matches the initial `None` value of a `_version_info_token` instance attribute.

A successful refresh always causes `_refresh_token` to differ from any `_version_info_token` that exists at that time. This is important--and, in particular, these tokens must not be tied to the value of `Git.GIT_PYTHON_GIT_EXECUTABLE`--because refreshing must properly update for a change to *which* git implementation is accessed by the same command or even via the same path. The most common and important such case is probably when `GIT_PYTHON_GIT_EXECUTABLE` is the default value of `"git"`, and git has been upgraded or a new version installed in a `$PATH` directory.

### Dropping the `LazyMixin` base class

An aspect of this PR that I recommend be double-checked in review is my assumption that `Git` having `LazyMixin` as a base class can be considered a mere implementation detail of `Git`, and thus that it is non-breaking to remove inheritance from that base class.

I removed it because it was only being used to implement caching for the `version_info` property. It is well-suited to some forms of caching elsewhere in GitPython, but not well-suited to the changes needed in `version_info` for #1829 and #1836 (as discussed in #1836). Because I stopped using `LazyMixin` for this, it was no longer used for anything. Removing it simplified the code, though if necessary it could be brought back and just not used for anything. (The new `version_info` implementation would still work with `LazyMixin` present but unused.)

#### Keeping the `super().__init__` call

`Git.__init__` contains this line:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/cmd.py#L778

This superficially appears related to the use of `LazyMixin`, but I believe that is not the case. `LazyMixin`, which is provided by the gitdb dependency, at least currently does not define an `__init__` method. Furthermore, such a `super()` call was present even before `LazyMixin` was brought in as a base class in e583a9e to provide caching for the `version_info` property (which was also introduced at that time). At that time, `Git` had no base class besides `object`, yet this was still present.

Calling `super().__init__()` helps with cooperative multiple inheritance in some cases. It is sometimes done even in the absence of a specific need, though sometimes it's not the correct logic, if the next class in the MRO has an `__init__` that is not callable with no arguments (or, rather, with only a `self` argument). Furthermore, you've [given feedback](https://github.com/gitpython-developers/GitPython/pull/1697#discussion_r1356428138) that I have understood to mean that having `Git` support inheritance is not a design priority.

So I don't think this `super().__init__()` is present based on a general preference for it. Instead, looking back to when it was first introduced, this was in the first commit that had the `Git` class, 039bfe3. At that time, `Git` did have a base class, [`MethodMissingMixin`](https://github.com/gitpython-developers/GitPython/blob/039bfe373c990fc6db3808d255abaff91235e82f/lib/git/method_missing.py#L1).

It may be that it should either be removed or commented with an explanation. But I think it is sufficiently unconnected to `LazyMixin` (and `version_info`) that no change should be made to it in connection with the changes in this PR.

#### Effect of directly accessing `_version_info` has changed

When `LazyMixin` was used, reading from the nonpublic `_version_info` attribute behaved the same as reading from `version_info`. Code outside the `Git` class should not do that, but there is one place in GitPython's own tests where it did:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/test/test_index.py#L531

That had to be changed to use `version_info` instead. If necessary, a different backing attribute could be used, and `_version_info` could be made an undocumented alternative to `version_info` that issues a warning but otherwise behaves the same. However, it seems to me that this is not necessary; `_version_info` was never advertised as part of the public interface of `Git`, as far as I can tell.

#### `GIT_PYTHON_GIT_EXECUTABLE` as an instance attribute is not supported

If `Git` instances could have their own `GIT_PYTHON_GIT_EXECUTABLE` attributes, then they would be used instead of the same-named class attribute that is written in a refresh. This cannot happen on a direct `Git` instance, because `Git` is slotted and has no such slot (and cannot, since the descriptor would conflict with the class attribute). But it is possible to make a subclass of `Git` where `GIT_PYTHON_GIT_EXECUTABLE` is an instance attribute. This has never been supported and should not be done, and will likely work poorly in other ways, but it is possible for caching to "support" it by detecting that this is the case and just never caching `version_info`.

I have neither done this nor commented about it, and even this mention of it may be gratuitous. However, in general the effect of attempting to set what is assumed to be a class attribute on an instance should be considered. My reasoning for not dealing with this at all is that, even if a reasonable use case arises to derive from `Git` and shadow base-class class attributes as instance attributes (which seems unlikely), the onus is then on the author of the derived class to override base-class members like `version_info` accordingly.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -833,28 +833,29 @@ class Git:
     @property
     def version_info(self) -> Tuple[int, int, int, int]:
         """
-        :return: tuple(int, int, int, int) tuple with integers representing the major, minor
-            and additional version numbers as parsed from git version.
+        :return: tuple(int, int, int, int) tuple with integers representing the major,
+            minor and additional version numbers as parsed from git version.
 
             This value is generated on demand and is cached.
         """
-        # Use a copy of this global state, in case of a concurrent refresh.
-        refresh_token = self._refresh_token
+        # Refreshing is global, but version_info caching is per-instance.
+        refresh_token = self._refresh_token  # Copy token in case of concurrent refresh.
 
-        # Ask git for its version if we haven't done so since the last refresh.
-        # (Refreshing is global, but version_info caching is per-instance.)
-        if self._version_info_token is not refresh_token:
-            # We only use the first 4 numbers, as everything else could be strings in fact (on Windows).
-            process_version = self._call_process("version")  # Should be as default *args and **kwargs used.
-            version_numbers = process_version.split(" ")[2]
+        # Use the cached version if obtained after the most recent refresh.
+        if self._version_info_token is refresh_token:
+            assert self._version_info is not None, "Bug: corrupted token-check state"
+            return self._version_info
 
-            self._version_info = cast(
-                Tuple[int, int, int, int],
-                tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
-            )
-            self._version_info_token = refresh_token
+        # We only use the first 4 numbers, as everything else could be strings in fact (on Windows).
+        process_version = self._call_process("version")  # Should be as default *args and **kwargs used.
+        version_numbers = process_version.split(" ")[2]
 
-        assert self._version_info is not None, "Bug: token check should never let None through"
+        self._version_info = cast(
+            Tuple[int, int, int, int],
+            tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
+        )
+        self._version_info_token = refresh_token
+
         return self._version_info
 
     @overload
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -25,7 +25,6 @@ from git.exc import (
     UnsafeProtocolError,
 )
 from git.util import (
-    LazyMixin,
     cygpath,
     expand_path,
     is_cygwin_git,
@@ -287,7 +286,7 @@ def dict_to_slots_and__excluded_are_none(self: object, d: Mapping[str, Any], exc
 ## -- End Utilities -- @}
 
 
-class Git(LazyMixin):
+class Git:
     """The Git class manages communication with the Git binary.
 
     It provides a convenient interface to calling the Git binary, such as in::
@@ -784,6 +783,7 @@ class Git(LazyMixin):
         self._environment: Dict[str, str] = {}
 
         # Cached command slots
+        self._version_info: Union[Tuple[int, int, int, int], None] = None
         self.cat_file_header: Union[None, TBD] = None
         self.cat_file_all: Union[None, TBD] = None
 
@@ -795,8 +795,8 @@ class Git(LazyMixin):
             Callable object that will execute call :meth:`_call_process` with
             your arguments.
         """
-        if name[0] == "_":
-            return LazyMixin.__getattr__(self, name)
+        if name.startswith("_"):
+            return super().__getattribute__(name)
         return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
 
     def set_persistent_git_options(self, **kwargs: Any) -> None:
@@ -811,20 +811,6 @@ class Git(LazyMixin):
 
         self._persistent_git_options = self.transform_kwargs(split_single_char_options=True, **kwargs)
 
-    def _set_cache_(self, attr: str) -> None:
-        if attr == "_version_info":
-            # We only use the first 4 numbers, as everything else could be strings in fact (on Windows).
-            process_version = self._call_process("version")  # Should be as default *args and **kwargs used.
-            version_numbers = process_version.split(" ")[2]
-
-            self._version_info = cast(
-                Tuple[int, int, int, int],
-                tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
-            )
-        else:
-            super()._set_cache_(attr)
-        # END handle version info
-
     @property
     def working_dir(self) -> Union[None, PathLike]:
         """:return: Git directory we are working on"""
@@ -838,6 +824,16 @@ class Git(LazyMixin):
 
             This value is generated on demand and is cached.
         """
+        if self._version_info is None:
+            # We only use the first 4 numbers, as everything else could be strings in fact (on Windows).
+            process_version = self._call_process("version")  # Should be as default *args and **kwargs used.
+            version_numbers = process_version.split(" ")[2]
+
+            self._version_info = cast(
+                Tuple[int, int, int, int],
+                tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
+            )
+
         return self._version_info
 
     @overload
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -794,7 +794,7 @@ class Git:
         self._environment: Dict[str, str] = {}
 
         # Cached version slots
-        self._version_info: Union[Tuple[int, int, int, int], None] = None
+        self._version_info: Union[Tuple[int, ...], None] = None
         self._version_info_token: object = None
 
         # Cached command slots
@@ -831,10 +831,10 @@ class Git:
         return self._working_dir
 
     @property
-    def version_info(self) -> Tuple[int, int, int, int]:
+    def version_info(self) -> Tuple[int, ...]:
         """
-        :return: tuple(int, int, int, int) tuple with integers representing the major,
-            minor and additional version numbers as parsed from git version.
+        :return:  tuple with integers representing the major, minor and additional
+            version numbers as parsed from git version. Up to four fields are used.
 
             This value is generated on demand and is cached.
         """
@@ -846,16 +846,14 @@ class Git:
             assert self._version_info is not None, "Bug: corrupted token-check state"
             return self._version_info
 
-        # We only use the first 4 numbers, as everything else could be strings in fact (on Windows).
-        process_version = self._call_process("version")  # Should be as default *args and **kwargs used.
-        version_numbers = process_version.split(" ")[2]
+        # Run "git version" and parse it.
+        process_version = self._call_process("version")
+        version_string = process_version.split(" ")[2]
+        version_fields = version_string.split(".")[:4]
+        self._version_info = tuple(int(n) for n in version_fields if n.isdigit())
 
-        self._version_info = cast(
-            Tuple[int, int, int, int],
-            tuple(int(n) for n in version_numbers.split(".")[:4] if n.isdigit()),
-        )
+        # This value will be considered valid until the next refresh.
         self._version_info_token = refresh_token
-
         return self._version_info
 
     @overload
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -306,12 +306,18 @@ class Git:
         "cat_file_all",
         "cat_file_header",
         "_version_info",
+        "_version_info_token",
         "_git_options",
         "_persistent_git_options",
         "_environment",
     )
 
-    _excluded_ = ("cat_file_all", "cat_file_header", "_version_info")
+    _excluded_ = (
+        "cat_file_all",
+        "cat_file_header",
+        "_version_info",
+        "_version_info_token",
+    )
 
     re_unsafe_protocol = re.compile(r"(.+)::.+")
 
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks so much for this fix, which was planned with incredible consideration.

The explanation done in the fantastic PR text is very helpful, and so are the tests. I read each one of them this time and found them very graspable. The sprinkled-in comments also helped.

Great work!

Issue #1794, opened 2024-01-10
Docs not deploying?


PeterJCLaw:
On PyPI this project has releases for 3.1.37 through 3.1.41, with the latter being the most recent at the time of writing. In contrast the docs only list up to 3.1.37 in the [changelog](https://gitpython.readthedocs.io/en/stable/changes.html).

The source for the [changelog](https://github.com/gitpython-developers/GitPython/blob/main/doc/source/changes.rst) on `main` does seem to contain entries for those missing releases, so maybe this is just that the docs aren't deployed or something?

Byron:
Thanks for letting me know! It looks like it stopped working 3 months ago, probably due to changes in readthedocs itself.

I was able to cobble a fix together which includes yet another yaml file. [It looks quite different too](https://gitpython.readthedocs.io/en/latest/intro.html) as the theme isn't supported anymore either, but I guess a wallpaper change was in order after more than 13 years or so 😁.

If you have any experience with theming (the new) sphinx, contributions for improvements are definitely welcome.

Please also note that I can't fix tags, but the next release should happen soon enough.

Issue #1842, opened 2024-02-23
Pull request preview builds for online docs


EliahKagan:
[Read the Docs](https://about.readthedocs.com/) supports [pull request builds](https://docs.readthedocs.io/en/stable/pull-requests.html), where pull requests trigger preview builds that are published to a different domain and marked with a warning banner to prevent confusion, and where the build status is reported as one of the CI checks for the PR.

I recommend enabling this for GitPython. Unlike some CI checks that are configured in GitHub Actions workflow files, my understanding is that Read the Docs pull request builds, including their status reporting as CI checks, do not use GitHub Actions. Instead, pull request builds are triggered via webhook (as are other Read the Docs builds), and the check status is reported to GitHub by authenticating with OAuth and posting it with (I think) the GitHub API.

For this reason, I don't believe I can propose the necessary changes by pull request. Instead, if you choose to enable pull request builds, [you can do so by following these official instructions](https://docs.readthedocs.io/en/stable/guides/pull-requests.html).

This should make it easier to verify the correctness of documentation changes in pull requests, as well as make it easier for contributors (including me) to become aware of, investigate, and attempt to improve problems in Read the Docs builds.

My immediate motivation for recommending this change is the hope that it would enable me to submit and test a fix for #1840. However, even if it turns out that the effect of changes to `.readthedocs.yaml` are not fully previewed in pull request builds--I think they are but I am not sure--it seems to me that this would be a worthwhile addition to the project's CI checks.

Currently the `pythonpackage.yml` workflow does test if a *local* documentation build is able to succeed, but this does not test if a remote (Read the Docs) build succeeds, and more importantly it does not produce a hosted preview of the built documentation that can be easily inspected.

Byron:
Thanks for researching this! I have enabled pull-request builds for GitPython and will be waiting for your confirmation that it actually works. The setup didn't involve more than ticking a checkbox.

EliahKagan:
Partial success:

- [#1843](https://github.com/gitpython-developers/GitPython/pull/1843) did trigger [an RTD pull request build](https://readthedocs.org/projects/gitpython/builds/23544223/), and the fix in that PR of installing GitPython and its documentation build dependencies [repopulated the API reference](https://gitpython--1843.org.readthedocs.build/en/1843/reference.html) (so the PR will fix [#1840](https://github.com/gitpython-developers/GitPython/issues/1840)) as well as enabling the theme to work again.
[...]

PR #1843, opened 2024-02-23
Build online docs (RTD) with -W and dependencies

### LINKED ISSUE #1840: API Reference not populated in online docs
In the [Read the Docs](http://gitpython.readthedocs.org/) deployment, [**the API Reference page**](https://gitpython.readthedocs.io/en/stable/reference.html#api-reference-toplevel) is unpopulated: the sections exist, but they do not contain any entries from the classes, functions, etc., that are supposed to go there.

This only affects the online documentation. Building the documentation locally still works with `make -C doc html`, provided the necessary dependencies are installed, and the API Reference page is fully populated. (This is how I have usually verified tricky or substantial changes to the documentation, and I was able to do it with no problems to check #1839.)

Furthermore, it only affects documentation from after #1794 was fixed. For a while before it was fixed, no new documentation was deployed online. But the API Reference page worked before that, and still works when accessed on readthedocs.org for those old versions.

It seems the changes made in 634151a to fix #1794 were sufficient to get documentation to build again, but not sufficient to set up Read the Docs in such a way that it would be able to populate members for code like:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/doc/source/reference.rst?plain=1#L128-L131

I don't know why this is or how best to investigate it, and I am not experienced with Read the Docs. But a couple of possibilities occur to me, both based in part on my observation that, when I build the documentation locally, the generated documentation shows values for attributes that are populated from environment variables, where the value I set for the environment variable is the value shown in the documentation. In other words, building the documentation relies on being able to import the modules and makes use of state set from running their top-level code.

If Read the Docs is itself (re)building the documentation--which if I understand correctly is what the new `.readthedocs.yaml` configures--then perhaps it is unable to successfully import those modules:

1. Does it have the necessary Python dependencies?
2. Does the container that builds the documentation have a `git` command?

I reiterate that Read the Docs deployment is not an area where I am currently knowledgeable.

EliahKagan:
Fixes #1840

This configures Read the Docs builds to be more like local builds in two ways:

- Pass `-W`, as is done in a local build with `make -C doc html`, so that if there are broken references, the build fails.
- Install dependencies. This [configures the Python environment](https://docs.readthedocs.io/en/stable/config-file/v2.html#python), via the `python.install` key, so that RTD builds install requirements.

More specifically on dependency installation, it does two things:

1. The equivalent of `pip install .`, which installs the project and its dependencies (though not any extras). This includes the gitdb dependency, which is needed to import GitPython's `git` module to populate sections in the API Reference page ([#1840](https://github.com/gitpython-developers/GitPython/issues/1840)).
2. The equivalent of `pip install -r doc/requirements.txt`, which installs the additional Sphinx-related dependencies used when building documentation locally.

Installing Sphinx-related dependencies is useful for three reasons:

- Least importantly, it should increase consistency between remote (RTD) and local documentation builds.
- It may be needed to avoid warnings that are not being fixed at this time, while still allowing the build to succeed with the `-W` option (see above on that change) that causes failure for immediately addressable problems. The effect of newer versions of Sphinx carrying a few extra hard-to-fix warnings for GitPython is noted in [#1802](https://github.com/gitpython-developers/GitPython/issues/1802) (and is why they are not upgraded in [#1803](https://github.com/gitpython-developers/GitPython/pull/1803)).
- One of the documentation build dependencies listed in `doc/requirements.txt` is `sphinx_rtd_theme`. In 634151a (for [#1794](https://github.com/gitpython-developers/GitPython/issues/1794)) the line specifying this theme was commented out, since it [apparently broke in the build](https://github.com/gitpython-developers/GitPython/issues/1794#issuecomment-1885644098). This may allow it to be used again (or can be replaced with another custom theme if desired).

This also reenables the `sphinx_rtd_theme` theme disabled in 634151a.

Finally, this makes minor changes to `.readthedocs.yml`'s comments and formatting so the comments are accurate for GitPython details and so the file is formatted in the same style as other YAML here.

**I did not test the Read the Docs configuration changes before opening this pull request**, so I suspect some changes may be needed. I am looking forward to seeing if pull request builds (#1842) are working and if they reflect changes made to `readthedocs.yml` such that these changes can be checked.

The only testing I did was to run `make -C doc html` locally to check that the changes to `conf.py` didn't break the documentation build and that it had the effect, locally, of restoring that theme. That worked as expected, but it is also the aspect of these changes *least* in need of testing.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,7 +4,7 @@
 # Required
 version: 2
 
-# Set the OS, Python version and other tools you might need
+# Set the OS, Python version and other tools you might need.
 build:
   os: ubuntu-22.04
   tools:
@@ -14,22 +14,24 @@ build:
     # rust: "1.70"
     # golang: "1.20"
 
-# Build documentation in the "docs/" directory with Sphinx
+# Build documentation in the "doc/" directory with Sphinx.
 sphinx:
   configuration: doc/source/conf.py
   # You can configure Sphinx to use a different builder, for instance use the dirhtml builder for simpler URLs
   # builder: "dirhtml"
   # Fail on all warnings to avoid broken references
-  # fail_on_warning: true
+  fail_on_warning: true
 
-# Optionally build your docs in additional formats such as PDF and ePub
+# Optionally build your docs in additional formats such as PDF and ePub.
 # formats:
-#   - pdf
-#   - epub
+# - pdf
+# - epub
 
 # Optional but recommended, declare the Python requirements required
-# to build your documentation
+# to build your documentation.
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
-# python:
-#   install:
-#     - requirements: docs/requirements.txt
+python:
+  install:
+  - method: pip
+    path: .
+  - requirements: doc/requirements.txt
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -93,7 +93,7 @@ pygments_style = "sphinx"
 # Options for HTML output
 # -----------------------
 
-# html_theme = "sphinx_rtd_theme"
+html_theme = "sphinx_rtd_theme"
 html_theme_options = {}
 
 # The name for this set of Sphinx documents.  If None, it defaults to


EliahKagan:
I've commented in #1842 about the RTD pull request build status.

Byron (approved):
Thanks so much for figuring this out!

It's good to see that the functional changes are very minor and clear in what they do, while producing the [prettiest GitPython docs yet](https://gitpython--1843.org.readthedocs.build/en/1843/reference.html#module-git.objects.commit).

PR #1846, opened 2024-02-25
`repo.blame` and `repo.blame_incremental` now accept `None` as the `rev` parameter.

### LINKED ISSUE #1835: `repo.blame` should allow passing `None` as the `rev` parameter.
This issue only affects the type hint given to the `rev` parameter, as passing `None` to the function does work and gives the expected behavior, which is running the `git blame -p -- <file>` command without a revision, which gives back the blame *including* changes to the local working directory that haven't been committed yet.

While the code works at runtime, language servers such as pylance give back an error, as `None` is not a valid type according to the type hints.

Gaubbe:
Fixes #1835.

As this pull request only affects the typings and does not change the behavior in any way, I did not write a test for the changes I made. 

While a test definitely *could* be written to see if the library does parse a blame output with uncommited changes, I don't perticularily think this is necessary, as the output format of the `git blame -p` command is the same as if there had been a commit, just with hard-coded values (such as the null SHA as the commit hash). Therefore, there is no reason the parsing would fail.

I did manually test both methods with the `None` rev parameter and they do give the expected behavior. 

Let me know if you want me to change anything else!

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -957,7 +957,7 @@ class Repo:
         # reveal_type(self.head.reference)  # => Reference
         return self.head.reference
 
-    def blame_incremental(self, rev: str | HEAD, file: str, **kwargs: Any) -> Iterator["BlameEntry"]:
+    def blame_incremental(self, rev: str | HEAD | None, file: str, **kwargs: Any) -> Iterator["BlameEntry"]:
         """Iterator for blame information for the given file at the given revision.
 
         Unlike :meth:`blame`, this does not return the actual file's contents, only a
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
Thanks a lot for your help!

After having had to look back to the issue it resolves to learn what `None` meant I realized that the docs for `rev` really need an overhaul. I think it would suffice to mention that `None` does in both instances.

Then, I think this will be a neat improvement!

Byron (approved):
Thanks a lot for your help with this! 
And the docs are great now 🙏.

PR #1850, opened 2024-02-29
Revise docstrings, comments, and a few messages

### LINKED ISSUE #1845: Unclear rationale for pipe usage in `output_stream` doc
The `Git.execute` method docstring includes this in the description of its `output_stream` parameter:

https://github.com/gitpython-developers/GitPython/blob/1e044ead62ea8ae31c879680a8645a0a301053ab/git/cmd.py#L990-L991

I think this is referring to the `subprocess` module, which could be clarified by changing `subprocess` to `` :mod:`subprocess` ``. But I'm not sure what issues this is referring to, or what information it seeks to impart to the reader.

To be clear, I'm not arguing that `output_stream` shouldn't use a pipe. Instead, I think this part of the docstring should be clarified (and I am unsure how to do so).

If this intends to be general, and vagueness is acceptable, yet it is considered better to have this "due to" clause than simply remove it, *and the issue really is in `subprocess` itself*, then it could be changed to:

```rst
due to limitations of :mod:`subprocess`
```

But I hope something more specific can be said.

The full context is:

https://github.com/gitpython-developers/GitPython/blob/1e044ead62ea8ae31c879680a8645a0a301053ab/git/cmd.py#L987-L994

### LINKED ISSUE #1847: Note in post_clear_cache docstring may be out of date
The `git.index.util.post_clear_cache` decorator has this note in its docstring:

https://github.com/gitpython-developers/GitPython/blob/fe1934c5c90e617d49a38a6c36b1eb17dea47013/git/index/util.py#L72-L74

This decorator is applied to the `merge_tree`, `move`, and `remove` methods of `git.index.IndexFile` (and nowhere else).

As such, that note in the decorator's docstring suggests an intended design direction for the `IndexFile` implementation in GitPython: that, if performance considerations permit, it should eventually make little to no use of `git` commands, unlike GitPython as a whole.

By now, my guess is that the intended direction is known, and my (somewhat less certain) guess is that this goal may have been abandoned, becuase:

- GitPython is considered to be in maintenance mode (per [the README](https://github.com/gitpython-developers/GitPython?tab=readme-ov-file#development-status)).
- [In a review comment](https://github.com/gitpython-developers/GitPython/pull/1770#discussion_r1425130807), if I understood correctly, you have expressed the hope that `IndexFile` is not heavily used, due to not having reached some of its original design goals.

*However*, there is also now a longer-term vision for GitPython, beyond maintaining the current feature set, articulated in https://github.com/Byron/gitoxide/discussions/1074 under "GitPython…".

If in the future the `IndexFile` implementation will be rewritten to use gitoxide's (future) Python bindings, then this docstring note is actually of *greater* relevance than before. But then it should still be updated to no longer characterize the aspiration to implement all of `IndexFile` without git subprocess calls as infeasible for performance reasons. And possibly even to mention gitoxide?

Because of that, I'm unsure if this note should be removed or updated (if, if updated, then how), or simply left as-is for the time being.

### LINKED ISSUE #1849: Note in GitCmdObjectDB should be removed or updated
Somewhat similar to #1845, the `git.db.GitCmdObjectDB` docstring contains this note:

https://github.com/gitpython-developers/GitPython/blob/87b8035c733018d6e8c097f79b61ad19a987f0d4/git/db.py#L34-L35

I think that note can simply be removed now, since as far as I know, it does not correspond to a current goal and is not needed to understand the code or how to use it. But I am not confident of this, because I am not sure exactly *what* it is actually referring to. If more features were implemented, would this have been:

- A change to `GitCmdObjectDB` itself?
- Switching back from `GitCmdObjectDB` to `GitDB`?
- A combination of the two, or something else related?

That only needs to be answered if it affects what should happen in that docstring (or elsewhere), which is why I've made this an issue rather than a discussion question.

I plan to remove it in a forthcoming PR with a number of other docstring revisions, but I'm opening this in case my understanding is mistaken.

EliahKagan:
Fixes #1845
Fixes #1847
Fixes #1849

This pull request is a sequel to #1725. Most changes are to docstrings. Notable changes:

- Applies revisions to recently discussed docstrings, including improving on the wording I added in [#1839](https://github.com/gitpython-developers/GitPython/pull/1839) and [#1844](https://github.com/gitpython-developers/GitPython/pull/1844), and making the changes discussed in [#1845](https://github.com/gitpython-developers/GitPython/issues/1845), [#1847](https://github.com/gitpython-developers/GitPython/issues/1847), and [#1849](https://github.com/gitpython-developers/GitPython/issues/1849).

- Revises docstrings throughout the `git` package along the lines of [#1725](https://github.com/gitpython-developers/GitPython/pull/1725), covering areas I outright missed, or where a further improvement or kind of improvement became apparent since then (either over time or due to recent rereading), or arose since then, or where I am better able to check the effect on rendered documentation, or where I am more comfortable making the change due to an improved understanding of the code being documented (or commented).

- Goes beyond that in making some new *kinds* of formatting changes to improve consistency that I had not included there, some that I had intentionally omitted at that time due to less experience with the project, and others that I only realized the benefit of more recently. See 1cd73ba for a list of many of these (that commit only covers a handful of files, but other commits made those kinds of changes to more files).

- Makes bigger formatting changes for readability, when reading the docstrings in the code, than in [#1725](https://github.com/gitpython-developers/GitPython/pull/1725). These also improve consistency, but their value is even more for readability:

  * Putting blank lines between all sections of all docstrings (in the `git` package).
  * Giving role-labeled sections a newline after the leading `:role:`/`:role args...:`.
  * Wrapping nearly all docstrings and most comments to 88 columns.
  * Formatting all occurrences of a class name that is intended to refer to that class with `:class:`, except in rare cases where there are disadvantages of doing so (e.g., the first line of a class docstring must not link to the class itself, or some readers may infer that it is a reference to another same-named class).

  The addition of blank lines and line breaks in docstrings is unneeded in some projects and often not done, but seems very useful to me *in GitPython*, where many docstrings have long sections (rather than, for example, sections like `:param foo: the predecessor of *bar*` that fit on a single line).

  The advantages of linking class names almost everywhere are that it:

  * makes clear *that* the class is what is being referred to (this is not always obvious, especially in GitPython due to subtleties of Git itself and, to a lesser extent, of GitPython's design), and
  * makes it easy to use the reference to get to information on something one is reading about.

- Codes all Sphinx references to classes, methods/functions, and attributes as either:

  * Relative to the current module, top-level function, class, or method.
  * Absolute.

  One effect is that there are far more absolute Sphinx references than before. This does not cause rendered Sphinx documentation to become more longwinded, because the references are written in such a way that the rendered text still omits the qualifiers (except on hover). This is mostly achieved by writing things like `` :class:`~a.b.c.D` ``, which is short for `` :class:`D <a.b.c.D>` ``, though occasionally I used the latter form: when the last two parts should be shown (`` :meth:`D.e <a.b.c.D.e>` ``), and in base classes' bulleted lists of documented subclasses where having the displayed name first lined it up better so as also to be readable unrendered in the code.

  There are important tradeoffs to consider here. **It is not at all obvious that this is the correct choice.** When I began to do it, it was not even with the intention that this would stay that way; I planned to make them all absolute, and thus obviously correct, and then thorough a combination of experimentation and consulting documentation about Sphinx and autodoc, make them relative again where correct, sufficiently reliable, and clear on reading. However, I noticed two things as I made many references absolute:

  * Clarity seems to be greatly improved by distinguishing things in the current module from other things that only *happen* to be in scope due to having been imported into it.
  * The entire module structure and logical organization of GitPython became clear when reading the docstrings in the code.

  Taken together with the advantage of avoiding confusing scoping issues (example: does Sphinx autodoc resolve references that are brought in scope due to `TYPE_CHECKING` imports?), I think it may actually be justified to leave it this way.

  **But it may not be justified, and this is an area that I want to call attention to for review.** The rendered documentation is not more longwinded as a result of doing this, but the *unrendered* documentation, i.e., the docstrings as written in the code, certainly are more longwinded in many cases. This potentially worsens readability.

  Overall readability seems improved to me, but there are two caveats to this:

  * I've been looking at the docstrings in this state for a while as I work on them, so what is readable to me may differ from what is readable in general.
  * I worry I may be giving with one hand and taking away with another: it may be that these absolute references are worsening readability, and by enough to make them unjustified in their current form, but that it is hard to know that this is so, because adding consistent spacing between sections (as described above) improves readability enough that it more than counterbalances it.

- Adds a small number of missing docstrings, and a moderately small number of previously omitted sections in docstrings. But this was not the focus.

- Slightly adjusts the wording of a few messages that GitPython outputs to signify errors. This is in cd8a312 (see [#1844](https://github.com/gitpython-developers/GitPython/pull/1844)), afd943a, and c67b2e2. These are the only behavioral changes. Besides adjusting code formatting slightly in small handful of cases, these are also the only code changes outside docstrings and comments.

There are some kinds of changes that are notably omitted from this pull request. These include:

- As in [#1725](https://github.com/gitpython-developers/GitPython/pull/1725), to limit the scope, I have not undertaken to update outdated documentation in `doc/source/` here, and nor applied any revisions to the project readme.
- Unlike [#1725](https://github.com/gitpython-developers/GitPython/pull/1725), I've kept the scope here a bit narrower: There are no modifications to type annotations, aspects of code style other than formatting (and very few of those), or changes of *any* kind in `doc/source/` (including no changes to its `conf.py`).
- This conspicuously excludes the `git.types` module from being revised. The reason is that, while docstrings should be revised there, it looks like it will make the most sense to do that together with changes to imports, which are beyond the intended scope of this pull request.
- This revises docstrings and (to a lesser extent) comments both in `git/` and `test/`, but the changes in `test/` are much less substantial, since there are fewer docstrings there, and tradeoffs there are different due to no part of the docstrings in the test suite being rendered by Sphinx. (This distinction is expanded on in 5cf5b60.)

### DIFF
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -127,7 +127,7 @@ def refresh(path: Optional[PathLike] = None) -> None:
         immediately, relative to the current directory.
 
     :note:
-        The *path* parameter is usually omitted and cannot be used to specify a custom
+        The `path` parameter is usually omitted and cannot be used to specify a custom
         command whose location is looked up in a path search on each call. See
         :meth:`Git.refresh <git.cmd.Git.refresh>` for details on how to achieve this.
 
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -129,12 +129,13 @@ def refresh(path: Optional[PathLike] = None) -> None:
     :note:
         The *path* parameter is usually omitted and cannot be used to specify a custom
         command whose location is looked up in a path search on each call. See
-        :meth:`Git.refresh` for details on how to achieve this.
+        :meth:`Git.refresh <git.cmd.Git.refresh>` for details on how to achieve this.
 
     :note:
-        This calls :meth:`Git.refresh` and sets other global configuration according to
-        the effect of doing so. As such, this function should usually be used instead of
-        using :meth:`Git.refresh` or :meth:`FetchInfo.refresh` directly.
+        This calls :meth:`Git.refresh <git.cmd.Git.refresh>` and sets other global
+        configuration according to the effect of doing so. As such, this function should
+        usually be used instead of using :meth:`Git.refresh <git.cmd.Git.refresh>` or
+        :meth:`FetchInfo.refresh <git.remote.FetchInfo.refresh>` directly.
 
     :note:
         This function is called automatically, with no arguments, at import time.
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -122,18 +122,22 @@ GIT_OK = None
 def refresh(path: Optional[PathLike] = None) -> None:
     """Convenience method for setting the git executable path.
 
-    :param path: Optional path to the Git executable. If not absolute, it is resolved
+    :param path:
+        Optional path to the Git executable. If not absolute, it is resolved
         immediately, relative to the current directory.
 
-    :note: The *path* parameter is usually omitted and cannot be used to specify a
-        custom command whose location is looked up in a path search on each call. See
+    :note:
+        The *path* parameter is usually omitted and cannot be used to specify a custom
+        command whose location is looked up in a path search on each call. See
         :meth:`Git.refresh` for details on how to achieve this.
 
-    :note: This calls :meth:`Git.refresh` and sets other global configuration according
-        to the effect of doing so. As such, this function should usually be used instead
-        of using :meth:`Git.refresh` or :meth:`FetchInfo.refresh` directly.
+    :note:
+        This calls :meth:`Git.refresh` and sets other global configuration according to
+        the effect of doing so. As such, this function should usually be used instead of
+        using :meth:`Git.refresh` or :meth:`FetchInfo.refresh` directly.
 
-    :note: This function is called automatically, with no arguments, at import time.
+    :note:
+        This function is called automatically, with no arguments, at import time.
     """
     global GIT_OK
     GIT_OK = False
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -106,7 +106,7 @@ def handle_process_output(
     decode_streams: bool = True,
     kill_after_timeout: Union[None, float] = None,
 ) -> None:
-    """Register for notifications to learn that process output is ready to read, and
+    R"""Register for notifications to learn that process output is ready to read, and
     dispatch lines to the respective line handlers.
 
     This function returns once the finalizer returns.
@@ -126,8 +126,11 @@ def handle_process_output(
     :param decode_streams:
         Assume stdout/stderr streams are binary and decode them before pushing their
         contents to handlers.
-        Set this to ``False`` if ``universal_newlines == True`` (then streams are in
-        text mode) or if decoding must happen later (i.e. for :class:`~git.diff.Diff`s).
+
+        This defaults to ``True``. Set it to ``False``:
+
+        - if ``universal_newlines == True``, as then streams are in text mode, or
+        - if decoding must happen later, such as for :class:`~git.diff.Diff`\s.
 
     :param kill_after_timeout:
         :class:`float` or ``None``, Default = ``None``
@@ -379,15 +382,14 @@ class Git:
 
     :note:
         The git executable is actually found during the refresh step in the top level
-        :mod:`__init__`. It can also be changed by explicitly calling
-        :func:`git.refresh`.
+        ``__init__``. It can also be changed by explicitly calling :func:`git.refresh`.
     """
 
     _refresh_token = object()  # Since None would match an initial _version_info_token.
 
     @classmethod
     def refresh(cls, path: Union[None, PathLike] = None) -> bool:
-        """This gets called by the refresh function (see the top level __init__).
+        """This gets called by the refresh function (see the top level ``__init__``).
 
         :param path:
             Optional path to the git executable. If not absolute, it is resolved
@@ -868,8 +870,7 @@ class Git:
         self.cat_file_all: Union[None, TBD] = None
 
     def __getattr__(self, name: str) -> Any:
-        """A convenience method as it allows to call the command as if it was
-        an object.
+        """A convenience method as it allows to call the command as if it was an object.
 
         :return:
             Callable object that will execute call :meth:`_call_process` with your
@@ -899,7 +900,7 @@ class Git:
     @property
     def version_info(self) -> Tuple[int, ...]:
         """
-        :return:  tuple with integers representing the major, minor and additional
+        :return: Tuple with integers representing the major, minor and additional
             version numbers as parsed from ``git version``. Up to four fields are used.
 
             This value is generated on demand and is cached.
@@ -1021,7 +1022,7 @@ class Git:
         :param output_stream:
             If set to a file-like object, data produced by the git command will be
             copied to the given stream instead of being returned as a string.
-            This feature only has any effect if `as_process` is False.
+            This feature only has any effect if `as_process` is ``False``.
 
         :param stdout_as_string:
             If ``False``, the command's standard output will be bytes. Otherwise, it
@@ -1030,10 +1031,10 @@ class Git:
 
         :param kill_after_timeout:
             Specifies a timeout in seconds for the git command, after which the process
-            should be killed. This will have no effect if `as_process` is set to True.
-            It is set to None by default and will let the process run until the timeout
-            is explicitly specified. Uses of this feature should be carefully
-            considered, due to the following limitations:
+            should be killed. This will have no effect if `as_process` is set to
+            ``True``. It is set to ``None`` by default and will let the process run
+            until the timeout is explicitly specified. Uses of this feature should be
+            carefully considered, due to the following limitations:
 
             1. This feature is not supported at all on Windows.
             2. Effectiveness may vary by operating system. ``ps --ppid`` is used to
@@ -1099,7 +1100,7 @@ class Git:
 
         :note:
            If you add additional keyword arguments to the signature of this method,
-           you must update the execute_kwargs tuple housed in this module.
+           you must update the ``execute_kwargs`` variable housed in this module.
         """
         # Remove password for the command if present.
         redacted_command = remove_password_if_present(command)
@@ -1420,9 +1421,10 @@ class Git:
         :param kwargs:
             Contains key-values for the following:
 
-            - The :meth:`execute()` kwds, as listed in :var:`execute_kwargs`.
+            - The :meth:`execute()` kwds, as listed in ``execute_kwargs``.
             - "Command options" to be converted by :meth:`transform_kwargs`.
-            - The ``insert_kwargs_after`` key which its value must match one of ``*args``.
+            - The ``insert_kwargs_after`` key which its value must match one of
+              ``*args``.
 
             It also contains any command options, to be appended after the matched arg.
 
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -479,7 +479,7 @@ class Git:
                         This initial message can be silenced or aggravated in the future by setting the
                         $%s environment variable. Use one of the following values:
                             - %s: for no message or exception
-                            - %s: for a warning message (logged at level CRITICAL, displayed by default)
+                            - %s: for a warning message (logging level CRITICAL, displayed by default)
                             - %s: for a raised exception
 
                         Example:
@@ -509,7 +509,7 @@ class Git:
 
                         Use only the following values:
                             - %s: for no message or exception
-                            - %s: for a warning message (logged at level CRITICAL, displayed by default)
+                            - %s: for a warning message (logging level CRITICAL, displayed by default)
                             - %s: for a raised exception
                         """
                         )
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -127,10 +127,10 @@ def handle_process_output(
         Assume stdout/stderr streams are binary and decode them before pushing their
         contents to handlers.
 
-        This defaults to ``True``. Set it to ``False``:
+        This defaults to ``True``. Set it to ``False`` if:
 
-        - if ``universal_newlines == True``, as then streams are in text mode, or
-        - if decoding must happen later, such as for :class:`~git.diff.Diff`\s.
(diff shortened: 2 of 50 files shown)


Byron:
Wow, what a massive amount of work!

That's over 300 queued CI jobs as well - I have never seen anything like it 😅. It's too bad that it's not possible to tell GitHub to build every  commit in the PR, so one is instead forced to push each commit individually (which has side-effects).

EliahKagan:
> That's over 300 queued CI jobs as well - I have never seen anything like it 😅.

Sorry--I should probably have checked with you first before pushing them that way, with this many commits!

EliahKagan:
It looks like CodeQL complains starting in [29c63ac](https://github.com/gitpython-developers/GitPython/pull/1850/commits/29c63ac8d7c75e4863cd355610da3cabd48a421c).

```diff
     @overload
-    def execute(self, command: Union[str, Sequence[Any]], *, as_process: Literal[True]) -> "AutoInterrupt":
+    def execute(
+        self,
+        command: Union[str, Sequence[Any]],
+        *,
+        as_process: Literal[True],
+    ) -> "AutoInterrupt":
         ...
```
That change, which reformats one of the `Git.execute` overloads, produces new warnings about the preexisting situations where CodeQL does not see `redacted_command`, `cmdstr` (derived from `redacted_command`), and the expression `safe_decode(stderr_value)` as safe.

Those uses are all down in the body of `Git.execute`, of course. I *presume* that, rather than having detected something new, CodeQL is redetecting what it has already detected. This actually makes sense: one of the `@overload` stubs that contains information about the arguments that can be passed to `Git.execute`, which can contain sensitive information and from which values are derived that CodeQL believes may not be sanitized, is now different code, so CodeQL is detecting the cause as a different (albeit equivalent) one.

This is to say that I *think* the CodeQL situation is analogous to that in #1813, even though it seems stranger here. If that is the case, then I think this is not cause for alarm. It seems unlikely it could be anything else, but I suggest verifying it by checking that corresponding warnings have also gone away, i.e., that CodeQL automatically dismisses equivalent warnings to each of the new ones.

It may be easier to examine once the CodeQL jobs have run for the last commit, and I expect automatic CodeQL comments to be posted as well, as in #1813.

Byron (approved):
This is incredible work! I don't even know how to express how incredible it is as I don't know how it's even humanly possible to read through this much code while doing all these small to bigger improvements, with most of them being related to something as abstract as whitespace. 
I cannot imagine how a workflow around these changes must look like unless there is a way to automatically build and preview exactly the part of the file one is currently working on.

As for my own review, I tried to use [the renderered docs of the PR](https://gitpython--1850.org.readthedocs.build/en/1850/reference.html?highlight=handle_process_output) sporadically to get an idea of how it looks like. Nice initially, it quickly turned out that plenty of functions I would have expected to be findable through search simply weren't, so a preview wasn't really possible after all.

I also admit that I skimmed most of the changes, which already take quite some amount of time.

Regarding the concern related to the format of class-links and the tradeoffs between the very usable rendered versions and the more wordy representation in text, I think favouring the rendering is the right choice (even in case it's not always shown, at least on readthedocs). It would also be my expectation that IDEs can use these docstring links to allow the user to jump to the linked classes directly, or maybe they even render the docs in-line while resolving the links. Thus, even in code only proper links should be useful.

I hope I didn't forget to address anything or skipped over something important, if that happened please do accept my apologies and let me know so I can address what's missing.

In any case, let's merge this 🙏!

EliahKagan:
> This is incredible work! I don't even know how to express how incredible it is as I don't know how it's even humanly possible to read through this much code while doing all these small to bigger improvements, with most of them being related to something as abstract as whitespace.

This was actually faster than #1725, in part because of its narrower scope and my greater familiarity with the code, but in part because it required much less code reading, though I did have to read docstrings. The approach I took here was different from there, which is also why the individual commits here may not have been broken up quite as nicely: I did this in more of a breadth-first fashion, in multiple passes, some reading carefully and others skimming, and each time applying things I noticed on the previous pass, both about the documentation and about what I had caught myself missing.

The biggest whitespace changes were *(a)* in formatting of `:role:`/`:role args...`-led sections, which were very fast to do manually, and *(b)* wrapping, which was simpler than in #1725 because I was applying an 88-column wrap everywhere except where it clearly worsened readability, rather than trying to infer wrapping in each place from local context.

Because single newlines (in most places equivalent to a single space when rendered) often appear (and occasionally I even added them) to improve readability or ease of editing in the docstrings themselves, and also because the wrapping tool I used (the [Rewrap](https://github.com/stkb/Rewrap) extension in VS Code) doesn't automatically observe all reStructuredText syntax rules when wrapping Python docstring text, I wasn't able to just select entire docstrings and auto-wrap them in most cases. But I was able to select a sizable contiguous chunks of lines at a time and do it, which also aided in keeping my place and noticing other issues to fix.

> I cannot imagine how a workflow around these changes must look like unless there is a way to automatically build and preview exactly the part of the file one is currently working on.

I don't have such a tool. VS Code renders docstrings when one hovers over the symbol, but it does so in kind of a fuzzy way that mostly interprets them as Markdown. I mostly worked in the docstring source itself, but I previewed changes with `make -C doc html` regularly. The [Live Server](https://github.com/ritwickdey/vscode-live-server) extension for VS Code helped here, though I probably could've instead used `file://` URLs and refreshed a lot.

> As for my own review, I tried to use [the renderered docs of the PR](https://gitpython--1850.org.readthedocs.build/en/1850/reference.html?highlight=handle_process_output) sporadically to get an idea of how it looks like. Nice initially, it quickly turned out that plenty of functions I would have expected to be findable through search simply weren't, so a preview wasn't really possible after all.

The API Reference is all on one page, so shortcomings in the provided search can sometimes be overcome by the web browser's own search (<kbd>Ctrl</kbd>+<kbd>F</kbd> for me). The first match will often be a mere *mention* of the symbol one seeks, but now that nearly all of these mentions link to the symbol, clicking it should be sufficient to get there. In hindsight I ought to have given that as the primary rationale for making internal class (and method) names links everywhere there was not strong reason not to do so: people often search for them and find something else that mentions them, and now they can usually click through immediately to what they are really looking for.

However, that the Read the Docs search--assuming I am right in thinking that this is what you mean--was not working as well as you expected deserves attention, and may be something we can fix. When looking into this, I found a problem which I hope to open a GitPython issue for soon. But I suspect it is different from what you are describing, because the problem I encountered does not affect searching in pull request builds. Can you say more about how you searched, and what you had expected would produce good results but did not?

> *[...]* It would also be my expectation that IDEs can use these docstring links to allow the user to jump to the linked classes directly, or maybe they even render the docs in-line while resolving the links. *[...]*

I don't have this functionality in VS Code for Sphinx documentation in Python docstrings, though there may be extensions that provide it, or maybe there is even something I'm unaware of that I just need to configure (but nothing in a docstring is ever treated by my editor as a symbol). I suspect I would get it if using a more IDE-ish IDE, such as PyCharm (or Writerside, a documentation IDE I've been meaning to try).

However, symbol search is fast and accurate in VS Code for most Python projects, including GitPython, so I was able to look up symbols efficiently (including by selecting them in docstrings and pressing <kbd>Ctrl</kbd>+<kbd>T</kbd>).

---

Regarding https://github.com/gitpython-developers/GitPython/pull/1850#issuecomment-1970779315:

I was unable to reproduce the CodeQL check failure or alerts in my own fork, after trying CodeQL checks on both the `push` and [`pull_request`](https://github.com/EliahKagan/GitPython/pull/4) triggers. It also somehow went away here as of [ee0301a](https://github.com/gitpython-developers/GitPython/pull/1850/commits/ee0301ab69e4f2af1248596a9efb897893c54da1). I don't know of any reason the content of that commit would ameliorate or otherwise affect CodeQL alerts.

---

Regarding https://github.com/gitpython-developers/GitPython/pull/1850#issuecomment-1970711452:

> It's too bad that it's not possible to tell GitHub to build every commit in the PR, so one is instead forced to push each commit individually

I wonder if there is some way to copy the status from the corresponding run in a fork that has already occurred.

> (which has side-effects).

Are there side effects of repeated pushes that I should be aware of, other than more entries in the remote reflog?

Byron:
> Can you say more about how you searched, and what you had expected would produce good results but did not?

I pasted the name of two functions, the second one I forgot but it was public (and pretty long), and the first one was `refresh()`.  There is two of these, and it only finds one of them which was initially confusing as the text didn't match. 
The other public method I searched for it didn't find at all. The browser search I tried to use, but it didn't work well for me.

Overall, I find the API documentation quality on a level of 'better than nothing', but it always was like that and by now I suppose I am spoiled by `rustdoc`.

> I wonder if there is some way to copy the status from the corresponding run in a fork that has already occurred.

This would be very clever, and could save a lot of compute as well! 

> Are there side effects of repeated pushes that I should be aware of, other than more entries in the remote (and local) reflog?

I get one email per push, in the moment the push is received😅. It's fine though, my email is setup none-invasively to not break focus.

EliahKagan:
> I pasted the name of two functions, the second one I forgot but it was public (and pretty long), and the first one was `refresh()`. There is two of these, and it only finds one of them which was initially confusing as the text didn't match.

When I search for `refresh` or `refresh()`, whether locally, hosted by Read the Docs for a PR, or hosted by Read the Docs for the tip of the main branch (i.e., "latest"), it finds both `Git.refresh` and `FetchInfo.refresh`. The one people should actually use has not been emitted at all, as noted in #1854, but it should be visible in the PR preview build for #1855 (and in "latest" if that PR is merged).

> The other public method I searched for it didn't find at all. The browser search I tried to use, but it didn't work well for me.

I have had some trouble with some other search terms. For example, `TagObject` doesn't always turn up particularly useful results. I still haven't reproduced the details of this and opened an issue for it, but when I do maybe that will help lead to an improvement.

> I get one email per push, in the moment the push is received😅. It's fine though, my email is setup none-invasively to not break focus.

I was unaware that notifications were generated for PRs that are opened as drafts and still marked draft!

EliahKagan:
> It's fine though, my email is setup none-invasively to not break focus.

I have taken this to mean that it is fine for me to continue pushing commits in the way I have been doing (in cases where they would otherwise not have CI check statuses). However, if that is not the case and you'd prefer I stop doing that, you can let me know.

PR #1855, opened 2024-03-02
Expand what is included in the API Reference

### LINKED ISSUE #1854: Too few modules are included in reference.rst
The [API Reference](https://gitpython.readthedocs.io/en/latest/reference.html) in the Sphinx documentation is missing some modules. But I am unsure exactly which ones should be added. There seem to be three cases.

At least one omission should be corrected:

1. The top-level `git` module is not listed, and while its `__version__` attribute it listed explicitly, `git.refresh` is not, so the `Git.refresh` method that should only be called directly in rare, unusual cases has its documentation included in the reference page, while the `git.refresh` function that is much more reasonable to call is omitted.

But for some of the other modules that aren't listed there, it seems intentional. These appear to divide into two cases:

2. Modules that are packages where everything interesting is defined in submodules.

4. Modules where it may make sense to avoid creating the appearance that users of GitPython should ever directly use anything from them.

These are all the modules of GitPython that are not listed in `doc/source/reference.rst`, such that they do not have their public symbols automatically collected, and presented with docstrings, by Sphinx autodoc:

```text
git
git.compat
git.db
git.index
git.objects
git.objects.submodule
git.refs
git.repo
git.types
```

Case 1, which should be included at least to get `git.refresh` and its docstring, is of course:

```text
git
```

Case 2, where maybe they shouldn't be included because nothing of interest is directly contained in them, are:

```text
git.index
git.objects
git.objects.submodule
git.refs
git.repo
```

Case 3, where maybe they shouldn't be included, or shouldn't be included without making clear in their module docstrings that they should not usually be used, are:

```text
git.compat
git.db
git.types
```

Regarding this third case, symbols in these modules are referenced in docstrings or type annotations. More of these references are references from a technical Sphinx perspective than before, as of #1850, but it was always necessary to be able to understand what was in them, to understand what some names referred to and their significance.

Therefore, for this third case, even if I am correct in my belief that they were omitted to discourage direct use, it seems to me that it may be best include them, with suitable guidance added in module docstrings (which autodoc will ensure is rendered at the top of the relevant section).

I am unsure what should be done here, or of all three of those modules of the third kind are really in the same situation. But even if they are and the approach I suggest is to be taken, I am unsure how strongly their use should be discouraged, and if a reason should be given (and if so, what reasons). This is what I know about them, that is related to this:

- `git.compat` is to some extent a relic of when Python 2 was supported, and besides that it is mostly used within GitPython to deal with encoding, yet it may no longer be needed if/when GitPython's handling of strings and bytes is improved in the future, and it may not be useful for use in codebases that handle such things correctly from the start.
- `git.db` is related to the gitdb library, and one of the symbols is includes in `__all__` is `GitDB`, whose use is today discouraged. The other symbol is `GitCmdObjectDB`. That might be seen today as a workaround for the way GitPython expected to receive an injected object database dependency with a `GitDB`-like interface. However, `GitCmdObjectDB` appears in various places and it's useful to understand it to understand the significance of it being the default. It seems to me that the `git.db` module should not only be included in documentation, but that no special effort should be made to discourage its use (though the `GitDB` could be clarified as not recommended).
- `git.types` provides types used throughout GitPython, and the `assume_never` function. The types do not have docstrings. However, it may still be useful to list them so references in Sphinx documentation to them are rendered as links and so it is clear they really are part of GitPython rather than external. In addition, some of the types are `TypedDict`s, and I *think* autodoc will list the members. (If what ought to be done hinges on that, then I can to an experiment first to check; you can let me know.)

My sense is that `git.db` and `git.types` should be promoted to have the same status as most modules; while `git.compat` should either be similarly promoted, or a discouragement to use it added to its docstring, or both.

These are all part of GitPython's public interface (they are named as public symbols and while the documentation currently omits pages for them it nowhere characterizes them as private or states that they should not be used in code outside GitPython). So to avoid breaking backward compatibility, they should remain public. But public use of them could be *deprecated* if appropriate. As far as I can tell, that should either be done in `git.compat` only or nowhere. If it is to be done for `git.compat` then it could be accompanied by renaming it to `git._compat` and having uses in GitPython use that name, while `git.compat` would re-export its public names.

EliahKagan:
Fixes #1854

This adds some missing modules in the API Reference. It adds the modules covered under "case 1" (top-level `git`, but only `refresh` and, as before, `__version__`) and "case 3" (the three leaf modules whose documentation wasn't previously emitted), as I classified them in #1854. It does not add "case 2".

Even if this is what should be done, there is some question about the specifics:

- I don't know what top-level entries should be grouped under in the existing naming convention. Its relationship to the actual module names is usually predictable but not precise. The general pattern doesn't allow me to predict a useful name for the section at the top that now contains `git.__version__` and `git.refresh`; it was previously called "Version", which is no longer suitable. I picked "Top-Level", but this feels not altogether in keeping with the rest of the style.
- I don't know what order the sections for `git.compat`, `git.db`, and `git.types` should come in, relative to the existing sections and each other. The existing sections are ordered in a largely but not entirely alphabetical fashion, and that seems like it may be intentional. I placed them at the end with the section for `git.util` and I kept those modules at the end sorted alphabetically relative to each other (so they are actually before `git.util`). That seems to make sense, since their documentation may be less often consulted than than of other modules. But I don't know if that's really true.

Each commit is self-contained, addressing one small part of these changes or other small changes that support or relate to them. While doing this, I noticed opportunities to improve some of the especially affected docstrings or those of closely related functions; those improvements are the first four commits.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -389,9 +389,12 @@ class Git:
 
     @classmethod
     def refresh(cls, path: Union[None, PathLike] = None) -> bool:
-        """This gets called by the :func:`git.refresh` function.
+        """Update information about the git executable :class:`Git` objects will use.
 
-        See the top level ``__init__.py``.
+        Called by the :func:`git.refresh` function in the top level ``__init__``.
+
+        This gets called by the :func:`git.refresh` function in the top-level
+        ``__init__``.
 
         :param path:
             Optional path to the git executable. If not absolute, it is resolved
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -389,7 +389,9 @@ class Git:
 
     @classmethod
     def refresh(cls, path: Union[None, PathLike] = None) -> bool:
-        """This gets called by the refresh function (see the top level ``__init__``).
+        """This gets called by the :func:`git.refresh` function.
+
+        See the top level ``__init__.py``.
 
         :param path:
             Optional path to the git executable. If not absolute, it is resolved
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1491,7 +1491,9 @@ class Git:
     def _parse_object_header(self, header_line: str) -> Tuple[str, str, int]:
         """
         :param header_line:
-            <hex_sha> type_string size_as_int
+            A line of the form::
+
+                <hex_sha> type_string size_as_int
 
         :return:
             (hex_sha, type_string, size_as_int)
@@ -1581,7 +1583,7 @@ class Git:
         return (hexsha, typename, size, data)
 
     def stream_object_data(self, ref: str) -> Tuple[str, str, int, "Git.CatFileContentStream"]:
-        """Similar to :meth:`get_object_header`, but returns the data as a stream.
+        """Similar to :meth:`get_object_data`, but returns the data as a stream.
 
         :return:
             (hexsha, type_string, size_as_int, stream)
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -393,9 +393,6 @@ class Git:
 
         Called by the :func:`git.refresh` function in the top level ``__init__``.
 
-        This gets called by the :func:`git.refresh` function in the top-level
-        ``__init__``.
-
         :param path:
             Optional path to the git executable. If not absolute, it is resolved
             immediately, relative to the current directory. (See note below.)
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -3,12 +3,15 @@
 API Reference
 =============
 
-Version
--------
+Top-Level
+---------
 
 .. py:data:: git.__version__
 
    Current GitPython version.
+
+.. automodule:: git
+   :members: refresh
 
 Objects.Base
 ------------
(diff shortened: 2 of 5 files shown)


git/cmd.py
```diff
@@ -389,7 +389,12 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
 
     @classmethod
     def refresh(cls, path: Union[None, PathLike] = None) -> bool:
-        """This gets called by the refresh function (see the top level ``__init__``).
+        """Update information about the git executable :class:`Git` objects will use.
+
+        Called by the :func:`git.refresh` function in the top level ``__init__``.
+
+        This gets called by the :func:`git.refresh` function in the top-level
+        ``__init__``.
```
Byron, line 397:
```suggestion
```

This seems like a duplicate now.

Byron (approved):
Thanks so much, this is a huge improvement!

For a lack of a better idea, `Top-Level` seems to do the job, and so does the order of the modules that now are documented additionally.

git/cmd.py
EliahKagan: Thanks, I'm not sure how I missed that!

PR #1859, opened 2024-03-04
Improve static typing and docstrings related to git object types

### LINKED ISSUE #1857: assert_never raises different exception than documented
`git.types.assert_never` is documented to raise `AssertionError` if `raise_error` is `True` and `exc` is `None`:

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/types.py#L83

But it actually raises `ValueError` in this situation, its full implementation being:

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/types.py#L85-L89

For comparison. there is an `assert_never` function in Python's `typing` module since 3.11 (and in `typing_extensions`). This function is similar but only has one parameter--which the first parameter of `git.types.assert_never` corresponds to--and it documents *that* it raises an exception if called at runtime, but does not document the type of exception raised. At least currently in Python 3.12, [it raises `AssertionError`](https://github.com/python/cpython/blob/0a9356ce6ddca117588c95480f0ef1ef33df03cb/Lib/typing.py#L2406).

The GitPython function is a public member of `git.types`, and it is used internally in GitPython. I don't know if the docstring should be fixed to describe the existing behavior, or if that behavior should itself be considered the bug here and changed to match the docstring.

### LINKED ISSUE #1858: Commit_ish is much broader than commit-ish
In git, if I understand correctly, a *commit-ish* is a git object from which a commit can be reached by dereferencing it zero or more times, which is to say that all commits are commit-ish, some tag objects are commit-ish--those that, through (possibly repeated) dereferencing, eventually reach a commit--and no other types of git objects are ever commit-ish.

As [gitglossary(7) says](https://git-scm.com/docs/gitglossary#def_commit-ish):

> #### commit-ish (also committish)
>
> A [commit object](https://git-scm.com/docs/gitglossary#def_commit_object) or an [object](https://git-scm.com/docs/gitglossary#def_object) that can be recursively [dereferenced](https://git-scm.com/docs/gitglossary#def_dereference) to a commit object. The following are all commit-ishes: a commit object, a [tag object](https://git-scm.com/docs/gitglossary#def_tag_object) that points to a commit object, a tag object that points to a tag object that points to a commit object, etc.

Therefore, *all* instances of GitPython's `Commit` class, and *some* instances of GitPython's `TagObject` class, encapsulate git objects that are actually commit-ish.

But GitPython has a `Commit_ish` union type in the `git.types` module, and that `Commit_ish` type is considerably broader:

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/types.py#L53

These four classes are the GitPython classes whose instances encapsulate any of the [four types of git objects](https://git-scm.com/docs/gitglossary#def_object_type) (of which blobs and trees are never actually commit-ish):

> #### object type
>
> One of the identifiers "[commit](https://git-scm.com/docs/gitglossary#def_commit_object)", "[tree](https://git-scm.com/docs/gitglossary#def_tree_object)", "[tag](https://git-scm.com/docs/gitglossary#def_tag_object)" or "[blob](https://git-scm.com/docs/gitglossary#def_blob_object)" describing the type of an [object](https://git-scm.com/docs/gitglossary#def_object).

GitPython uses its `Commit_ish` type in accordance with this much broader concept, at least some of the time and possibly always. For example, `Commit_ish` is given as the return type of `Object.new`:

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/objects/base.py#L77-L78

`Commit_ish` cannot simply be replaced by `Object` because GitPython's `Object` class is also, through `IndexObject`, a superclass of `Submodule` (and the `RootModule` subclass of `Submodule`):

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/objects/submodule/base.py#L82

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/objects/submodule/base.py#L87-L88

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/objects/submodule/base.py#L100-L101

However, elsewhere in GitPython, `Commit_ish` is used where it seems only a commit is intended to be allowed, though it is unclear if this is unintentional, intentional but only to allow type checkers to allow some code that can only reasonably be checked at runtime, or intentional for some other reason. For example, the `Repo.commit` method, when called with one argument, looks up a commit in the repository it represents from a `Commit_ish` or string, and returns the commit it finds as a `Commit`:

https://github.com/gitpython-developers/GitPython/blob/b2c3d8b72301dc4723e83de676fd7086d5c3c381/git/repo/base.py#L698

This leads to a situation where one can write code that type checkers allow and that may appear intended to work, but that always fails, and in a way that may be unclear to users less familiar with git concepts:

```text
>>> import git
>>> repo = git.Repo()
>>> tree = repo.tree()
>>> repo.commit(tree)
  File "C:\Users\ek\source\repos\GitPython\git\repo\base.py", line 709, in commit
    return self.rev_parse(str(rev) + "^0")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ek\source\repos\GitPython\git\repo\fun.py", line 379, in rev_parse
    obj = to_commit(obj)
          ^^^^^^^^^^^^^^
  File "C:\Users\ek\source\repos\GitPython\git\repo\fun.py", line 221, in to_commit
    raise ValueError("Cannot convert object %r to type commit" % obj)
ValueError: Cannot convert object <git.Tree "d5538cc6cc8839ccb0168baf9f98aebcedfd9c2c"> to type commit
```

An argument that this specific situation with `Repo.commit` is not a typing bug is that this operation is fundamentally one that can only be checked at runtime in some cases. After all, an argument of type `str` is also allowed and it cannot known until runtime what object a string happens to name. Even so, the method docstring should possibly be expanded to clarify this issue. Or perhaps if the situation with `Commit_ish` is improved, then the potential for confusion will go away.

One way to improve this situation is to clearly document it in a docstring for the `Commit_ish` type. But if possible it seems to me that more should be done:

- If known, the reason for the current situation should be stated there.
- Its relationship to other types should be clarified where otherwise confusing. For example, `Object` may benefit from greater clarity about what it ideally represents (git objects) versus the entirety of what it represents (that an `Object` can also be a `Submodule`), and the way that `Tree_ish` is narrower than all tree-ish git objects while `Commit_ish` is broader than all commit-ish git objects can be noted in one of their docstrings.
- Maybe `Commit_ish` should be deprecated and one or more new types introduced, replacing all uses of it in GitPython.

If I am making a fundamental mistake about git concepts here, and GitPython's `Commit_ish` has a closer and more intuitive relationship to commit-ish git objects than I think it does, then I apologize.

I have not figured out very much from GitPython's revision history what the reason for defining `Commit_ish` as it is currently defined is, or alternatively why this union of all four actual git object types was introduced with the narrower-seeming name `Commit_ish`. However, the `Commit_ish` type was introduced in 82b131c (#1282), where the annotations it was used to replace had listed all four types `Commit`, `TagObject`, `Tree`, and `Blob` as explicit alternatives.

### LINKED ISSUE #1866: Wrong type annotation on `Index.commit` for `parent_commits`
With `GitPython==3.1.42` one can observe following annotation of `parent_commits` argument to `IndexFile.commit` ([1]):

```
...
parent_commits: Union[Commit_ish, None] = None,
...
```

which is obviously not correct, because that argument just passed unmodified downwards to `Commit.create_from_tree`, which has following annotation instead ([2]):

```
parent_commits: Union[None, List["Commit"]] = None
```

That leads to nasty Mypy warnings when one's trying to use `IndexFile.commit` with anything different than `None`.

---
[1] - https://github.com/gitpython-developers/GitPython/blob/3.1.42/git/index/base.py#L1080
[2] - https://github.com/gitpython-developers/GitPython/blob/3.1.42/git/objects/commit.py#L506

### LINKED ISSUE #1873: Module docstring says TagReference but means TagObject
The `git.objects.tag` module holds the `TagObject` class that represents tag objects, representing annotated tags, which are git objects, while of course the `TagReference` class represents lightweight tags, which are git references.

When updating docstrings in 6126997 (#1850), I tried to say this, and even was aware that it was easy to write the wrong identifier while editing multiple docstrings, and I checked for this mistake, yet still it was not enough and I made the docstring say the exact opposite:

https://github.com/gitpython-developers/GitPython/blob/e880c33284f2abff55d3cddfef2302663b178005/git/objects/tag.py#L8-L9

A fix for this is included in #1859, at 1cdec7a, and I anticipate that #1859, expanded to properly fix #1858 and integrate changes from other PRs, will be ready for review very soon. Nonetheless I'm opening this issue so that, if it turns out there are delays and a release has to be made, that commit can be cherry-picked or its fix otherwise applied.

### LINKED ISSUE #1874: Diffable.diff is misleadingly annotated regarding special `other` values
The `other` parameter of `Diffable.diff` is annotated this way:

https://github.com/gitpython-developers/GitPython/blob/e880c33284f2abff55d3cddfef2302663b178005/git/diff.py#L110

`Diffable.diff` accepts a few special values:

- `None` for a working tree diff.
- `Diffable.Index`, a `type` object that is meant to be used opaquely rather than instantiated, to compare to the index.
- `NULL_TREE`, a direct instance of `object` (not to be confused with `Object`), for a diff against an empty tree.

#### The type of `NULL_TREE` should be expressed better than `object`

The main problem here, which exists in the documentary effect of the annotations even if a type checker is never used, is the presence of `object` in the union. Since `object` is the root of the type hierarchy in Python, a union of anything with `object` is equivalent to just `object`. But the intent of the annotation is not to express that it is correct to pass arbitrary objects as `other`. Instead, the idea is that it is okay to pass `NULL_TREE`.

This is also the cause of the incompatible override type error on the `IndexFile.diff` override in `git.index.base`, which does not include the `object` alternative:

https://github.com/gitpython-developers/GitPython/blob/e880c33284f2abff55d3cddfef2302663b178005/git/index/base.py#L1482

There are several places in GitPython's source code where an overridden method has an incompatible signature, violating substitutability. The reason this place is of interest is that it is hard to tell efficiently by examining the code or type errors *what* conceptually this means. It turns out it means the override does not support `NULL_TREE`:

```text
>>> import git
>>> repo = git.Repo()
>>> repo.index.diff(git.NULL_TREE)
  File "C:\Users\ek\source\repos\GitPython\git\index\base.py", line 1521, in diff
    raise ValueError("other must be None, Diffable.Index, a Tree or Commit, was %r" % other)
ValueError: other must be None, Diffable.Index, a Tree or Commit, was <object object at 0x00000265862B8A30>
```

But `object` does not express (to humans or tools) that what is really meant is the literal value `NULL_TREE`.

We cannot currently write `Literal[NULL_TREE]`, but if `NULL_TREE` is made into an enumeration constant, then this can be done, though cumbersomely it seems it can only be written--in annotations--as `ET.NULL_TREE` where `ET` is the declared enumeration type. Or, if `ET` has `NULL_TREE` as its only value, then writing `ET` should be sufficient, though this may be less intuitive due to not involving `Literal` in any way while expressing a literal.

To be clear, this will not cause the mypy error about an incompatible override to go away. But it will make the error make sense, as well as making the *suppression* comment on it make sense in terms of what it is expressing. *And it will allow type checkers to catch errors about incorrect values of `other`.*

#### `Diffable.Index` should ideally also be improved

`Diffable.Index` is used as an opaque constant, as is `NULL_TREE`. But `Diffable.Index` is a class. This is confusing because it should not be instantiated. Its static type can be expressed pretty specifically, as `Type[Diffable.Index]` (as is done; I'm omitting the quotes here for simplicity). However, this does not allow `if`-`else` logic to be type-checked exhaustively, because there is no guarantee that the `Diffable.Index` class is the only value of the static type `Type[Diffable.Index]`:

```python
class Derived(Diffable.Index):
    pass
```

Hopefully no one is doing that--just as hopefully no one is instantiating it--but `mypy` and other tools, including editors, cannot tell that `Diffable.Index` is the only value of `Type[Diffable.Index]`. Humans who notice that it is intended to be used as an opaque constant *will* recognize this, however. Also, it *should* be possible to get type checkers to recognize it as the only value, and type-check `if`-`else` logic exhaustively, by decorating the definition of `Index` with `@final`. This works for some type checkers, at least `pyright` and `pylance`. But it does not work for `mypy`.

The bigger issue is that it is easy to confuse and, for example, instantiate it.

`Diffable.Index` can also be made an enumeration constant, and all reasonable usage will continue to work. *At runtime.* However, if someone wants to statically annotate their own method to accept or return it, then they would need to change the annotations, since expressing it with `Type[Diffable.Index]` would no longer be able to work. (Two separate special objects cannot both be used, because existing subclasses of `Diffable` outside of GitPython would only be covering one. But if it is redefined, existing code can continue to cover it at runtime, since it should only ever have been, or be, used as an opaque constant.)

It is also possible for unreasonable usage to break at runtime, such as attempting to check if `x` is `Diffable.Index` by writing `issubclass(x, Diffable.Index)`. I am less worried about this.

It seems to me that this change is worth making for `Diffable.Index` as well, but this is a judgment call, and the new interface should perhaps be given special attention in review. When I searched for existing code outside GitPython, its forks, vendored copies of it, etc., that used these annotations in a way where they would cause a new static type error, I didn't find anything, but I only used GitHub code search, and I may not have found everything even on GitHub. My argument for making this change is to achieve greater correctness without breaking anything at runtime, and *not* that I am at all confident nobody will have to change their code to keep it passing static checks.

#### Connection to #1859

I've included fixes for both of these things in #1859. See [65863a2](https://github.com/gitpython-developers/GitPython/pull/1859/commits/65863a28c2bfebe084e1c86d409fbf68b0b33a76) especially, whose docstring has some more information about the specific design choices there.

If you decide you don't want the change to `Diffable.Index` there, that can be undone without sacrificing the rest. Although I am in favor of this aspect of the change, it should not be accepted on the basis of the sunk-cost fallacy. Furthermore, the sunk cost would be quite low, because it was already helpful in enabling me to figure out what needed to be done in other related parts of the code.

The purpose of this issue is twofold:

- To allow the bug to be tracked if [#1859](https://github.com/gitpython-developers/GitPython/pull/1859) is changed in such a way as not to fully solve it.
- To described it separately from that PR so it's easier to understand. I considered describing it in comments there, but thought opening an issue would be better.

### LINKED ISSUE #1875: Top-level refresh function not listed in __all__
Since #1659, `git.__all__` is written literally, rather than being dynamically constructed based on what has been defined at the time it is bound. However, the fix was based on what `__all__` had been being populated with at the time, and therefore inherited two weaknesses:

- Lots of imports that don't make sense to use from the `git` module (because they are from the standard library) are included. This is a fairly minor issue, but I suggest marking them as deprecated. Probably `DeprecationWarning` should not be raised when accessing them, though, because although people should not use wildcard imports in general, such warnings when using them in a REPL could obscure more important deprecation warnings.
- **`git.refresh` isn't public**, at least not by the convention that when `__all__` exists everything excluded from it is non-public. This is mitigated by the references in the documentation that say to use it and to prefer it to the `Git.refresh` method, and by its inclusion in the autodoc-generated API reference page, but it would still be best for it to be listed in `git.__all__`. The reason it isn't is that, back when `__all__` had been constructed dynamically, this was *above the `def` for `refresh`*.

I've included a fix for this in #1859, since I was adding to `__all__` for #1874 already. Because PEP-8 recommends `__all__` come before most code in a module, and the reason that couldn't be done before was that it was being built dynamically and so had to be defined below everything that was to be included in it, I went ahead and did that too. As with #1874, in view of the grown scope of #1859, it seemed clearer to have an issue for this that #1859 lists as fixing, rather than jam it into comments there.

---

The other aspect of the situation with `git.__all__` following #1659 is the question of whether its content really has been stable. After all, it was constructed from wildcard imports from modules whose contents sometimes changed and not all of which had `__all__` attributes, and so potentially could have changed heavily across releases. If it had, then it might be acceptable to remove things that are only there for compatibility (because: compatibility with what?) and also would be necessary to add at least a commented warning when making other substantial changes, to avoid obscuring the issue.

Fortunately, that is not the case. It has been remarkably stable. I wrote [a script to check](https://github.com/EliahKagan/deltall) and [the results can be checked here](https://gist.github.com/EliahKagan/12c137a437069fae267facb1e4132180). (I actually did this around the time of #1659, but there was nothing to raise an alarm about, and the matter didn't come up again until now.)

EliahKagan:
Fixes #1857
Fixes #1858
Fixes #1866
Fixes #1873
Fixes #1874
Fixes #1875

#### Main changes

This expands the docstrings of `Object` and some of its subclasses. These revisions are, in large part, to explain the classes' relationships to the git concept of objects and to the corresponding git object types, including links to relevant definitions in [gitglossary(7)](https://git-scm.com/docs/gitglossary).

This also adds docstrings to all the types introduced in `git.types`--those that relate directly to git object types and those that do not. Adding documentation in `git.types` relates topically to the docstring expansions for `Object` and some of its subclasses, and I tried to make all new and edited docstrings make sense whether read by themselves or together with other docstrings across modules. It also relates to the goal for `git.types` documentation articulated in https://github.com/gitpython-developers/GitPython/issues/1854#issuecomment-1974749614.

#### Other changes

This also includes some other changes that are less core to that scope but that seem practically useful to include. It revises the docstring for `assert_never`, which had been the only docstring in `git.types`, and modifies the string of the default exception it raises to print the `repr` rather than the `str`. It removes some long-unused nonpublic names at module level, and revises some method docstrings of classes whose class docstrings were being revised. It also expands the `git.compat` module docstring, toward the goal for it that was also articulated in https://github.com/gitpython-developers/GitPython/issues/1854#issuecomment-1974749614.

#### Considerations for `assert_never`

The change to `assert_never`, which includes making the docstring accurately describe the current (and established) behavior of the function, may fix #1857. However, as described there I am unsure if this is ultimately the right change there.

#### Considerations for `Commit_ish` docstring (including possible inaccuracies)

Even more important is the relationship of the `Commit_ish` docstring to #1858. I have said this PR "may" fix that issue, but I strongly suspect that either further refinement will be needed or that a significantly greater change, possibly outside the scope of this PR, may be needed.

Furthermore, some of the text in that docstring may require improvement before being suitable for merging. To facilitate writing the docstrings in `git.types`, including related docstrings and even other parts of the `Commit_ish` docstring, I wrote text in two parts of the `Commit_ish` docstring **that sound plausible but that I am not at all sure are correct.**

This is in regard to the question, detailed in #1858, of *why* `Commit_ish` union includes the never-commitish `Tree` and `Blob` classes. (Even if what I wrote turns out to be correct, it may need to be clarified.) I have opened review comments on those potentially inaccurate lines.


git/types.py
```diff
@@ -33,85 +33,165 @@
...
+    This is done because of the way GitPython uses it as a static type annotation.
+
+    :class:`~git.objects.tag.TagObject`, some but not all of whose instances are
+    tree-ish (those representing git tag objects that ultimately resolve to a tree or
+    commit), is not covered as part of this union type.
+"""
+
 Commit_ish = Union["Commit", "TagObject", "Blob", "Tree"]
+"""Union of the :class:`~git.objects.base.Object`-based types that represent git object
+types. This union is often usable where a commit-ish is expected, but is not actually
```
EliahKagan, line 66: This is the first of two places in the `Commit_ish` docstring where I have said something that sounds plausible but may be inaccurate. It is really fair to say that this is "often usable where a commit-ish is expected"? And does saying this have documentary value? Here I am making it sound like the name and definition of `Commit_ish` make some amount of sense in view of what a commit-ish actually is, in the operation of GitPython. If that is not true, then this should be revised.

git/types.py
```diff
@@ -33,85 +33,165 @@
...
+
+    * A :class:`~git.objects.commit.Commit` is of course always commit-ish, and a
+      :class:`~git.objects.tag.TagObject` is commit-ish if, when peeled (recursively
+      followed), a :class:`~git.objects.commit.Commit` is obtained.
+    * However, :class:`~git.objects.blob.Blob` and :class:`~git.objects.tree.Tree` are
+      also included, and they represent git objects that are never really commit-ish.
+
+    This is an inversion of the situation with :class:`Tree_ish`, which is narrower than
+    all tree-ish objects. It is done for practical reasons including backward
+    compatibility.
```
EliahKagan, line 86: This is the second of two places in the `Commit_ish` docstring where I have said something that sounds plausible but may be inaccurate. The last sentence here makes two vague claims, which even if vague should only be kept if true (or maybe removed even then, if not useful), and which preferably should be made less vague.

Byron (commented):
Thanks so much for your help with this!

I made a few remarks, but the biggest question certainly is if `Commit_ish` can be fixed instead of trying to justify it in the documentation. To me it seems possible, but I may be very wrong about that.

Thanks for sharing your thoughts.

EliahKagan:
As detailed below, this is not yet finished or ready to merge, even though I am pushing some more commits.

> *[...]* but the biggest question certainly is if `Commit_ish` can be fixed instead of trying to justify it in the documentation. To me it seems possible, but I may be very wrong about that.

Yes, I will do this. I think I am most of the way done. But while I'd hoped to finish it up by the end of the weekend (i.e., by about right now), I definitely won't manage it by then. So I figured I'd give an update.

This is more interesting--and also more involved, to do well--than it may seem at first. Different uses of `Commit_ish` need to be replaced with different things. Two unions are needed to correspond to most (though still not all) such cases. Those changes have other effects, and also require other changes to be made to accommodate them. This also has tendrils into some other problems in the type annotations, some of which are obviously related and others of which are related in subtle ways where, it seems, the effect of an incorrect annotation is amplified by other annotations that are incorrect in order to accommodate it. I think this task cannot reasonably be done without fixing those as well. But another way to say that is that it is an opportunity to make those other improvements to the annotations.

But that is getting a bit ahead of things. To do that with any confidence that I was fixing more errors than I am introducing and making things better rather than worse, I needed the help of mypy to check the annotations. Unfortunately, mypy reports 21 errors on the main branch. This is really 23 errors, since two errors occur only on Windows, and 2 other ones occur only on non-Windows platforms. This needed to be thinned out for mypy to be useful for me here, but I was reluctant to add new suppressions. This was, in part, because doing so would somewhat weaken my ability to use mypy to examine my changes (the less mypy can reveal the worse, and new suppressions done without examining the situation may be more likely to be hiding something important than long-standing ones). But it was also in part because, unlike xfail markings as added in #1745, mypy suppressions don't much help in keeping track of the errors being set aside.

Therefore I went ahead and fixed most of these errors, decreasing the number of mypy errors from 21 (or 23, depending how you count) to 5. The output of mypy (with these fixes applied) is now readable, and the remaining errors each indicate something that should be fixed in GitPython or its annotations but that is beyond the (even now much-increased) scope of this PR. Really, I should not say I *fixed* them all, since for a few of them I eliminated them by fixing existing suppressions that were not working. However, I did not introduce any new suppressions that were not clearly already intended before with a preexisting attempt written as a `type: ignore` comment (some of which would once have worked but became misplaced during autoformatting), nor any new casts, uses of `Any`, etc.

**The commits I am pushing now are mostly just those related to fixing those mypy errors.** I have more done--as I mentioned above, I think I am pretty near finished with the whole thing. But at this time, the commits relating to the mypy errors, constituting a substantial pre-fixup allowing the rest to proceed more smoothly, are what seem like it might make sense to review. *This is if you feel like doing so, which you are under no obligation to do.* To be clear, I am doing all these changes with the intention that they be scrutable and reviewable together, and there is absolutely no need for to look at any of this now.

These changes include a redefinition of `Diffable.Index` that I believe is sufficiently backward compatible but that perhaps will merit special scrutiny. It is possible to do without it if necessary. Although it seems to me that it is worth keeping in, it can be removed. The insight gained by doing it, and by examining the result with `mypy`, actually has made it worthwhile even if it ends up not making the cut; a new commit could be added to replace it with a less type-correct but more conservative change.

(My full progress for this PR as of this time, including the commits not yet pushed that constitute most of a fix for #1858, *can* also be examined at 030001d and its [ancestors](https://github.com/gitpython-developers/GitPython/commits/030001ded690c760fc4bb637adcba14000161701/), but I emphasize that I don't think it is in good shape to be reviewed. In addition, the commits pushed here are those I am very unlikely to want to rebase, while for those after them this is less certain. I mention this only to make it slightly easier to find it if for some reason you were going to anyway. If you do look at them, which I very much suggest not bothering to do, then I hope you won't be alarmed by `Old_commit_ish`; this type will be removed altogether after the fixup is complete.)

Byron:
Thanks for the detailed update! There is never any pressure or expectations from me, I am just glad you are contributing :).

> This is more interesting--and also more involved, to do well--than it may seem at first. Different uses of `Commit_ish` need to be replaced with different things. Two unions are needed to correspond to most (though still not all) such cases. Those changes have other effects, and also require other changes to be made to accommodate them. This also has tendrils into some other problems in the type annotations, some of which are obviously related and others of which are related in subtle ways where, it seems, the effect of an incorrect annotation is amplified by other annotations that are incorrect in order to accommodate it. I think this task cannot reasonably be done without fixing those as well. But another way to say that is that it is an opportunity to make those other improvements to the annotations.

I absolutely love how expressively this is described - I could nearly 'feel' what it means to do this, that was quite incredible.

Beside that, I am happy to wait with my (armchair) review until it's good and ready, and hope you will avoid rebasing this time to not add additional work or burden you. To me, rebases are preferred if it's all the same, but that's not the case here.

EliahKagan:
As mentioned above, I've pushed some more commits of what's already done, and I'm also going to try my hand at retitling this, but this is not yet finished.

The specific reason for doing this now is to allow its relationship to #1866 to be inspected by anyone who is interested (including myself), but overall I don't think it would be especially useful to actually review this unfinished PR at this time.

EliahKagan:
I think this is ready.

Rather than writing a lot here about the most non-obvious aspects of the change, I've opened three more issues to describe them, #1873, #1874, and #1875, linked and marked as fixed in the edited PR description.

The core changes here remain the type annotation fixes throughout, reducing the number of mypy errors to 5, and of course; the fix for #1858, and directly and indirectly related annotations; and of course the originally scoped changes in adding docstrings in `git.types` and expanding them elsewhere, which, due to also addressing #1858, are able to be done much better than when I had originally opened this PR.

git/types.py
```diff
@@ -4,114 +4,248 @@
...
+PathLike = Union[str, "os.PathLike[str]"]
+"""A :class:`str` (Unicode) based file or directory path."""
 
 TBD = Any
+"""Alias of :class:`~typing.Any`, when a type hint is meant to become more specific."""
+
 _T = TypeVar("_T")
+"""Type variable used internally in GitPython."""
+
+AnyGitObject = Union["Commit", "Tree", "TagObject", "Blob"]
```
Byron, line 49: That's a great name!

git/types.py
```diff
@@ -4,114 +4,248 @@
...
+"""Union of :class:`~git.objects.base.Object`-based types that are inherently tree-ish.
+
+See gitglossary(7) on "tree-ish": https://git-scm.com/docs/gitglossary#def_tree-ish
+
+:note:
+    This union comprises **only** the :class:`~git.objects.commit.Commit` and
+    :class:`~git.objects.tree.Tree` classes, **all** of whose instances are tree-ish.
+    This has been done because of the way GitPython uses it as a static type annotation.
+
+    :class:`~git.objects.tag.TagObject`, some but not all of whose instances are
```
Byron, line 86: Thanks for mentioning this. It's great that this type includes what GitPython can handle though, as one day it might mean it can be taught to resolve tags as well.

Byron (approved):
What an impressive attention to detail, I can hardly imagine how it's possible to comb through this many files while keeping a keen eye out on complex types, and end up improving them.

Whether perfectly accurate or not, I believe this PR marks a huge leap towards a typing that makes sense and helps with usage.

Thank you!

git/types.py
EliahKagan: I had avoided changing this since it is not inherently a wrong definition of `Tree_ish`, but actually the situation you are describing, where it can be broader, is already the case. Furthermore, both `Commit_ish` and `Tree_ish` in `git.types` make a lot more sense if they both include `TagObject` rather than one of them including it and the other not. I've opened #1878 to add it.

EliahKagan:
Because of its effects on type checking of code that uses the GitPython library, this PR should probably be distinguished in some way, and as such may merit highlighting per https://github.com/gitpython-developers/GitPython/issues/1806#issuecomment-1978912760. If #1878 is accepted, then it contributes to this effect on type annotations, but so long as the effect is discernible I don't think they both have to be highlighted.

This touches on some questions about the goals being achieved by marking pull requests in this way, but that is not specific to this so I'll ask about it in the #1806 comments instead.

Byron:
Thanks for pointing this out - I have marked it. Please do feel free to do so yourself in future, I trust your judgement and know I am likely to forget it

EliahKagan:
Now that the purpose of the highlighting is clarified, per https://github.com/gitpython-developers/GitPython/issues/1806#issuecomment-1999173330, I'll go ahead and add it where I'm aware it's applicable. in future pull requests I open.

PR #1864, opened 2024-03-11
Use upgraded pip in test fixture virtual environment

EliahKagan:
A couple of GitPython's tests create their own virtual environments. One of those tests, `test_installation`, uses `pip` in it. But it does not upgrade `pip` in that virtual environment before using it, so an old version is often used. (The virtual environment is deliberately isolated, so `pip` being current in the environment used to run the tests is not sufficient.)

This creates a problem with `pip` on Cygwin. `pip` vendors `pytoml` to read `pyproject.toml` files. But the version of `pytoml` vendored by some old versions of `pip`--including the one currently used in a virtual environment in which `pip` has not been upgraded--is apparently unable to parse some syntactically correct TOML, at least on some platforms. This was undetected because the contents of `pyproject.toml` on the main branch do not trigger the bug.

This is the cause of the [Cygwin test failure](https://github.com/gitpython-developers/GitPython/actions/runs/8231126052/job/22516287983) in #1862 (at [a083e81](https://github.com/gitpython-developers/GitPython/pull/1862/commits/a083e8115807ff0a9b08a81d53e56190003e5354)). The failing test is `test_installation`; the detailed information on the failure [begins here in the log](https://github.com/gitpython-developers/GitPython/actions/runs/8231126052/job/22516287983#step:13:393). The key is [**these lines**](https://github.com/gitpython-developers/GitPython/actions/runs/8231126052/job/22516287983?pr=1862#step:13:447) from the error message:

```text
pip._vendor.pytoml.core.TomlError: /tmp/pip-req-build-yjylhtt5/pyproject.toml(61, 1): msg
WARNING: You are using pip version 19.2.3, however version 24.0 is available.
```

Often, such warnings are unrelated to errors that precede them, but in this case the low version of `pip` really is the cause. Changing the `VirtualEnvironment` test fixture to upgrade `pip` when it creates a virtual environment where `pip` is going to be used is sufficient to fix the problem.

**This pull request makes that change.** It also upgrades `setuptools` when the Python version is lower than 3.12, which is recommended to do. From Python 3.9 on, `venv.create` has an `upgrade_deps` parameter that would make all of this very straightforward...except that using that would make `test_installation` incompatible with Python 3.7 and Python 3.8 (including on systems other than Cygwin). So I ended up doing it manually instead.

---

This pull request has the main branch in the upstream GitPython repository as its base, but I have [also tested these changes](https://github.com/EliahKagan/GitPython/commits/ruff-pip-toml/) on top of #1862 (as it stands currently, with [a083e81](https://github.com/gitpython-developers/GitPython/pull/1862/commits/a083e8115807ff0a9b08a81d53e56190003e5354) at its tip) and verified that they fix the problem reported there and allow all CI checks to pass, including the Cygwin test workflow and the linting workflow with Ruff.

I'm not sure what the best way is to integrate the changes here and in #1862. This could be merged to main, and then #1862 could be updated. Or this could be merged, and #1862 merged without updating it, trusting that the checks would go green on the main branch. Or the changes here could be replayed there (e.g., by fast-forwarding it to my [ruff-pip-toml](https://github.com/EliahKagan/GitPython/commits/ruff-pip-toml/) branch), or that could be merged first with the failing check and then this could be merged, etc. (I have no objection if these changes are integrated in a way results in this pull request itself being closed without merging.)

### DIFF
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -10,6 +10,8 @@ import io
 import logging
 import os
 import os.path as osp
+import subprocess
+import sys
 import tempfile
 import textwrap
 import time
@@ -403,13 +405,20 @@ class VirtualEnvironment:
 
     __slots__ = ("_env_dir",)
 
-    def __init__(self, env_dir, *, need_pip):
+    def __init__(self, env_dir, *, with_pip):
         if os.name == "nt":
             self._env_dir = osp.realpath(env_dir)
-            venv.create(self.env_dir, symlinks=False, with_pip=need_pip, upgrade_deps=need_pip)
+            venv.create(self.env_dir, symlinks=False, with_pip=with_pip)
         else:
             self._env_dir = env_dir
-            venv.create(self.env_dir, symlinks=True, with_pip=need_pip, upgrade_deps=need_pip)
+            venv.create(self.env_dir, symlinks=True, with_pip=with_pip)
+
+        if with_pip:
+            # The upgrade_deps parameter to venv.create is 3.9+ only, so do it this way.
+            command = [self.python, "-m", "pip", "install", "--upgrade", "pip"]
+            if sys.version_info < (3, 12):
+                command.append("setuptools")
+            subprocess.check_output(command)
 
     @property
     def env_dir(self):
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -403,13 +403,13 @@ class VirtualEnvironment:
 
     __slots__ = ("_env_dir",)
 
-    def __init__(self, env_dir, *, with_pip):
+    def __init__(self, env_dir, *, need_pip):
         if os.name == "nt":
             self._env_dir = osp.realpath(env_dir)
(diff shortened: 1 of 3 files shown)


Borda (approved):

Byron (approved):
Thanks so much for helping with this!

The line that indicates the culprit is something I once again missed. It looks like I really can't investigate test-failures anymore, while also admitting that I only scroll all the way down to see the one-line summary of the failure. Searching for the test-case should help, next time.

Supporting this many python versions and platforms isn't easy, and I am impressed by your ability to dig through all that anyway.

EliahKagan:
Until an improvement is made related to the xfail distraction issue discussed at https://github.com/gitpython-developers/GitPython/pull/1853#issuecomment-1974005684 and https://github.com/gitpython-developers/GitPython/pull/1853#issuecomment-1974737033, a technique I've found useful--in addition to searching for the test-case name as you mention--is to scan upward for when the percentages shown at the ends of most lines change color from yellow/gold to red. This always marks the the first test with "FAIL" status in the test run.

PR #1868, opened 2024-03-12
Update readme and tox.ini for recent tooling changes

EliahKagan:
There are some changes that I should have proposed and/or made in `README.md` and `tox.ini` related to #1862, which are made slightly more needed by #1865. This brings the readme and tox configuration up to date for both #1862, which was merged, and #1865, which I think will be merged soon ~~pending one small change~~. Further related changes should be eventually be made in both, but they will depend on future decisions, and nothing bad will happen if changes end up not being made for an extended time after merging both #1865 and this PR.

The changes here seem significantly more cumbersome to recommend be included in #1865 than to do separately, and because they also apply to the situation brought about since #1862, I don't think they need to be done there or done before that is merged. I suggest #1865 be merged before this; the changes here will not be all correct until #1865 comes in anyway. #1865 could be merged and then this immediately merged, but I think it is also fine to merge #1865 and request changes here; that is, I think #1865 is reasonable to merge even if this is delayed. Merging them in the other order would also be okay, I think.

The core issue is that some actions that were presented as not changing file contents now change them. Currently on the main branch:

1. I believe running `tox` with no arguments is always in practice assumed not to change source code in the working tree. But its `lint` environment changes source code due to running `ruff` with `--fix` through `pre-commit`. With [#1865](https://github.com/gitpython-developers/GitPython/pull/1865), this will happen in more cases, but the issue already exists. As noted below and in commit messages, I believe `--fix` is nonetheless good and should be kept. (When I reviewed [#1862](https://github.com/gitpython-developers/GitPython/pull/1862) I did not mention this, because I agreed with the change, which is what users of `pre-commit` will expect. In hindsight, I could've simplified things by proposing these changes at that time.)
2. The readme documents `make lint` as linting and running a formatting check without modifying any files. This is likewise not fully accurate since [#1862](https://github.com/gitpython-developers/GitPython/pull/1862). The issue is not that `make lint` needs to behave differently, because *(a)* in hindsight I shouldn't have added that target, *(b)* I'm not sure anyone uses it, and *(c)* it was added to address a situation where checking for code style and formatting problems involved multiple commands and tool and plugin packages, instead of just `ruff`. So it's sufficient for the readme to stop recommending it as a way to avoid linting without automatic code changes.

This PR fixes *(1)* by making the tox `lint` environment not run unless explicitly listed on the command line, for now, and fixes *(2)* by changing the readme. This also fixes some related outdated material in the readme, improves how tools are described, and makes some some other small improvements when it seemed like I could do them without significantly complicating this PR or its review.

Most information about these changes is in the two most important commits, [c66257e](https://github.com/gitpython-developers/GitPython/pull/1868/commits/c66257eac7f344814d95bc45243f30659d767350) and [91f967a](https://github.com/gitpython-developers/GitPython/pull/1868/commits/91f967a9bb579050f2450ce59f657e225101b05a). This includes some information about why I believe `--fix` should be kept.

- I've also included information in [91f967a](https://github.com/gitpython-developers/GitPython/pull/1868/commits/91f967a9bb579050f2450ce59f657e225101b05a) about what it does *not* do and why it does not do it, which may be considered excessive in a commit message, and I'd be pleased to amend that and move that information into this PR description on request.
- The reason I've included that information is so that no matter what order things happen in, and even if some are omitted, and regardless of what related changes do or don't happen in the future, and even if I am making a mistake in thinking that it is better to have this additional PR rather than delay and expand the scope of [#1865](https://github.com/gitpython-developers/GitPython/pull/1865), the full context will still be readily available to anyone who looks at that commit while trying to figure something out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ probably the skills to scratch that itch of mine: implement `git` in a way that 
 If you like the idea and want to learn more, please head over to [gitoxide](https://github.com/Byron/gitoxide), an
 implementation of 'git' in [Rust](https://www.rust-lang.org).
 
-*(Please note that `gitoxide` is not currently available for use in Python, and that Rust is required)*
+*(Please note that `gitoxide` is not currently available for use in Python, and that Rust is required.)*
 
 ## GitPython
 
@@ -39,9 +39,9 @@ The project is open to contributions of all kinds, as well as new maintainers.
 
 ### REQUIREMENTS
 
-GitPython needs the `git` executable to be installed on the system and available in your `PATH` for most operations.
-If it is not in your `PATH`, you can help GitPython find it by setting
-the `GIT_PYTHON_GIT_EXECUTABLE=<path/to/git>` environment variable.
+GitPython needs the `git` executable to be installed on the system and available in your
+`PATH` for most operations. If it is not in your `PATH`, you can help GitPython find it
+by setting the `GIT_PYTHON_GIT_EXECUTABLE=<path/to/git>` environment variable.
 
 - Git (1.7.x or newer)
 - Python >= 3.7
@@ -57,7 +57,7 @@ GitPython and its required package dependencies can be installed in any of the f
 
 To obtain and install a copy [from PyPI](https://pypi.org/project/GitPython/), run:
 
-```bash
+```sh
 pip install GitPython
 ```
 
@@ -67,7 +67,7 @@ pip install GitPython
 
 If you have downloaded the source code, run this from inside the unpacked `GitPython` directory:
 
-```bash
+```sh
 pip install .
 ```
 
@@ -75,7 +75,7 @@ pip install .
 
 To clone the [the GitHub repository](https://github.com/gitpython-developers/GitPython) from source to work on the code, you can do it like so:
 
-```bash
+```sh
 git clone https://github.com/gitpython-developers/GitPython
 cd GitPython
 ./init-tests-after-clone.sh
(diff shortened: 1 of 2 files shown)


EliahKagan:
Due to the small `README.md` changes in [1b8812a](https://github.com/gitpython-developers/GitPython/pull/1865/commits/1b8812a968a76f47acdf80bc1d15e6b0dd5d84bf) in #1865, this PR may end up having a trivial conflict in `README.md` with changes in [c66257e](https://github.com/gitpython-developers/GitPython/pull/1868/commits/c66257eac7f344814d95bc45243f30659d767350). Although that does not supersede this PR or the other readme changes [c66257e](https://github.com/gitpython-developers/GitPython/pull/1868/commits/c66257eac7f344814d95bc45243f30659d767350) makes, I suggest #1865 simply be merged, after which I can rebase this assuming there is a conflict.

EliahKagan:
I've rebased this onto main to get the changes from #1865, and reworded commit messages accordingly, including removing some of the less important text in the last commit message (the situation is simpler now that #1865 is removed, especially since it ended up including the removal of the `lint` target in `Makefile`, so that extra information is out of date as well as less important).

Byron (approved):
Thanks a lot for these updates and the incredible reasoning and documentation that comes with them.

If all of GitPython would have been done like these, many things would be clearer today I am sure. Thus, I don't think any commit message should be changed or truncated, and you are always welcome to leave as much information as you like and where you like.

PR #1871, opened 2024-03-13
Slightly broaden Ruff, and update and clarify tool configuration

EliahKagan:
Changes:

- Lets Ruff scan `doc/source/conf.py`, and other Python modules in `doc` outside `build` if there were any (00ff7e3).
- Adds comments on all suppressed Ruff rules and puts all of `pyproject.toml` in closer to the same style (4814775).
- Adds Ruff (and ShellCheck, which could've been added before) to the file currently used to list tools not listed as dependencies (a8a73ff).
- Bumps pre-commit hook versions, which brings small benefits for Ruff and ShellCheck (ff1ebf8).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,6 @@
 repos:
-
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.3.0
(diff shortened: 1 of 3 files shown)


Byron (approved):

Issue #1869, opened 2024-03-12
Submodule.__init__ `parent_commit` conversion/validation is implied but not done


EliahKagan:
The `parent_commit` parameter of `Submodule.__init__` is documented this way:

https://github.com/gitpython-developers/GitPython/blob/edb8d26e20d16f72240f7040c353d01a3a1a9740/git/objects/submodule/base.py#L135-L136

`Submodule.__init__` binds this directly to the private `_parent_commit` attribute:

https://github.com/gitpython-developers/GitPython/blob/edb8d26e20d16f72240f7040c353d01a3a1a9740/git/objects/submodule/base.py#L147

But this is at odds with the documented relationship to `Submodule.set_parent_commit`. That method's `commit` parameter corresponds to the `parent_commit` parameter of `Submodule.__init__`, in that its `commit` parameter is used to identify the commit, if any, to set as `_parent_commit`. However, `set_parent_commit` performs both conversion and validation.

This is the relevant fragment of `set_parent_commit`'s docstring:

https://github.com/gitpython-developers/GitPython/blob/edb8d26e20d16f72240f7040c353d01a3a1a9740/git/objects/submodule/base.py#L1253-L1255

When `commit` is `None`, it sets `None` to `_parent_commit`. Otherwise, however, `commit` may not already be a `Commit` object, and that is okay, because a commit is looked up from it:

https://github.com/gitpython-developers/GitPython/blob/edb8d26e20d16f72240f7040c353d01a3a1a9740/git/objects/submodule/base.py#L1274

That's the conversion. Then validation is performed, with `_parent_commit` ending up set to the commit that `commit` identified only if there is such a suitable commit:

https://github.com/gitpython-developers/GitPython/blob/edb8d26e20d16f72240f7040c353d01a3a1a9740/git/objects/submodule/base.py#L1275-L1289

The type annotations do not reveal the intent, as they are among those using `Commit_ish` that need to be updated with the fix for `Commit_ish`, and that I am fixing up in #1859. My immediate motivation for opening this issue is that I'm having trouble figuring out how to annotate them, because due to the inconsistency between the docstring and the implementations, I don't know what is intended to be accepted.

[...]

PR #1877, opened 2024-03-14
Describe `Submodule.__init__` parent_commit parameter

EliahKagan:
This includes a brief description of the `Submodule.__init__` *parent_commit* parameter in its docstring, rather than only referring to the `set_parent_commit` method, whose semantics differ due to conversation and validation, and which accepts more types than just `Commit` or `None`.

The wording is based on wording in `set_parent_commit`, adjusted for the difference in types, and `set_parent_commit` remains reference for further details.

This builds on 1f03e7f (#1859) in improving the situation described in #1869.


Byron (approved):

Issue #1716, opened 2023-10-18
3.1.38 meant to fix #1656 but there is no supporting gitdb release


EliahKagan:
To support #1659 here in GitPython, gitdb received [#98](https://github.com/gitpython-developers/gitdb/pull/98), and the GitPython repository's submodule reference to gitdb was updated in #1704 (and #1705). Thus #1656 *might* be<sup>**1**</sup> fixed when developing against the cloned GitPython repository with its gitdb submodule.

[With the intention](https://github.com/gitpython-developers/GitPython/pull/1659#issuecomment-1765731586) of fixing #1656 in the far more common case of using GitPython via PyPI packages, GitPython 3.1.38 was released. But no new release of gitdb was ever made. Since the fix depends on changes to gitdb, it seems unlikely that 3.1.38 really fixes #1656.

I think it might be possible to make further refinements to `__alL__` and surrounding code, slightly enhancing #1659. However, since 3.1.38 was already released, it seems to me that such things can wait, and that it's reasonable to make a gitdb release (bumping the version in its [`setup.py`](https://github.com/gitpython-developers/gitdb/blob/6a2270635842c4287ae14d94fdc3f9e1ddd0527a/setup.py#L10) and [`gitdb/__init__.py`](https://github.com/gitpython-developers/gitdb/blob/6a2270635842c4287ae14d94fdc3f9e1ddd0527a/gitdb/__init__.py#L31), or changing the [current approach](https://github.com/gitpython-developers/gitdb/blob/6a2270635842c4287ae14d94fdc3f9e1ddd0527a/setup.py#L3-L5) and bumping it in just one place), and to include a dependency bump requiring it the next GitPython release (which could be the same release that ships #1715, if the timing works out).

I'd be pleased to open pull requests to help with this issue, either following this plan or another one, if that would be helpful.

---

<sup>**1**</sup> I had originally said it was likely fixed when developing against the cloned GitPython. However, that is probably not the case. First, most type checkers probably wouldn't understand how that version is added to `sys.path`, and it is probably good design of them not to (since they are *static* checkers). Second, as documented in #1717, it turns out this is never actually happening.

Byron:
Thanks for bringing this to my attention!

> I'd be pleased to open pull requests to help with this issue, either following this plan or another one, if that would be helpful.

Yes, I'd very much like your help with this, thank you 🙏.
[...]

PR #1880, opened 2024-03-19
Replace all wildcard imports with explicit imports

### LINKED ISSUE #1349: mypy is failing on `from git import Blob`

EliahKagan:
Fixes #1349

This actually makes a number of related changes, several of which the reduction in mypy errors in #1859 helped in validating:

- **All wildcard imports (`from blah import *`) are replaced with imports of specific names.** This is the key change, with everything else done to support it, or directly facilitated by it, or conceptually related to it, or practically easy to do alongside it.

  Wildcard imports are inherently brittle, discouraged in PEP-8, and have already led to [some difficulties](https://github.com/gitpython-developers/GitPython/blob/0a609b94ffdde6c7cccc5d6409677c4e6c914bb8/test/test_util.py#L140-L142) for GitPython (see fc86a23 and the "Another ambiguity, possibly relevant" section of [#1802](https://github.com/gitpython-developers/GitPython/issues/1802)). Even more importantly, this makes things easier for type checkers, especially for code that uses GitPython from the outside where common and useful configurations otherwise report errors due to the wildcards' lack of explicitness about what names they introduce.

  Furthermore, the combination of the mypy error reduction in [#1859](https://github.com/gitpython-developers/GitPython/pull/1859) with the elimination of wildcard imports makes it feasible to verify that newly introduced dynamic behavior is done in a way that avoids breaking static type checking for users of GitPython. This allows deprecation warnings on access to specific module attributes, as may be useful [in `git.compat`](https://github.com/gitpython-developers/GitPython/pull/1732), as well as [a dynamically generated `git.__version__` attribute](https://github.com/gitpython-developers/GitPython/issues/1716#issuecomment-1773746757), to be safely approachable. (But such now-doable enhancements are not part of this pull request.)

  I wrote a short script (1e5a944, f705fd6) to verify that [changes to module contents](https://github.com/gitpython-developers/GitPython/blob/ecdb6aa25a95e2cb3a650a26f3ffd9c45e127ed1/ab.diff) were minimal and as expected, including verifying what object attributes reference, not just what the attributes' names were.

- Top-level names that are imported only to match what was imported before, but that are nonpublic and inadvisable to use, are commented as such where they are imported. Most of these could possibly be removed very soon.

- The test suite has a few new tests, to test a particularly tricky aspect of the observable behavior of imports that is easy to change accidentally but should only ever be changed intentionally (the `git.util` vs. `git.objects.util` issue).

- `__all__` is added where absent but clearly useful, which is everywhere under `git` where it was absent except `git.types`. This was only a handful of places.

- `git.objects.__all__`, which was dynamically generated as a list comprehension--as `git.__all__` was prior to  [#1659](https://github.com/gitpython-developers/GitPython/pull/1659)--is likewise replaced with a literal `__all__`. As detailed in the f89d065 commit message, I used `deltall` (see also [#1875](https://github.com/gitpython-developers/GitPython/issues/1875)) to verify that this was the right course.

- The `__init__` for `git.objects.submodule` no longer patches `IndexObject` and `Object` into `git.objects.submodule.util`. This had been done to avoid a circular import error, but at some point ceased to be needed. Because `git.objects.submodule.util` has defined `__all__` for a long time, which omits them (and seems always to have omitted them), and because it also does not use them, I didn't replace that patching with any other way of getting them in there. However, it looks like it would be feasible to import them in a way that would not cause problems, if that is called for.

- Also related to that situation, `git.objects.submodule` had been the only subpackage whose `__init__` didn't make names from its own submodules available directly in it. Because neither the circular import error nor the dynamic patching workaround are being done, there is no impediment to this, which improves consistency across the project and seemed to me to be what users would expect, so I included that.

- `__all__` is now placed at the PEP-8 recommended location near the top of each module, above imports except in rare cases where that is infeasible.

- A number of linter suppressions are removed, most of which are no longer needed due to wildcard imports no longer being used. Some others are made more specific, including by splitting them into suppressions that apply to specific names in a `from` import. Some of what facilitates this is actually the switch from flake8 to Ruff in [#1862](https://github.com/gitpython-developers/GitPython/pull/1862).

- Imports across the project are fully sorted and grouped, in a consistent style that I think reflects the sensibility of the style that has been in place for an extended time. This is something that I had done bits of as part of other pull requests, and it was convenient to "complete" it here. I use scare quotes because it is likely I have missed some things. Ideally this should eventually be checked--or, better, applied--by an automated tool. Ruff can sort imports but I don't think it supports anything like this style, while isort is more configurable but might still be difficult to get to do it. However, the goal here is not this specific style, but instead to have something as consistent and helpful as possible if the style is not changed or before it is changed, while also making it easier to identify the advantages and disadvantages of an automatically applicable style over the best the current style has to offer.

- Minor cleanup alongside these changes, including removal of old commented-out code (done in its own commits so it's easy to find in case that code is wanted again) and rewording the `git.objects.util` module docstring to distinguish that module from `git.util`. The latter is the reason for [that `__doc__` attribute change](https://github.com/gitpython-developers/GitPython/blob/ecdb6aa25a95e2cb3a650a26f3ffd9c45e127ed1/ab.diff#L38-L39).

A few special considerations:

- It [was suggested](https://github.com/gitpython-developers/GitPython/issues/1349#issuecomment-1472595657) that [#1349](https://github.com/gitpython-developers/GitPython/issues/1349) could be fixed by writing `git.__all__` literally rather than dynamically generating it. But that improvement was since made in [#1659](https://github.com/gitpython-developers/GitPython/pull/1659), and while [#1656](https://github.com/gitpython-developers/GitPython/issues/1656) was fixed, [#1349](https://github.com/gitpython-developers/GitPython/issues/1349) remained. I've re-verified that this is the case by creating a repository that uses GitPython in a way that more closely approximates installing it from PyPI than is common during development, and ensures that the non-default `mypy` configuration in `pyproject.toml` is not used, while still allowing arbitrary files to be tested, by doing a non-editable installation into a different directory with a separate virtual environment. [More details are in this gist.](https://gist.github.com/EliahKagan/271598adb9f83683790eb437d03efb66) This confirmed that, as of the main branch at 0a609b9, [#1349](https://github.com/gitpython-developers/GitPython/issues/1349) was reproducible, while at the tip of this feature branch, d524c76, it was resolved.

- I am not sure what the best choice is for whether intermediate subpackages--Python modules like `git.objects` and `git.refs` that are under the top-level `git` package but contain their own Python submodules--should include their Python submodules' names in `__all__`. In GitPython (though not all projects), everything is imported as a consequence of importing the top-level `git` module, so these are already available, and they are also already public, in the sense that they are available due to having been imported, with the fully qualified names they are guaranteed to have as a consequence of being imported. In most but not all cases--`git.objects` being a major counterexample due to its previously dynamically built `__all__` that omitted modules--these were already importable in wildcard imports due to the absence of `__all__`.

  However, that this was the case had caused some problems, including several modules that are not direct children of the top-level `git` module ending up as attributes of it (most of which are simply private attributes that should not be used, but one of which, being `git.index.util`, is the cause of the `git.util`/`git.index.util` ambiguity). Something different could be done in `git.objects` than elsewhere for maximum similarity to the previous state of affairs, but I decided instead to omit them everywhere, based on the top-level `git` module not listing its own direct Python submodules in its `__all__`, and the possibility that omitting them would help mitigate the risks of wildcard imports outside of GitPython.

- Some modules have unused imports not listed in `__all__`, present intentionally for use in other GitPython modules but not public outside the project. As [commented on #1854](https://github.com/gitpython-developers/GitPython/issues/1854#issuecomment-1977531932), I'm unsure whether or how that should ultimately be improved, but I definitely consider it beyond the scope of this pull request. As such, `implicit_reexport = true` remains set for `mypy` in `pyproject.toml` (`mypy` otherwise warns about this situation wherever other modules use such imports).


EliahKagan:
It looks like https://github.com/gitpython-developers/GitPython/actions/runs/8339331003/job/22821123508, in [5b2771d](https://github.com/gitpython-developers/GitPython/pull/1880/commits/5b2771d23af2156fb7e12ee6406bffbabcb9e95d), is due to a network error and would go away if run again.

In contrast, the subsequent commit, [fc86a23](https://github.com/gitpython-developers/GitPython/pull/1880/commits/fc86a238f99a6406ed949a7d690ec0f6be2f31e0), is expected to having a failing status (which should go green again starting in the commit after that).

test/test_imports.py
```diff
@@ -0,0 +1,32 @@
+# This module is part of GitPython and is released under the
+# 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
+
+import sys
+
+import git
+
+
+def test_git_util_attribute_is_git_index_util():
```
Byron, line 9: It's great to have tests for this, good idea!

Byron (approved):
Thanks so much for fixing these imports! It sounds like doing so is going to have major benefits, despite being a conceptually simple change.

> * I wrote a short script ([1e5a944](https://github.com/gitpython-developers/GitPython/commit/1e5a9449573bfe70987dcaf44e120895131e9c66), [f705fd6](https://github.com/gitpython-developers/GitPython/commit/f705fd6dde6047c7787809aeb691988fa4af80d8)) to verify that [changes to module contents](https://github.com/gitpython-developers/GitPython/blob/ecdb6aa25a95e2cb3a650a26f3ffd9c45e127ed1/ab.diff) were minimal and as expected, including verifying what object attributes reference, not just what the attributes' names were.

Incredibly diligent, thanks so much.

I will merge this PR once CI has caught up and is green.

PR #1881, opened 2024-03-19
Clarify how tag objects are usually tree-ish and commit-ish

EliahKagan:
This revises the docstrings of the `Tree_ish` and `Commit_ish` unions, to make clearer that tags objects are usually used to identify commits and are thus usually tree-ish and commit-ish.

This change relates to the discussion in #1878, starting at https://github.com/gitpython-developers/GitPython/pull/1878#pullrequestreview-1938343498.

### DIFF
--- a/git/types.py
+++ b/git/types.py
@@ -75,7 +75,7 @@ For more general information on git objects and their types as git understands t
 """
 
 Tree_ish = Union["Commit", "Tree", "TagObject"]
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot!

Now that I am reading this, I also understand where some of the confusion came from. Somehow, I wasn't truly aware that the overall context was `Tree_ish` and `Commit_ish`, and instead was hyper-focussed on "`TagObject`  is typically…", not "sometimes".

Issue #1891, opened 2024-03-31
Detailed xfail tracebacks sometimes distract from unexpected failures


EliahKagan:
This issue is to track the problem noticed in #1853 and #1862 (see comments in both) where the specific details of new test failures are made less obvious in reading `pytest` output, in CI checks, due to the otherwise useful pytest 8 feature to show more such details.

From https://github.com/gitpython-developers/GitPython/pull/1853#issuecomment-1974005684:

> `pytest` output [now contains details](https://github.com/pytest-dev/pytest/issues/11233) of tests with xfail (expected failure) status, which are unrelated, and that may have distracted from the related test's failure. This [changed in major version 8 of `pytest`](https://docs.pytest.org/en/8.0.x/changelog.html#pytest-8-0-0rc2-2024-01-17) (which GitPython is using automatically except on Python 3.7 which doesn't support it). Overall, for most projects, it seems to me that this is an improvement, since it's useful to be able to see the specific details of xfailing tests without extra effort (and check if they really match up with the messages written in the xfail decorators). However, it may be that for GitPython's tests it makes common tasks more cumbersome. If so, then the `pytest` configuration can be changed (this does not require downgrading `pytest`).

I'm unsure what the best solution is, but if necessary the `-ra` in

https://github.com/gitpython-developers/GitPython/blob/53640535cf8314366a01da081947dd8504a299cd/pyproject.toml#L6

can be changed to something that does not imply `x` (xfail). The "full list of available characters" in the [Producing a detailed summary report](https://docs.pytest.org/en/8.0.x/how-to/output.html#producing-a-detailed-summary-report) section of the documentation. I'm unsure of what should be done and I don't know of an easy way to eliminate the distracting effects while still having the benefit of seeing specific output of xfailing tests.

PR #1892, opened 2024-03-31
Don't suppress pytest warning summaries

EliahKagan:
This removes the `--disable-warnings` option from those specified in `pyproject.toml` for `pytest` to use automatically.

I originally tried this in the hope that it would show a warning for misspelled keyword arguments passed when constructing pytest marks (which should not be confused with misspelling the mark itself, which is something pytest can warn for). That bug (#1893) was discovered in #1888. *This turns out not to be useful for that at all*, so I'm opening this PR separate from anything related to that.

Nonetheless, there doesn't seem to be a significant disadvantage to showing this output, which is not easy to mistake for errors or test failures, and which might potentially help catch other kind of cases where a misspelling that intuitively feels like it would lead to an error instead produces a major change in semantics for the tests, in addition to other problems.

There are also some resource-related warnings that may help point the way to debugging things, especially in future examinations of how GitPython interacts with garbage collection on Windows.

Therefore, I think it makes sense to allow these warnings summaries to be shown.

I believe this does not worsen #1891, because these do not look like errors. I suspect it might very slightly mitigate it in some cases, even, since the warnings section near the xfailures section makes it easier to notice that these sections are not just for reporting errors and actual test failures. However, I definitely do not advocate this on the basis of that possibility; in my opinion, allowing the warning summaries to be shown is justified in spite of how it increases the overall length and number of sections of the test output, not *because* of that.

Deprecation warnings are separately suppressed and this does not change that, though perhaps it ought to, since eventually all areas in the tests that allow a deprecation warning to emanate from them should probably be marked for that specifically. This is also separate from how the new tests in `test/deprecation` (added in #1886) that check for the presence or absence of deprecation warnings work: those tests will produce hard failures if their expectations are not met, regardless of the changes here or to whether deprecation warnings elsewhere are allowed into the warnings summaries.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
-addopts = "--cov=git --cov-report=term --disable-warnings -ra"
+addopts = "--cov=git --cov-report=term -ra"
 filterwarnings = "ignore::DeprecationWarning"
 python_files = "test_*.py"
 tmp_path_retention_policy = "failed"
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! Anything on the way to improving on #1891 is appreciated.

PR #1894, opened 2024-04-01
Update the comment about `--mixed` and paths

EliahKagan:
This updates the comment in `HEAD.reset` about why `--mixed` is omitted from the git command constructed to perform a reset where paths are being passed, adding specific information about the git versions where this is deprecated, and changing the more-info link from an old GitPython issue that is no longer retrievable to #1876.

See #1876 for details.


Byron (approved):
Thanks a lot!

PR #1899, opened 2024-04-05
Fixed an error updating shallow submodules

EduardTalanov:
When we try to execute 
submodule_update(recursive=True, init=True) 
in Repo class with shallow submodules, command
git fetch --progress -v origin 
sometimes print log such as
=5 = [up to date]      zm/160_uwp              -> origin/zm/160_uwp
but it doesn't satisfy filter:
r"^ *(.) (\[[\w \.$@]+\]|[\w\.$@]+) +(.+) -> ([^ ]+)( \(.*\)?$)?"
and throw exception.
 I've updated it so that it works correctly.

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -316,7 +316,7 @@ class FetchInfo(IterableObj):
         ERROR,
     ) = [1 << x for x in range(8)]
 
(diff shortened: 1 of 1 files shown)


Byron:
Do these numbers in `=5 = [up to date] zm/160_uwp -> origin/zm/160_uwp` (`=5 =`) look like an issue with parallelism on the server side, or do they have meaning?

EduardTalanov:
I don't think so. I've found this number only in one branch. Other branches didn't have this number.

EduardTalanov:
@dependabot recreate

Issue #1905, opened 2024-04-22
Fuzz Tests Are Crashing at Start-up on ClusterFuzz


DaveLak:
PR #1901 was successful in fixing the broken build but resurfaced another issue that is preventing the fuzzer from running.

## The Problem

It appears that a Git executable is not available in the ClusterFuzz container environment where fuzz tests are executed, causing an error in the fuzz harnesses when GitPython attempts to initialize.

 This issue has been previously seen and reported on the OSS-Fuzz issue tracker: https://github.com/google/oss-fuzz/issues/10600

<details><summary>Relevant Portion of the ClusterFuzz Crash Logs from 2024-04-20</summary>

```shell
# <Prior output omitted for brevity>
INFO: Instrumenting git.index.util
INFO: Instrumenting git.remote
INFO: Instrumenting [git.repo.fun](http://git.repo.fun/)
Traceback (most recent call last):
  File "git/__init__.py", line 296, in <module>
  File "git/__init__.py", line 287, in refresh
  File "git/cmd.py", line 631, in refresh
ImportError: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh(<full-path-to-git-executable>)

All git commands will error until this is rectified.

This initial message can be silenced or aggravated in the future by setting the
$GIT_PYTHON_REFRESH environment variable. Use one of the following values:
    - quiet|q|silence|s|silent|none|n|0: for no message or exception
    - warn|w|warning|log|l|1: for a warning message (logging level CRITICAL, displayed by default)
    - error|e|exception|raise|r|2: for a raised exception

Example:
    export GIT_PYTHON_REFRESH=quiet


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "[fuzz_config.py](http://fuzz_config.py/)", line 26, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 419, in exec_module
  File "git/__init__.py", line 298, in <module>
ImportError: Failed to initialize: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh(<full-path-to-git-executable>)

All git commands will error until this is rectified.

This initial message can be silenced or aggravated in the future by setting the
[...]

PR #1906, opened 2024-04-22
Fix Fuzzer Crash in ClusterFuzz Due to Missing Git Executable

DaveLak:
A Git executable is not globally available in the ClusterFuzz container
environment where OSS-Fuzz executes fuzz tests, causing an error in the fuzz
harnesses when GitPython attempts to initialize, crashing the tests before they
can run.

To avoid this issue, we bundle the `git` binary that is available in the OSS-Fuzz
build container with the fuzz harness via Pyinstaller's `--add-binary` flag in
`build.sh` and use GitPython's `git.refresh(<full-path-to-git-executable>)`
method inside a Pyinstaller runtime check to initialize GitPython with the
bundled Git executable when running from the bundled application.

In all other execution environments, we assume a `git` executable is available
globally.

Fixes:
- https://github.com/gitpython-developers/GitPython/issues/1905
- https://github.com/google/oss-fuzz/issues/10600

### DIFF
--- a/fuzzing/fuzz-targets/fuzz_config.py
+++ b/fuzzing/fuzz-targets/fuzz_config.py
@@ -20,16 +20,21 @@
 import atheris
 import sys
 import io
+import os
 from configparser import MissingSectionHeaderError, ParsingError
 
 with atheris.instrument_imports():
-    from git import GitConfigParser
+    import git
 
 
 def TestOneInput(data):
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for the fix!

I wonder how this could ever work on OSS-fuzz.

DaveLak:
> I wonder how this could ever work on OSS-fuzz.

Different images everywhere 🫣

Issue #1887, opened 2024-03-30
Unhandled IndexError when calling .read() on a malformed config file


DaveLak:
Hi, I noticed the fuzzing tests that OSS-Fuzz runs on this project are broken and while I was working on fixing them I believe I came across a minor bug:

## The Bug

An `Uncaught Python exception:` `IndexError: string index out of range` can be triggered if when trying to call `.read()` on `GitConfigParser` if it was initialized with a malformed config file.

### Current Behavior

It's easiest to demonstrate, so please consider this example:

```python
import io
from git.config import GitConfigParser


def reproduce_issue():

    malformed_config_content_bytestring = b'[-]\nk:"v\n"'

    problematic_config_file = io.BytesIO(malformed_config_content_bytestring)

    # problematic_config_file looks now like this:
    """
    [-]
    k:"v
    "
    """

    # We have to name the file otherwise we'll trigger
    # `AttributeError: '_io.BytesIO' object has no attribute 'name'` here:
    # https://github.com/gitpython-developers/GitPython/blob/c7675d2cedcd737f20359a4a786e213510452413/git/config.py#L623
    problematic_config_file.name = "fuzzedconfig.config"

    # This is fine
    git_config = GitConfigParser(problematic_config_file)
    
    # The next line raised an unhandled `IndexError: string index out of range`
    git_config.read()


if __name__ == "__main__":
    reproduce_issue()

```

Assuming that code is in `/path/to/example/config_indexerror_reproduction.py` then
```sh
python config_indexerror_reproduction.py
```
produces something akin to:

```sh
Traceback (most recent call last):
  File "/path/to/example/config_indexerror_reproduction.py", line 30, in <module>
    reproduce_issue()
  File "/path/to/example/config_indexerror_reproduction.py", line 26, in reproduce_issue
    git_config.read()
  File "/path/to/example/.venv/lib/python3.12/site-packages/git/config.py", line 607, in read
    self._read(file_path, file_path.name)
  File "/path/to/example/.venv/lib/python3.12/site-packages/git/config.py", line 514, in _read
    cursect.setlast(optname, optval + string_decode(line))
                                      ^^^^^^^^^^^^^^^^^^^
  File "/path/to/example/.venv/lib/python3.12/site-packages/git/config.py", line 441, in string_decode
    if v[-1] == "\\":
       ~^^^^
IndexError: string index out of range
```

<details><summary>My Reproduction Environment Details</summary>

The reproduction code above was tested on:

```sh
[...]

PR #1908, opened 2024-04-22
Fix `IndexError` in `GitConfigParser` When a Quoted Config Value Contains a Trailing New Line

DaveLak:
Fixes: https://github.com/gitpython-developers/GitPython/issues/1887

Improve the guarding `if` check in `GitConfigParser`'s `string_decode` function to safely handle empty strings and prevent `IndexError`s when accessing string elements.

This resolves an IndexError in the `GitConfigParser`'s `.read()` method when the config file contains a quoted value ending with a new line.

### DIFF
--- a/fuzzing/fuzz-targets/fuzz_config.py
+++ b/fuzzing/fuzz-targets/fuzz_config.py
@@ -35,7 +35,7 @@ def TestOneInput(data):
     except (MissingSectionHeaderError, ParsingError, UnicodeDecodeError):
         return -1  # Reject inputs raising expected exceptions
     except ValueError as e:
-        if isinstance(e, ValueError) and "embedded null byte" in str(e):
+        if "embedded null byte" in str(e):
             # The `os.path.expanduser` function, which does not accept strings
             # containing null bytes might raise this.
             return -1
--- a/fuzzing/fuzz-targets/fuzz_config.py
+++ b/fuzzing/fuzz-targets/fuzz_config.py
@@ -34,12 +34,8 @@ def TestOneInput(data):
         git_config.read()
     except (MissingSectionHeaderError, ParsingError, UnicodeDecodeError):
         return -1  # Reject inputs raising expected exceptions
-    except (IndexError, ValueError) as e:
-        if isinstance(e, IndexError) and "string index out of range" in str(e):
-            # Known possibility that might be patched
-            # See: https://github.com/gitpython-developers/GitPython/issues/1887
-            pass
-        elif isinstance(e, ValueError) and "embedded null byte" in str(e):
+    except ValueError as e:
+        if isinstance(e, ValueError) and "embedded null byte" in str(e):
             # The `os.path.expanduser` function, which does not accept strings
             # containing null bytes might raise this.
             return -1
--- a/git/config.py
+++ b/git/config.py
@@ -452,7 +452,7 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
         e = None  # None, or an exception.
 
         def string_decode(v: str) -> str:
-            if v and v[-1] == "\\":
+            if v and v.endswith("\\"):
                 v = v[:-1]
             # END cut trailing escapes to prevent decode error
 
--- a/git/config.py
+++ b/git/config.py
@@ -452,7 +452,7 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
         e = None  # None, or an exception.
 
         def string_decode(v: str) -> str:
-            if v[-1] == "\\":
+            if v and v[-1] == "\\":
                 v = v[:-1]
             # END cut trailing escapes to prevent decode error
 
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -142,6 +142,14 @@ class TestBase(TestCase):
             )
             self.assertEqual(len(config.sections()), 23)
 
+    def test_config_value_with_trailing_new_line(self):
+        config_content = b'[section-header]\nkey:"value\n"'
+        config_file = io.BytesIO(config_content)
+        config_file.name = "multiline_value.config"
+
+        git_config = GitConfigParser(config_file)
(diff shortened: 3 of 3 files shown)


fuzzing/fuzz-targets/fuzz_config.py
```diff
@@ -34,12 +34,8 @@ def TestOneInput(data):
...
     except (MissingSectionHeaderError, ParsingError, UnicodeDecodeError):
         return -1  # Reject inputs raising expected exceptions
-    except (IndexError, ValueError) as e:
-        if isinstance(e, IndexError) and "string index out of range" in str(e):
-            # Known possibility that might be patched
-            # See: https://github.com/gitpython-developers/GitPython/issues/1887
-            pass
-        elif isinstance(e, ValueError) and "embedded null byte" in str(e):
+    except ValueError as e:
+        if isinstance(e, ValueError) and "embedded null byte" in str(e):
```
DaveLak, line 38: The `isinstance` check is not strictly necessary, but I think it reads better with it, so I left it. I'm happy to remove it if that's preferred.

Byron: When I read it, I wonder why it's necessary and can't figure it out.
And since it's not necessary, I think it would be better to remove the `isinstance` check.

Byron:
```suggestion
        if "embedded null byte" in str(e):
```

Byron (approved):

I will merge right after CI is green after applying my tiny patch.

git/config.py
```diff
@@ -452,7 +452,7 @@ def _read(self, fp: Union[BufferedReader, IO[bytes]], fpname: str) -> None:
         e = None  # None, or an exception.
 
         def string_decode(v: str) -> str:
-            if v[-1] == "\\":
+            if v and v[-1] == "\\":
```
EliahKagan, line 455: I think this can be made a bit clearer (and more idiomatic) by doing:

```suggestion
            if v.endswith("\\"):
```

After all, we want to know if it ends with a backslash, which `endswith` safely and naturally expresses.

EliahKagan (commented):
I've suggested a small change that I think improves clarity.

Separately from the changes here but relevant to the logic they are a part of, I'm a bit worried about the case where there are *two* (or more) trailing backslashes, so that removing one of them still leaves a trailing backslash. However, that was present before, and this PR definitely need not be expanded to address it! I'm also not certain I'm right to be worried about that; maybe it is less important to cover, or a part of a separate class of malformed configuration file inputs that are intended to trigger decoding errors.

git/config.py
Byron: I also think that's better, however, `v` seems to be nullable which isn't checked for anymore. I will apply the patch with that change, just to be sure.

Byron:
```suggestion
            if v and v.endswith("\\"):
```

Byron:
> I've suggested a small change that I think improves clarity.
> 
> Separately from the changes here but relevant to the logic they are a part of, I'm a bit worried about the case where there are _two_ (or more) trailing backslashes, so that removing one of them still leaves a trailing backslash. However, that was present before, and this PR definitely need not be expanded to address it! I'm also not certain I'm right to be worried about that; maybe it is less important to cover, or a part of a separate class of malformed configuration file inputs that are intended to trigger decoding errors.

Thanks for taking another look - admittedly I forgot to merge this PR. Cygwin is the bane of this CI as it takes unnatural amounts of time :/.

Having worked on the `gitoxide` version of this I can say that this implementation here is more hole than cheese (if that makes sense :)), but it seems to work well enough for the common cases, still. Since GitPython is in maintenance mode, I don't think there is any need to try to improve it beyond fixing bugs.

git/config.py
```diff
@@ -452,7 +452,7 @@ def _read(self, fp: Union[BufferedReader, IO[bytes]], fpname: str) -> None:
         e = None  # None, or an exception.
 
         def string_decode(v: str) -> str:
-            if v[-1] == "\\":
+            if v and v.endswith("\\"):
```
EliahKagan, line 455: Is there a reason to separately check if `v` is truthy? In my mind, the reason to call `endswith` is that it takes care of the entire check in a single simple expression, regardless of whether the string was empty (falsy) or nonempty (truthy).

```suggestion
            if v.endswith("\\"):
```

Unless the parameter type annotation is incorrect and `v` could be some non-string such as the `None` object, I recommend against separately checking if `v` is truthy. The main benefit of calling `endswith` is that it does the full check regardless of whether the string is nonempty (truthy), making a more complex expression, such as one with the `and` operator, unnecessary.

The only way it would be necessary to check if `v` is truthy before calling `endswith` is if the parameter type annotation is incorrect and `v` is not always a string. If that is the case, then it would be a bigger issue (that's the main reason I'm posting this additional comment, even after the PR is merged).

EliahKagan: Sorry, for some reason I didn't see [that comment](https://github.com/gitpython-developers/GitPython/pull/1908#discussion_r1580488620)! That addresses this proactively and my above comment can be disregarded entirely.

If the code is modified further in the future then the `str` annotation, which is inconsistent with a `None` value, could be examined and potentially changed.

git/config.py
EliahKagan: For some reason, I didn't see, or carelessly glossed over, [that comment](https://github.com/gitpython-developers/GitPython/pull/1908#discussion_r1580488620), and this posted [this unnecessary comment](https://github.com/gitpython-developers/GitPython/pull/1908#discussion_r1580500339) that can be disregarded in its entirety. Unfortunately, GitHub doesn't allow me to mark that review thread as resolved, and I don't want to delete it lest that cause further confusion if it has been seen. My apologies to you both!

git/config.py
Byron: By the way, I tried to enable branch protections which should enable auto-merge, and thus resolve the issue with long-running CI sometimes preventing a merge in addition to me forgetting to do it later. It didn't yet work for this PR, but I will see for the next one.

PR #1910, opened 2024-04-29
Replace the Suboptimal `fuzz_tree.py` Harness With a Better Alternative

DaveLak:
As discussed in the initial fuzzing integration PR[^1], `fuzz_tree.py`'s implementation was not ideal in terms of coverage and its reading/writing to hard-coded paths inside `/tmp` was problematic as (among other concerns), it causes intermittent crashes on ClusterFuzz[^2] when multiple workers execute the test at the same time on the same machine.

The changes here replace `fuzz_tree.py` completely with a completely new `fuzz_repo.py` fuzz target which:

- Uses `tempfile.TemporaryDirectory()` to safely manage tmpdir creation and tear down, including during multi-worker execution runs.
- Retains the same feature coverage as `fuzz_tree.py`, but it also adds considerably more from much smaller data inputs and with less memory consumed (and it doesn't even have a seed corpus or target specific dictionary yet.)
- Can likely be improved further in the future by exercising additional features of `Repo` to the harness.

Here are some very rough stats to give a very rough idea of the difference:

| Metric     | `fuzz_repo.py` (my local testing)        | `fuzz_tree.py` (most recent successful CF run)          |
|------------|------------------------|-------------------------|
| Coverage   | 789                    | 163                     |
| Features   | 1023                   | 269                     |
| Corpus     | 19 entries / 57 bytes  | 30 entries / 1872 bytes |

### Note on License

Because `fuzz_tree.py` was removed and `fuzz_repo.py` was not derived from it, the Apache License call outs in the docs were also updated as they only apply to the singe `fuzz_config.py` file now.

[^1]: https://github.com/gitpython-developers/GitPython/pull/1901#discussion_r1565001609
[^2]: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=68355


DaveLak:
CC @EliahKagan

DaveLak:
and FWIW, if removing `fuzz_tree` is not desired for any reason, I did test out `tempfile.TemporaryDirectory()` there and it works well enough to address the primary concerns raised. My fork has the commit with it: https://github.com/DaveLak/GitPython/commit/3a049599924d081fbeb1a8662ee6d2a2af4af80c

EliahKagan (approved):

Byron (approved):
Thanks a million!

More efficient fuzzing is great!

PR #1913, opened 2024-05-08
Update OSS-Fuzz Scripts to Use New QA-Assets Repo Structure

DaveLak:
**merging https://github.com/gitpython-developers/qa-assets/pull/2 should be coordinated with merging this PR.**

This change is required to support the changes to the seed data repo structure introduced in:
https://github.com/gitpython-developers/qa-assets/pull/2

This moves most of the seed data related build steps into the OSS-Fuzz Docker image build via `container-environment-bootstrap.sh`. This includes moveing the dictionaries into that repo.

The fuzzing/README.md here should be updated in a follow-up with a link to the qa-assets repo (and probably some context setting about corpora in general) but I have opted to defer that as I think the functionality added by the seed data improvements is valuable as is and shouldn't be blocked by documentation writers block.


Byron (approved):
Thanks a lot for driving fuzzing forward, relentlessly!

I will set this to auto-merge so when done, you could merge the PR in `qa-assets` yourself.

DaveLak:
> Thanks a lot for driving fuzzing forward, relentlessly!

Thanks for letting me! I've still got my eyes on `gitoxide` fuzzing, so don't worry 😄 

> I will set this to auto-merge so when done, you could merge the PR in `qa-assets` yourself.

Thanks! Will do.

Byron:
It turns out there is no auto-merging for me, so I will just merge both myself when I get back.

PR #1915, opened 2024-05-15
Instrument test utility functions to increase fuzzer efficiency

DaveLak:
Fuzz Introspector was reporting a high percentage of fuzz blockers in the `fuzz_diff` test. This means the fuzzing engine was unable to gain visibility into functions lower in the call stack than the blocking functions, making it less effective at producing interesting input data.

This clears a large percentage of the fuzz blockers by adding fuzzer instrumentation to them via the `@atheris.instrument_func` decorator.

**Fuzz Introspector Before & After Screenshots**

<details><summary>Before</summary>

image

</details> 

<details><summary>After</summary>

image

</details>


Byron (approved):
Thanks a million! This looks like a tiny change for massive gains.

PR #1925, opened 2024-06-02
typing fixes - DiffIndex generic type and IndexFile items

Andrej730:
Example issue with DiffIndex:
```python
repo = git.Repo(path_dir)
diff = repo.index.diff(None)
modified_files = [d for d in repo.index.diff(None)]
reveal_type(modified_files) # list[Unknown] instead of list[Diff]
```
Example issue with IndexFile:
```python
path: os.PathLike = ...
repo = git.Repo(path_dir)
# Argument of type "PathLike[Unknown]"
# cannot be assigned to parameter "items" of type "Sequence[git.types.PathLike | Blob | BaseIndexEntry | Submodule]" in function "add"
# "PathLike[Unknown]" is incompatible with "Sequence[git.types.PathLike | Blob | BaseIndexEntry | Submodule]"
repo.index.add(path)
```

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -187,7 +187,7 @@ class Diffable:
         paths: Union[PathLike, List[PathLike], Tuple[PathLike, ...], None] = None,
         create_patch: bool = False,
         **kwargs: Any,
-    ) -> "DiffIndex":
+    ) -> "DiffIndex[Diff]":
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot for the typing fix, it's much appreciated.

PR #1926, opened 2024-06-04
Fix Improper Import Order Breaking `fuzz_submodule` Fuzzer

DaveLak:
ClusterFuzz runs of the `fuzz_submodule` target have been failing because the `git` import was placed before the condition that sets the Git executable path 😞 

The order in which `git` is imported matters because it attempts to find a Git executable as the import is loaded (via `refresh()` in `git/__init__.py`.) As per #1909, we configure the ClusterFuzz environment to use a bundled Git executable via the env variable condition in all fuzz targets.


Byron (approved):
Thanks a lot!

Let's try it out :).

Issue #1924, opened 2024-05-31
Can't catch GitCommandError


zerothi:
I am trying to do a cherry-pick, but abort when it fails:

```python
import git
...
git = repo.git
try:
    git.cherry_pick(commit)
except git.GitCommandError as e:
    git.cherry_pick(abort=True)
...
```
but I get this:
```
    except git.GitCommandError as e:
   TypeError: catching classes that do not inherit from BaseException is not allowed
```
Now I can obviously do `except Exception as e`... But, I would like to be explicit to defer code-paths.

Byron:
Thanks for reporting!

This seems fixable and testable, and even though I don't know what's the best way to fix this, somebody coming along will.
Thanks for helping.

EliahKagan:
The problem is not with the `GitCommandError` exception, which [works fine in `except` clauses](https://github.com/search?q=repo%3Agitpython-developers%2FGitPython%20%22except%20GitCommandError%22&type=code) and, like all exceptions in Python code, does derive ultimately from `BaseException`.

This is instead a bug in the code you showed. That code does not catch that exception, because that exception is not what really appears in the `except` clause. The problem is this assignment statement:

```python
git = repo.git
```

Before that, `git` was the `git` module, whose `GitCommandError` attribute is an exception. After that, `git` is an instance of the `Git` class. All subsequent uses of `git` are as if `repo.git` had been used. So then you have:

```python
except git.GitCommandError as e:
```

Because you reassigned `git`, this is like:

```python
except repo.git.GitCommandError as e:
```

In most cases, a bug like this would result in an `AttributeError`. But the attributes of a `Git` instance, unless otherwise present, are dynamically generated functions that, when called, invoke the `git` executable to run `git` commands. For example, `repo.git.gc` is a function that, when called, runs a `git gc` command. Likewise, `repo.git.GitCommandError` is a function that, when called, runs a `git GitCommandError` command.

[...]

PR #1932, opened 2024-06-13
fixed doc to not faulty do #1924

zerothi:
Small change on documentation to make it clear how to use the `repo.git` object (not to confuse `import git`).

This is not a fix for #1924, but was related due to this documentation string.


Byron (approved):
Thanks a lot, that's much appreciated!

PR #1935, opened 2024-06-24
Change aliases to work around mypy issue.

### LINKED ISSUE #1934: mypy is confused by property aliases
Thank you very much for this very useful python package. Unfortunately mypy is very confused by property aliases such as 

```python
    @property
    def references(self) -> "IterableList[Reference]":
        """A list of :class:`~git.refs.reference.Reference` objects representing tags,
        heads and remote references.

        :return:
            ``git.IterableList(Reference, ...)``
        """
        return Reference.list_items(self)

    # Alias for references.
    refs = references
```

that can be found in [`git/repo/base.py`](https://github.com/gitpython-developers/GitPython/blob/4c21e514cd9b5acdc34891fc4b52e9b599810b3e/git/repo/base.py#L405C1-L416C22) or the analogous alias `branches` for `heads`. See https://github.com/python/mypy/issues/6700

You can reproduce the issue with the following `test.py`:
```python
from git.repo import Repo

repo = Repo()
for branch in repo.branches:
    print(branch)
```
about which `mypy` says:
```
test.py:4: error: "Callable[[], IterableList[Head]]" has no attribute "__iter__" (not iterable)  [attr-defined]
```

I am not aware of any fix that would not uglify the code here, and that mypy issue doesn’t seem likely to see improvement in the near future. This is sad, but doing all the work to add type annotations and still having users facing incomprehensible error messages in perfectly legitimate code is also sad. 

Could you please tell me whether you would welcome a PR “fixing” this?

PatrickMassot:
Fixes #1934
Note this should also gives better LSP support to these property aliases.

### DIFF
--- a/git/remote.py
+++ b/git/remote.py
@@ -828,8 +828,15 @@ class Remote(LazyMixin, IterableObj):
             name._clear_cache()
         return name
 
-    # `rm` is an alias.
-    rm = remove
+    @classmethod
+    def rm(cls, repo: "Repo", name: str) -> str:
+        """Alias of remove.
+        Remove the remote with the given name.
+
+        :return:
+            The passed remote name to remove
+        """
+        return cls.remove(repo, name)
 
     def rename(self, new_name: str) -> "Remote":
         """Rename self to the given `new_name`.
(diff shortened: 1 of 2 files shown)


Byron (approved):
That's great, thanks so much!

PR #1941, opened 2024-07-10
Rename README.md to pkg install git python

Kokami657:
git clone https://github.com/username/spaceman-bot.git


EliahKagan (commented):
It looks like this pull request was opened by accident, since it doesn't really make sense to rename `README.md` to `pkg install git python`.

PR #1943, opened 2024-07-17
Add type of change to files_dict of a commit

JonasScharpf:
This allows to not only get the total, inserted or deleted number of lines being changed but also the type of change like Added (A), Copied (C), Deleted (D), Modified (M), Renamed (R), type changed (T), Unmerged (U), Unknown (X), or pairing Broken (B)

### DIFF
--- a/test/test_stats.py
+++ b/test/test_stats.py
@@ -14,13 +14,19 @@ class TestStats(TestBase):
         output = fixture("diff_numstat").decode(defenc)
         stats = Stats._list_from_string(self.rorepo, output)
 
-        self.assertEqual(2, stats.total["files"])
-        self.assertEqual(52, stats.total["lines"])
-        self.assertEqual(29, stats.total["insertions"])
+        self.assertEqual(3, stats.total["files"])
+        self.assertEqual(59, stats.total["lines"])
+        self.assertEqual(36, stats.total["insertions"])
         self.assertEqual(23, stats.total["deletions"])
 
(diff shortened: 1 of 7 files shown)


test/test_stats.py
```diff
@@ -14,13 +14,19 @@ def test_list_from_string(self):
         output = fixture("diff_numstat").decode(defenc)
         stats = Stats._list_from_string(self.rorepo, output)
 
-        self.assertEqual(2, stats.total["files"])
-        self.assertEqual(52, stats.total["lines"])
-        self.assertEqual(29, stats.total["insertions"])
+        self.assertEqual(3, stats.total["files"])
```
Byron, line 17: Can you help me understand why this amount (and the following two) changed?

Byron (commented):
Thanks a lot for contributing this improvement. It really does look great already.

There is one question left, could you take a look?
Thanks

test/test_stats.py
JonasScharpf: I added another `diff_numstat file with change type `A` just to check. The other two files `a.txt` and `b.txt` have either deletions or changes and are of change type `C`, `b.txt` is potentially `R` but not sure about that.
Anyway if there will be a more dynamic testing with really adding, removing, changing files in a fake repo it would be good to not only test for `M`, so this could be a starting point

Byron: Ah, right, this adds up perfectly. Apologies for not seeing it earlier.

Byron (approved):
Thanks a lot for contributing, much appreciated.

Issue #1893, opened 2024-04-01
Some xfail markings fail to validate their exception types


EliahKagan:
This bug was discovered by @Borda in #1888. Because that PR's central focus is tooling to catch typos, I'm creating this issue so that this bug in the test suite is clear, and so it will not be forgotten regardless of how development proceeds there. This affects three test cases in `test_index.py`.

In #1777, I introduced `test_run_commit_hook`. This is expected to fail in a couple of situations where it should nonetheless be feasible to run the test suite. One of these is when the tests are run on a (native) Windows system that has no `bash` that can be found in a path search, neither in `System32` or otherwise and neither WSL-related or otherwise. The test included an xfail marking to cover that case:

https://github.com/gitpython-developers/GitPython/blob/53640535cf8314366a01da081947dd8504a299cd/test/test_index.py#L1018-L1022

In that PR I also added those xfail markings to other tests to which they applied, and I used copy and paste to reproduce the parts that should be exactly the same across them.

But that contains a typo, passing a `rasies` argument when constructing the mark, rather than the correct `raises` argument. The wrong argument is apparently ignored, so the marking is unconstrained, and even if those tests fail with a different exception -- one that is not covered by other xfail markings, and that is not intended to be covered by the ones with the misspelling, including `AssertionError` from an `assert` statement in the test -- the problem would not be caught.

This is distinct from the more common problem of misspelling a mark, which is also not an error, but which `pytest` does have the ability to warn about.

**Fortunately, the way those tests fail is as the xfail markings attempt to describe and will actually describe once fixed.** I have retested the cases not covered on CI, and they work with the changes to `test_index.py` appearing in [a20d5db](https://github.com/gitpython-developers/GitPython/pull/1888/commits/a20d5db6dd2943057e8709769d131db9cf821611) (#1888). I only checked out the changes to `test_index.py` in testing this, not the other files. In a virtual machine running Windows Server 2022 (version 10.0.20348), I have just tested the cases where:

- `bash.exe` exists, is the WSL-related command in `System32`, but there is no WSL system installed.
- `bash.exe` exists, is not the WSL-related command in `System32` but is instead Git Bash, and is working.
- `bash.exe` does not exist anywhere found in a `Popen` call, not even in `System32`.

[...]

PR #1945, opened 2024-07-24
Use Alpine Linux in WSL on CI

EliahKagan:
Some of the CI tests use WSL. This switches the WSL distribution from Debian to Alpine, which might be slightly faster. For the way it is being used here, the main expected speed improvement would be to how long the image would take to download\*, as Alpine is smaller.

(The reason for this is thus unrelated to the reason for the Alpine docker CI test job added in #1826. There, the goal was to test on a wider variety of systems and environments, and that runs the whole test suite in Alpine. This just changes the WSL distro, used by a few tests on Windows, from Debian to Alpine.)

Two things have changed that, taken together, have unblocked this:

- https://github.com/Vampire/setup-wsl/issues/50 was fixed, so the action we are using is able to install Alpine Linux. See: https://github.com/gitpython-developers/GitPython/pull/1917#pullrequestreview-2081550232

- #1893 was fixed in #1888. So if switching the WSL distro from Debian to Alpine breaks any tests, including by making them fail in an unexpected way that raises the wrong exception, we are likely to find out.

---

\* Alpine Linux doesn't ship with `bash` so this has the `setup-wsl` action install the `bash` package. Updating package indexes and installing the package in the WSL system adds to the overhead. It is still much faster to set up than Debian... except that obtaining the Alpine Linux image, even though it is smaller than the Debian image, tends to take longer, because it is not available from as fast of a source. This relates to https://github.com/Vampire/setup-wsl/issues/50#issuecomment-1808995756; the issue there was about the action not being able to download from the second source at all, which was fixed.

The effect is that Alpine Linux takes longer to download from outside of GitHub, but is faster to retrieve from the GitHub Actions cache and install. Looking at the total time difference added up from all six Windows jobs that use it, this change does slightly better than with Debian. But the slight increase in complexity would likely make it unjustified given the mixed returns. But that's on the feature branch. I believe the cached download will be reusable across separate workflow runs if this is merged, so I would expect a significant overall improvement, with a total time spent getting WSL set up decreased decreased well under half the original total time. So I think it's probably justified.

This is subjective and I will not be sad if you choose to close this without merging.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -46,7 +46,7 @@ jobs:
       if: startsWith(matrix.os, 'windows')
       uses: Vampire/setup-wsl@v3.1.1
       with:
-        distribution: Debian
+        distribution: Alpine
 
     - name: Prepare this repo for tests
       run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -47,6 +47,7 @@ jobs:
       uses: Vampire/setup-wsl@v3.1.1
       with:
         distribution: Alpine
+        additional-packages: bash
 
     - name: Prepare this repo for tests
       run: |


Byron (approved):
Thanks a lot being on top of this!

Let's merge and see if we are lucky with compile-time improvements due to effective GitHub CI caching.

PR #1948, opened 2024-07-30
Deprecate default of Git.show(strip_newline_in_stdout=False)

SonOfLilit:
As per our discussion in #1377.

Also fix a few tests that were broken by my git being configured to create new repos with `main` as the initial branch, so that all my tests pass.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1210,6 +1210,13 @@ class Git(metaclass=_GitMeta):
         if self.GIT_PYTHON_TRACE and (self.GIT_PYTHON_TRACE != "full" or as_process):
             _logger.info(" ".join(redacted_command))
 
+        if strip_newline_in_stdout and command[:2] == ["git", "show"]:
+            warnings.warn(
+                "Git.show() has strip_newline_in_stdout=True by default, which probably isn't what you want and will "
+                "change in a future version. It is recommended to use Git.show(..., strip_newline_in_stdout=False)",
+                DeprecationWarning
+            )
+
         # Allow the user to have the command executed in their working dir.
         try:
             cwd = self._working_dir or os.getcwd()  # type: Union[None, str]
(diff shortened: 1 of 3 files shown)


git/cmd.py
```diff
@@ -1210,6 +1210,14 @@ def execute(
         if self.GIT_PYTHON_TRACE and (self.GIT_PYTHON_TRACE != "full" or as_process):
             _logger.info(" ".join(redacted_command))
 
+        if strip_newline_in_stdout and command[:2] == ["git", "show"]:
```
Byron, line 1213: `"git"` shouldn't be hard-coded, and I am pretty sure there is a variable for this somewhere.

git/cmd.py
```diff
@@ -1210,6 +1210,14 @@ def execute(
...
+            warnings.warn(
+                "Git.show() has strip_newline_in_stdout=True by default, which probably isn't what you want and will "
```
Byron, line 1215:
```suggestion
                "Git.show() has strip_newline_in_stdout=True by default, which probably isn't what you want and may "
```

test/test_repo.py
```diff
@@ -1377,13 +1377,23 @@ def test_rebasing(self, rw_dir):
...
         fp = osp.join(rw_dir, "hello.txt")
         with open(fp, "w") as fs:
             fs.write("hello\n")
         r.git.add(Git.polish_url(fp))
         r.git.commit(message="init")
-        self.assertEqual(r.git.show("HEAD:hello.txt", strip_newline_in_stdout=False), "hello\n")
+        return r
+
+    @with_rw_directory
+    def test_warn_when_strip_newline_in_stdout(self, rw_dir):
```
Byron, line 1393: Let's put the additional assertion done in `test_do_not_strip_newline_in_stdout` here so that it's clearer what's under test: The deprecation warning if `git show` is used and `strip_newline_in_stdout=True`. This way, `test_do_not_strip_newline_in_stdout` can be removed. I'd also hope that the assertion then can be changed to use the default value for `strip_newline_in_stdout`, which is probably `True`, the cause of the confusion in this case.

Byron (changes requested):
Thanks a lot for contributing.

I think with a few minor modifications this PR can be merged.

Byron:
Closing due to inactivity.

PR #1951, opened 2024-08-09
Gracefully handle `PermissionError` exceptions that crash fuzzer

DaveLak:
Fuzzing inputs sometimes produce directory paths that are protected inside the fuzzer execution environment. This is not an issue in GitPython's code, so it should not crash the fuzzer.

Fixes OSS-Fuzz Issue 69456:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69870


Byron (approved):
Thanks a lot for maintaining the fuzzing!

PR #1952, opened 2024-08-13
Fuzzing: Gracefully Handle Uninteresting Error to Fix OSS-Fuzz Issue

DaveLak:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=71095

Fuzzing data can generate filenames that trigger:
> OSError: [Errno 36] File name too long

The changes here add handling for these exceptions because they di not indicate a bug and should not crash the fuzzer.


Byron (approved):
Looks good to me, thanks for making this work!

It's interesting that the Rust version of this doesn't have this entire class of issues - the only thing that matters are panics, errors like these are considered valid behaviour.

Of course, in Python there is no such differentiation, hence the extra-trouble of having to allow all kinds of exceptions that the fuzzer can legitimately run into.

PR #1953, opened 2024-08-16
Update and adjust pre-commit hooks

EliahKagan:
This updates pre-commit hooks to their latest stable versions and adjusts exclusions and tool configuration to avoid false positives, as well as adding one more hook one of whose benefits is to catch problems from false positives of an existing hook.

#### Bumping versions

[Ruff 0.6](https://astral.sh/blog/ruff-v0.6.0) has come out, with some improvements, so we may as well use it. Updating the hook for it achieves that, and while doing so, I think makes sense to update the others.

Over time, hooks that are not updated will become more out of date, so that the number of breakages that have to be solved when a bug or feature consideration requires an update is greater. No changes were needed to accommodate Ruff 0.6. But more false positives occur with the newer version of codespell. This addresses them by adding some more explicitly ignored words in its `pyproject.toml` configuration.

#### Avoiding symlink path corruption

Separate from any changes introduced by those updates, there is a problem with the way the `end-of-file-fixer` operates. As covered in more detail in commit messages, on Windows, symlinks are often checked out as regular files. But these files should not have newlines appended to the end of them, because when such a change is committed, it is adding a newline to the end of the path the repository stores for the symlink. Then, when the symlink is actually checked out, it is broken.

This excludes files named like licenses from being scanned and "fixed" by end-of-file-fixer, since that is both the current situation where the problem happens and the most likely way it would arise in the future without being detected through use of the repository.

It also adds a pre-commit hook (not used in this repository before) to check all symlinks. This will never find that error in a local repository where it arises, because it only checks actual symlinks, but it will find it on CI after it arises if it ever does again. There may also be a further benefit to it, since dangling symlinks could be accidentally introduced in the future for other reasons.


Byron (approved):
Thanks a lot for your help, much appreciated.

EliahKagan:
No problem! Should anything further be done on it before it is ready to be merged?

Byron:
I might have failed to press the button 🤦‍♂️

EliahKagan:
No problem!

EliahKagan:
I've noticed that the merge commit [failed CI](https://github.com/gitpython-developers/GitPython/actions/runs/10434173090/job/28896746547) on the main branch. But this appears unrelated to any of the changes here, because it is a random failure due to #1676. I expect that the failed check would succeed if re-run.

PR #1961, opened 2024-09-14
_to_relative_path to support mixing slashes and backslashes

Andrej730:
Working on Windows you sometime end up having some paths with backslashes (windows native) and some with slashes - this PR will resolve the issue using gitpython for those kind of cases (see example below). It will also fix the issues if paths contain redundant separators or "..".

```
import git

repo = git.Repo(r"C:\gittest")
repo.index.add(r"C:\gittest\1.txt")
# Traceback (most recent call last):
#   File "c:\second_test.py", line 5, in <module>
#     repo.index.add(r"C:/gittest/2.txt")
#   File "Python311\Lib\site-packages\git\index\base.py", line 879, in add
#     paths, entries = self._preprocess_add_items(items)
#                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "Python311\Lib\site-packages\git\index\base.py", line 672, in _preprocess_add_items
#     paths.append(self._to_relative_path(item))
#                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "Python311\Lib\site-packages\git\index\base.py", line 657, in _to_relative_path
#     raise ValueError("Absolute path %r is not in git repository at %r" % (path, self.repo.working_tree_dir))
# ValueError: Absolute path 'C:/gittest/2.txt' is not in git repository at 'C:\\gittest'

repo.index.add(r"C:/gittest/2.txt")

repo.index.commit("test")
```

### DIFF
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -653,7 +653,7 @@ class IndexFile(LazyMixin, git_diff.Diffable, Serializable):
             return path
         if self.repo.bare:
             raise InvalidGitRepositoryError("require non-bare repository")
-        if not str(path).startswith(str(self.repo.working_tree_dir)):
+        if not osp.normpath(str(path)).startswith(osp.normpath(str(self.repo.working_tree_dir))):
             raise ValueError("Absolute path %r is not in git repository at %r" % (path, self.repo.working_tree_dir))
         return os.path.relpath(path, self.repo.working_tree_dir)
 
--- a/git/index/base.py
+++ b/git/index/base.py
@@ -653,7 +653,7 @@ class IndexFile(LazyMixin, git_diff.Diffable, Serializable):
             return path
         if self.repo.bare:
             raise InvalidGitRepositoryError("require non-bare repository")
-        if not osp.normpath(str(path)).startswith(osp.normpath(str(self.repo.working_tree_dir))):
+        if not osp.normpath(str(path)).startswith(str(self.repo.working_tree_dir)):
             raise ValueError("Absolute path %r is not in git repository at %r" % (path, self.repo.working_tree_dir))
         return os.path.relpath(path, self.repo.working_tree_dir)
 
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -1181,6 +1181,18 @@ class TestIndex(TestBase):
 
         rw_repo.index.add(file)
 
+    @with_rw_repo("HEAD")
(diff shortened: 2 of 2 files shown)


git/index/base.py
```diff
@@ -653,7 +653,7 @@ def _to_relative_path(self, path: PathLike) -> PathLike:
             return path
         if self.repo.bare:
             raise InvalidGitRepositoryError("require non-bare repository")
-        if not str(path).startswith(str(self.repo.working_tree_dir)):
+        if not osp.normpath(str(path)).startswith(osp.normpath(str(self.repo.working_tree_dir))):
```
Byron, line 656: Is `self.repo.working_tree_dir` usually non-normalized? To me it seems it should be left alone here, or normalized upon repository initialization.

Byron (changes requested):
Thanks a lot - this looks like an improvement.

Could you also add a test that fails with the previous implementation?

Maybe as a question on the side: what prevents the caller from normalizing the input paths right away?

git/index/base.py
Andrej730: You're right, it's always normalized in here:
https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/repo/base.py#L233

I've also noticed that `working_tree_dir` is never `os.PathLike` it's always either `str` or `None` - so we can also change typing for `working_tree_dir` and `_working_tree_dir` and remove `str` conversion for it. What do you think?

Andrej730:
> Could you also add a test that fails with the previous implementation?

Added.


> Maybe as a question on the side: what prevents the caller from normalizing the input paths right away?

Well, normalizing the inputs is one error traceback back away indeed but usually when some method has a string-path argument user expects the method to handle any kind of paths.

git/index/base.py
Byron: To me it doesn't look it it's never `os.PathLike`, `_working_tree_dir` is initialised by many different possible branches. `expand_path` alone returns a `PathLike` for example.
On the other hand, I am not implying I *know* the type, it's a guess at best, which is probably why I'd just leave it as is to maybe increase the chance to not break something.

Byron:
Thanks for the test!

Once CI passes I think this can be merged.

Byron (approved):
Thanks a lot!

Issue #1963, opened 2024-10-03
There seems to be an error in the documentation comments


alex20230721:
it's not work in line 185 There seems to be an error in the documentation comments

can use:
write this :repo = git.Repo("\~/practiceS/long-story/.git")
or write this :repo = git.Repo("~/practiceS/long-story")

it's not work:
 repo = git.Repo("\~/practiceS/long-story.git") 
and
repo = git.Repo("~/practiceS/long-story/")

<img width="765" alt="image" src="https://github.com/user-attachments/assets/ae709f2b-2ac4-4060-b877-cbc7da38f0a4">

alex20230721:
git/repo/base.py

Byron:
What's the expected outcome?

Since it's related to documentation, maybe open a PR with the fix or what you think should be done.

PR #1965, opened 2024-10-04
Update base.py

alex20230721:
https://github.com/gitpython-developers/GitPython/issues/1963#issue-2563470127

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -179,7 +179,8 @@ class Repo:
         R"""Create a new :class:`Repo` instance.
 
         :param path:
-            The path to either the root git directory, the .git directory under the root git directory or the bare git repo::
+            The path to either the root git directory, the .git directory under the root git directory 
+            or the bare git repo::
 
                 repo = Repo("/Users/mtrier/Development/git-python")
                 repo = Repo("/Users/mtrier/Development/git-python.git")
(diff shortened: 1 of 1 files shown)


alex20230721:
Sorry, I just learned about the concept of bare repo today.

> Bare repositories are typically cloned with the `.git` extension, hence the example is representing something that is common. There is no point in trying to make it work for a non-bare clone.
> 
> However, you could add another line that shows that one can _also_ point at the `.git` directory that is typically sitting in a worktree.

Sorry, I just learned about the concept of bare repo today.

alex20230721:
> Bare repositories are typically cloned with the `.git` extension, hence the example is representing something that is common. There is no point in trying to make it work for a non-bare clone.
> 
> However, you could add another line that shows that one can _also_ point at the `.git` directory that is typically sitting in a worktree.

Sorry, I just learned about the concept of bare repo today.

git/repo/base.py
```diff
@@ -179,7 +179,8 @@ def __init__(
         R"""Create a new :class:`Repo` instance.
 
         :param path:
-            The path to either the root git directory or the bare git repo::
+            The path to either the root git directory, the .git directory under the root git directory
+            or the bare git repo::
```
Byron, line 183:
```suggestion
            The path to either the worktree directory or the .git directory itself::
```

Byron (approved):
Thank you, let me see if I can finalize this PR now.

PR #1987, opened 2025-01-02
Fix various version-related CI breakages

EliahKagan:
This applies the fixes to GitPython that correspond to https://github.com/gitpython-developers/smmap/pull/58, as well as other fixes which were along the lines of the prediction in https://github.com/gitpython-developers/GitPython/commit/e51bf80ad576256f2fbeead41ea3f0b667c77055#commitcomment-150862474. Taken together, this fixes CI for GitPython.

The Cygwin and Alpine Linux failures were due to not using a virtual environment.

- In Cygwin, I had deliberately not done that, since we're not doing it the other non-container workflows, and I wanted the Cygwin workflow to be similar, with the hope that its job might even one day become a matrix job in `pythonpackage.yml`. But there had already been signs, accumulating over time, that this was not the way to go, such as the extra step to make sure there would be a `pip` command in the global environment. Ultimately, what not using a venv broke was `pytest-cov`, which didn't have a dependency it needed, even after all dependencies appeared to install okay. Using a venv fixed that automatically.

- In Alpine Linux, I was deliberately using a virtual environment, since `python` in Alpine Linux discourages using `pip` to install packages in the global environment managed by the system, which by default is automatically rejected, and which would not necessarily work well even if forced to happen. However, at some point, the virtual environment stopped actually being activated. This causes that error, which had previously only occurred in this workflow while it was originally being developled, to return.

  I am not clear on specifically why this happened, but I think it is due to the interaction between:

  1. The environment variables that are documented to be set when a venv is activated.
  2. Subtle distinctions between activating a venv in the usual way and setting those variables (maybe).
  3. How setting variables for steps via `GITHUB_ENV` composes with `sudo` passthrough.

  Of those, the third is the most significant: `sudo` is being used as the `shell` for most steps in this job, in order to run code as a non-root user (since some tests do not work when run as root and are not, if I recall correctly, intended to). But it was not configured to pass through variables related to Python virtual environments. I'm not sure what changed, exactly, but in hindsight I don't think I could ever have proved that it was *supposed* to work as I intended, when implemented the way I had. The fix here is slightly ugly in that it explicitly activates the venv in each of the steps that are supposed to use it, but I believe this is more simple and robust than alternatives I considered.

As for the other, unrelated Alpine problem in WSL, where the image didn't downloading, upgrading the setup-wsl action fixed it. (To do this, I cherry-picked a commit from the Dependabot PR that had proposed it.)

See commit messages for full details on each of the changes themselves.

### DIFF
--- a/.github/workflows/alpine-test.yml
+++ b/.github/workflows/alpine-test.yml
@@ -16,10 +16,10 @@ jobs:
     steps:
     - name: Prepare Alpine Linux
       run: |
-        apk add sudo git git-daemon python3 py3-pip
+        apk add sudo git git-daemon python3 py3-pip py3-setuptools py3-virtualenv py3-wheel
         echo 'Defaults env_keep += "CI GITHUB_* RUNNER_*"' >/etc/sudoers.d/ci_env
         addgroup -g 127 docker
-        adduser -D -u 1001 runner
+        adduser -D -u 1001 runner  # TODO: Check if this still works on GHA as intended.
         adduser runner docker
       shell: sh -exo pipefail {0}  # Run this as root, not the "runner" user.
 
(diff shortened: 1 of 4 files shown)


Byron (approved):
Thanks so much - I definitely couldn't have produced a fix like this, if at all!

Let's hope it will last till 2026 :).

EliahKagan:
> Let's hope it will last till 2026 :).

Well, there's still #1955. I hope to do that long before 2026 :smile: (unless someone else gets to it first). Further CI refinements could probably be included there. Actually, that's pretty much all I have there now, though I have also updated the description to state in detail what the situation is with `os.path` in Python 3.13 in Windows.

PR #1988, opened 2025-01-04
Support Python 3.12 on Cygwin

EliahKagan:
Cygwin has had Python 3.9 as its latest version for a while, with progress both on moving past 3.9.16 for Python 3.9, as well as on packaging Python 3.*y* for higher *y*, stalled due to the problem described in #1814. However, [packaging work has resumed](https://cygwin.com/pipermail/cygwin/2024-December/256957.html). This includes [Cygwin builds of Python 3.12](https://cygwin.com/cgi-bin2/package-grep.cgi?grep=python312) (which is [successfully packaged](https://cygwin.com/pipermail/cygwin/2024-December/256959.html) for the parallel/forked MSYS2 target). Such builds are marked "test" and do not work properly yet (reports and discussion can be seen in [this month's list archive](https://cygwin.com/pipermail/cygwin/2025-January/thread.html)).

Since GitPython has long-standing support for Cygwin, which I suspect remains one of the important use cases of GitPython (with some alternatives not supporting Cygwin), it seems to me that it is worthwhile to support Python 3.12 on Cygwin, as well as to test it on CI. Since the Cygwin downstream build of Python 3.12 is currently experimental and not fully working, I don't expect GitPython to work with it yet. This PR is for testing that, and for keeping track of changes that are known to be wanted whenever Cygwin Python 3.12 is working. I have encountered substantial differences between CI and my local setup with Cygwin in the past, so testing this on CI now may help avoid surprises. It may also lead to the discovery of bugs, or information about known bugs, that could be useful to report on the Cygwin mailing list. This PR is *definitely not* ready to be merged.

This PR is separate from and in parallel to #1955 (which is about Python 3.13 and blocked due to a need to change some GitPython code to support that version on Windows), because they are conceptually separate with no known overlap in blockers or in changes that would fix them, and because it seems to me that having two PRs makes the situation clear, both to me and to other people who might wish to contribute.

PR #1989, opened 2025-01-04
Do some CI cleanup to make reports clearer and future changes easier

EliahKagan:
Neither #1955 nor #1988 is ready yet, but both include commits that clean up CI and that I think are already valuable. Some of the changes make GitHub Actions checks easier to read (8a05390, 41377d5), while another removes some complexity that is no longer needed (73ddb22). So I cherry-picked those commits onto a new feature branch, which this PR proposes merging. If this is merged, I'll rebase those other PRs to drop the corresponding commits from them. Details on the changes are in the commit messages.

### DIFF
--- a/.github/workflows/alpine-test.yml
+++ b/.github/workflows/alpine-test.yml
@@ -3,7 +3,7 @@ name: test-alpine
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
 
     container:
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Set up Cygwin
       uses: egor-tensin/setup-cygwin@v4
       with:
-        packages: python39=3.9.16-1 python39-pip python39-virtualenv git
+        packages: python39 python39-pip python39-virtualenv git
 
     - name: Arrange for verbose output
       run: |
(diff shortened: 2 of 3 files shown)


Byron (approved):
That makes sense, thanks a lot!

*The initial error seemed spurious so I cancelled the workflow to try again.*

EliahKagan:
Yes, [that error](https://github.com/gitpython-developers/GitPython/actions/runs/12612467751/job/35149221389?pr=1989#step:12:1103) is due to #1676.

EliahKagan:
Another part of #1955 that could be brought in now is that GitPython is expected to, and seems to, completely work on Python 3.13 on all platforms *except* Windows--where it probably mostly works, but does not fully work. To fully support running, building, and testing GitPython with Python 3.13 fully non-Windows systems, either only #1954 was needed, or if the bug https://github.com/gitpython-developers/gitdb/pull/115 fixed was 3.13-specific, then it was also needed for that. The `os.path.isabs` issue does not apply outside Windows, nor does its generalization in terms of the need to check for how leading-`/` and leading-`\` paths are treated overall.

So Python 3.13 could be tested now on non-Windows systems. I don't know if you want that, or if you prefer to wait until it works on Windows and having that come in with #1955. (The special-casing, if it is done, would not be cumbersome. One way to do it would be the same way we exclude Python 3.7 from being tested on macOS.)

Byron:
Thanks for the heads-up.

> So Python 3.13 could be tested now on non-Windows systems. I don't know if you want that, or if you prefer to wait until it works on Windows and having that come in with #1955.

Since you'd be the one driving this I would leave it up to you to decide if it's worth having conditional testing to start testing earlier. To me it seems beneficial to have it as well.

EliahKagan:
I've opened #1990 for this.

PR #1990, opened 2025-01-05
Test Python 3.13 regularly on Ubuntu and macOS on CI

EliahKagan:
This enables testing Python 3.13 in the main CI test workflow for Ubuntu and macOS, as discussed in comments in #1989, but not Windows yet (pending #1955).

I think it makes sense to test these on CI, because:

- They're working now, and if they stop working, we would want to know.
- While Python 3.13 shouldn't be added as a classifier in `setup.py` yet since it's not fully working on Windows, testing other platforms on CI makes it easier for users to know they are meant (and believed) to work.
- The Ubuntu and macOS job are fast, so this probably isn't adding too much load, delay, energy usage, etc.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -13,10 +13,12 @@ jobs:
     strategy:
       matrix:
         os-type: [ubuntu, macos, windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         exclude:
         - os-type: macos
-          python-version: "3.7"
+          python-version: "3.7"  # Not available for the ARM-based macOS runners.
+        - os-type: windows
+          python-version: "3.13"  # FIXME: Fix and enable Python 3.13 on Windows (#1955).
         include:
         - os-ver: latest
         - os-type: ubuntu


Byron:
Thanks a lot!

EliahKagan:
No problem! It looks like this may have been meant to be merged. If so, auto-merge is still waiting on a review. (After it is merged, I'll rebase #1955 again.)

Byron (approved):

Byron:
Thanks for the reminder! Apparently merges only happen after approval. However, I noticed that once approved, they can be merged even while CI is running. Probably because `main` isn't protected.

EliahKagan:
It sounds like protecting `main` may be the way to go, so that approvals can be made either before or after CI finishes, without risk of inadvertently merging something that fails the tests. Is there any reason not to do that? Are there any changes to the workflows themselves that should be made to go along with or support it?

Byron:
I took at look and it seems like `main` is already protected, and as requirement for merging, an approval is set. What's also enabled is the need for checks to pass, but thanks to the worst UX, the checkbox does nothing. Instead, one has to pick each and every job individually, without a way to pick all at once.

This is where it stopped, again, and it's kind of where the famous auto-merge tends to stop for me 😅.

EliahKagan:
If I understand correctly, this is more manageable in `gitoxide` due to https://github.com/GitoxideLabs/gitoxide/commit/943350a71318763e0f9c6a7b795c3f66af963849 (https://github.com/GitoxideLabs/gitoxide/pull/1551), plus the refinement in https://github.com/GitoxideLabs/gitoxide/commit/4e672ef9b62287a5940787a3fcf97e97976ccd03 (https://github.com/GitoxideLabs/gitoxide/pull/1668). Sometime soon, I can look into making a PR here with something like that.

PR #1991, opened 2025-01-05
Affirm that gitdb and smmap advisories can also be created

EliahKagan:
This expands `SECURITY.md` to affirm the claims in the new `SECURITY.md` files in gitdb and smmap that vulnerabilities found in them can be reported in the GitPython repository with the same link as one would use to report a GitPython vulnerability, as well as to note how the distinction between affected package can be specified when it is known at the time a vulnerability is reported.

Along with https://github.com/gitpython-developers/smmap/pull/59 and https://github.com/gitpython-developers/gitdb/pull/117, this fixes https://github.com/gitpython-developers/gitdb/issues/116.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,4 +11,6 @@ Only the latest version of GitPython can receive security updates. If a vulnerab
 
 ## Reporting a Vulnerability
 
-Please report private portions of a vulnerability to <https://github.com/gitpython-developers/GitPython/security/advisories/new>. Doing so helps to receive updates and collaborate on the matter, without disclosing it publicliy right away.
+Please report private portions of a vulnerability to <https://github.com/gitpython-developers/GitPython/security/advisories/new>. Doing so helps to receive updates and collaborate on the matter, without disclosing it publicly right away.
+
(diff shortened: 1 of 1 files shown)


EliahKagan:
The setup-wsl action failed on Windows in Python 3.12 here (here's [the log](https://github.com/gitpython-developers/GitPython/actions/runs/12618020423/job/35160952489)). It is extremely unlikely to relate to the actual change in this PR, which modifies only `SECURITY.md`. I've done a no-change amend and forced pushed to achieve the effect of rerunning the tests, which I expect to pass.

I do wonder if Alpine Linux for WSL is more likely to have this problem than Debian, which was used before and which I believe setup-wsl still downloads from a different source.

Byron (approved):
Thanks a lot, this will do perfectly!

EliahKagan:
Actually, the links I added are broken! The branch name part of the URLs is wrong, so they give not-found errors. (I had tested the links from gitdb and smmap to here, but not the ones here to gitdb and smmap.) I've opened #1992 to fix this.

Issue #1995, opened 2025-01-09
rev_parse does not handle <tag>^{commit} correctly


flichtenheld:
`git rev-parse <tag>^{commit} ` resolves to the commit that the tag points to.

`repo.rev_parse("<tag>^{commit}")` says `ValueError: Could not accommodate requested object type 'commit', got tag`

It should instead return `TagObject.object` (at least if that is a commit object).

flichtenheld:
Re-verified that is present in the current release (3.1.44)

flichtenheld:
Interestingly it tries to resolve the tag when you specify `^{blob}` although it is not clear to me when this would ever work given that `TagObject.object` is usually a commit.

flichtenheld:
Found from the test suite that `<tag>^{}` works. So this a usable work-around. Will prepare a patch anyway.

Byron:
Thanks a lot for digging into this!

I may also add that this implementation is definitely not according to 'spec' and probably just superficially resembles `git rev-parse` semantics.
Probably it's just good enough for the common cases, which is a reason for issues around it being quite rare.

PR #1996, opened 2025-01-09
Repo.rev_parse: Handle <tag>^{commit} correctly

flichtenheld:
This should resolve to commit object.

Fixes: #1995


Byron (approved):
Thanks a lot for the fix, much appreciated!

PR #1997, opened 2025-01-10
Fuzzing: Fix Broken Fuzz Test for Git Submodule Handling

DaveLak:
Hey there 👋 

As you may or may not have seen, the `fuzz_submodule.py` harness has been failing for some time now and I believe it is responisble for the "fuzzing build: failing" status on OSS-Fuzz.

I looked into it and it appears to be an issue with the test implementation, not GitPython.

The changes here should fix the issue causing the failure!

Please see the commit message for additional details.

### DIFF
--- a/fuzzing/fuzz-targets/fuzz_submodule.py
+++ b/fuzzing/fuzz-targets/fuzz_submodule.py
@@ -9,9 +9,15 @@ from utils import (
     get_max_filename_length,
 )
 
-# Setup the git environment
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks for your continued support with this, it's much appreciated!

DaveLak:
This worked. Issue now resolved on the OSS-Fuzz tracker: https://issues.oss-fuzz.com/379326001

Issue #2004, opened 2025-02-20
Cygwin job is broken on "Set up virtualenv" step


EliahKagan:
The Cygwin CI test job is failing now, including [when rerun](https://github.com/EliahKagan/GitPython/actions/runs/13041816861/job/37557892745) at the tip of the main branch where it formerly passed. The failure occurs in the "Set up virtualenv" step, which runs `python -m venv .venv`. The log does not how a detailed error message, just:

```text
Run python -m venv .venv
  python -m venv .venv
  echo 'BASH_ENV=.venv/bin/activate' >>"$GITHUB_ENV"
  shell: C:\tools\cygwin\bin\bash.exe --login --norc -eo pipefail -o igncr "{0}"
  env:
    CHERE_INVOKING: 1
    CYGWIN_NOWINPATH: 1
    CYGWIN: 
+ python -m venv .venv
Error: Command '['/cygdrive/d/a/GitPython/GitPython/.venv/bin/python', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
Error: Process completed with exit code 1.
```

I first noticed this when checking the current status of #1988 (see https://github.com/gitpython-developers/GitPython/pull/1988#issuecomment-2672396018), but it is not specific to that PR.

I do not know why this happens.

Byron:
It is very puzzling indeed, and seemingly impossible to error out without any form of message.

EliahKagan:
The absence of output is due to `venv` not showing the error in its output. In 4605dd690b6ebe4c0c07f3910607aae407d6070e, I had `venv` not install `pip` but instead attempted to do so in a very similar way to how `venv` does, in a subsequent step. [That produces](https://github.com/EliahKagan/GitPython/actions/runs/13454947366/job/37596811693#step:10:30) this error:

```text
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 188, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.9/runpy.py", line 147, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/usr/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/usr/lib/python3.9/ensurepip/__init__.py", line [30](https://github.com/EliahKagan/GitPython/actions/runs/13454947366/job/37596811693#step:10:31), in <module>
[...]

PR #2007, opened 2025-02-25
Work around Cygwin CI failure #2004, except for `test_installation`

EliahKagan:
I don't have a complete fix for #2004 yet, but since it's confirmed to fail on the main branch of this upstream repository (mentioned in https://github.com/gitpython-developers/GitPython/pull/1988#issuecomment-2680840638, though not due to #1988), I figured my partial fix might be worth putting into place. With these changes, the problem in the Cygwin test job turns from being unable to run any tests to being able to run all the tests but with one new failure compared to the way it was before #2004 (i.e. with one non-xfail failure).

This uses the official `pip` bootstrap script https://bootstrap.pypa.io/get-pip.py to install `pip` in the virtual environment. This allows the actual installation of GitPython and its dependencies, including all test dependencies, to complete without problems. Aside from tests that are expected to fail and are already marked xfail, all tests pass *except* for `test_installation`, which fails for the same reason. It presumably would also pass if the bootstrap script were used there too, but I believe the current intent of that test is to verify that installation succeeds when done in an ordinary way.

I am likewise reluctant to mark `test_installation` as xfail on Cygwin, unless that turns out to be truly necessary, because none of this happens on my local Cygwin environment on a Windows 10 development machine. It seems like it is triggered by some combination of changing packages and something wrong or otherwise specific about the Cygwin workflow.

This PR includes some changes to the workflow that are not actually necessary for the workaround it confers. The changes that *are* needed are:

- Refraining from having `venv` automatically attempt to install `pip`.
- Installing `pip` in the virtual environment explicitly using the bootstrap script.
- Installing the `wget` Cygwin package to download the bootstrap script without extra complexity or risking encoding/line-ending mismatches.

All the rest is to make the workflow run a little faster, and *maybe* be slightly more robust in ways that are not relevant to #2004 and not very important. The most significant of them is switching back to the Cygwin installation action we had previously been using, which is less configurable (it does not support installing Cygwin packages at specific versions, which we had needed to do at one time), but which is faster because it does not first have to install Chocolatey to use it as an intermediary for installing Cygwin.

I kept those changes here because I think it's easier to iterate on this if it is left in. But if you'd like I'd be pleased to undo the nonessential changes, which can always be brought back later and separately. I have tested in 7fd5f15a02dfdd5f6471f6f290483b7c4bff34bc to make sure that this really works, so undoing those changes could be as simple as adding that one commit.

### DIFF
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install Cygwin
       uses: cygwin/cygwin-install-action@v5
       with:
-        packages: python39 python39-pip python39-virtualenv git
+        packages: python39 python39-pip python39-virtualenv git wget
         add-to-path: false  # No need to change $PATH outside the Cygwin environment.
 
     - name: Arrange for verbose output
--- a/.github/workflows/cygwin-test.yml
+++ b/.github/workflows/cygwin-test.yml
@@ -57,7 +57,7 @@ jobs:
 
     - name: Set up virtualenv
       run: |
-        python -m venv .venv
+        python3.9 -m venv .venv
         echo 'BASH_ENV=.venv/bin/activate' >>"$GITHUB_ENV"
 
     - name: Update PyPA packages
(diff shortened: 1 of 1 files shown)


EliahKagan:
I tried a few more things, but I did not find a good fix or better workaround, so I've opened #2009.

PR #2012, opened 2025-03-07
`IndexFile._to_relative_path` - fix case where absolute path gets stripped of trailing slash

kamilkozik7:
I encountered this issue on MacOS, python 3.7 and later

Added fix and appropriate unit test.

Stripping of trailing slash seems to be happening in `os.path.relpath`, altough I am not sure whether that is expected behaviour. In case it is not expected, and ever gets fixed, I added an additional check (` and not result.endswith(os.sep)`) to make sure that the trailing slash doesn't get duplicated.


kelibezhani (approved):
LGTM

Byron (approved):
Great, thanks a lot for the fix!
Looks good to me, too!

PR #2018, opened 2025-03-23
Fix some incorrect sphinx markups in the docstrings

koyuki7w:
Fixed some markups so [the api reference](https://gitpython.readthedocs.io/en/stable/reference.html) is rendered correctly.


Byron (approved):
Thanks a lot, appreciated!

PR #2027, opened 2025-05-22
use `sys.platform == "cygwin"` to figure out when we are using cygwin

gcmarx:
The current logic for detecting cygwin is:
- if `sys.platform == "win32"`, then it's not cygwin
- if the passed-in git executable is None, then it's not cygwin git
- finally, is there a `uname` executable in the same folder as the git executable, and if so, does the output of that command include `"CYGWIN"`? if so, then it's cygwin

In the [python 3.7 docs for sys.platform](https://docs.python.org/3.7/library/sys.html#sys.platform), Cygwin systems have `sys.platform == "cygwin"`. Since Python 3.7 is the oldest version still supported by GitPython, it stands to reason that we can rely on that being true for all supported Python versions.

The logic I propose is:
- if the git executable you passed in is None, it's not cygwin git
- if `sys.platform == 'cygwin'`, then it's cygwin git

This is simple enough for a single expression, so I replaced the body of the `is_cygwin` function with that expression and removed `is_cygwin_git` and friends completely.

I don't know if there's such a thing as `sys.platform == 'cygwin'` and the `git` not being cygwin `git`, but if there is, I don't think the existing code deals with that anyway, so this seems to return the same result and would fail in the same way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ __pycache__/
 # Transient editor files
 *.swp
 *~
+\#*#
+.#*#
 
 # Editor configuration
 nbproject
--- a/AUTHORS
+++ b/AUTHORS
@@ -55,5 +55,6 @@ Contributors are:
 -Eliah Kagan <eliah.kagan _at_ gmail.com>
 -Ethan Lin <et.repositories _at_ gmail.com>
 -Jonas Scharpf <jonas.scharpf _at_ checkmk.com>
+-Gordon Marx
 
 Portions derived from other open source works and are clearly marked.
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -32,7 +32,6 @@ from git.exc import (
 from git.util import (
     cygpath,
     expand_path,
-    is_cygwin_git,
     patch_env,
     remove_password_if_present,
     stream_copy,
@@ -661,7 +660,7 @@ class Git(metaclass=_GitMeta):
 
     @classmethod
     def is_cygwin(cls) -> bool:
-        return is_cygwin_git(cls.GIT_PYTHON_GIT_EXECUTABLE)
+        return cls.GIT_PYTHON_GIT_EXECUTABLE is not None and sys.platform == "cygwin"
 
     @overload
     @classmethod
--- a/git/util.py
+++ b/git/util.py
@@ -96,7 +96,6 @@ from git.types import (
     Files_TD,
     Has_id_attribute,
     HSH_TD,
-    Literal,
     PathLike,
     Protocol,
     SupportsIndex,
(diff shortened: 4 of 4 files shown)


gcmarx:
only this PR or #2026 should be merged

EliahKagan (commented):
*This is preliminary and I'll try to say more in a forthcoming review of #2026, which I hope to do tomorrow.*

I think we'll probably want to prefer the approach in #2026 over this, at least in the short term. There is a deliberate and long-standing (though not always perfectly applied) distinction throughout the code of GitPython between whether the platform is Cygwin (i.e. whether the `python` interpreter running the code of GitPython is a Cygwin program) and whether the `git` executable that GitPython calls is a Cygwin program.

This appears to be relevant to the treatment of absolute paths in some operations that involve cloning to an absolute path or adding a submodule. Mose use of `is_cygwin_git` is through `Git.is_cygwin`, and searching for `Git.is_cygwin` reveals uses where it looks like this distinction can matter.

GitPython uses whatever `git` executable it is configured to use via environment variables or, if not configured to use a particular `git` executable, whatever executable it finds in a path search. In either case, this executable need not be built for the same target as GitPython.

I admittedly would not recommend most such mismatches between `python` and `git` if they can be avoided. I also think we should be willing to run the risk of breaking things in what I believe to be the unusual case of using Cygwin `git` from outside Cygwin, if doing so is necessary in order to avoid or fix breakages when Cygwin is not used--especially for bugs that affect users on Unix-like operating systems that aren't related to Windows. This is one of the reasons I anticipate that the approach in #2026 is probably okay.

But distinguishing Cygwin `python` using Cygwin `git` from Cygwin `python` using Git for Windows `git` seems like it is at least partially working. It's also fairly easy to produce the situation of Cygwin `python` using Git for Windows `git`. This can plausibly occur intentionally or by accident. Most Cygwin environments preserve Windows `PATH` entries, allowing them to appear after Cygwin-specific `bin` directories. If GitPython is used in Cygwin `python`, on a system where Cygwin `git` is not installed but Git for Windows *is* installed, then *usually* the non-Cygwin `git` will be found in the `PATH`, and GitPython will use it.

gcmarx:
Thanks for the thorough review!

Neither the value of `sys.platform` nor “`uname` in the same directory as `git`” seem like a bulletproof way to tell whether the `git` in question is Cygwin git or not. I’m wondering if the git executable itself can be convinced to tell us about itself, but don’t currently have a way to test that. I’ll try to poke around at that later this week.

EliahKagan:
> only this PR or #2026 should be merged

Since this was intended to be merged only if #2026 was not, and #2026 is merged, I've closed this.

It's plausible--though I think not most likely--that we'll eventually end up going this route and no longer attempting to determine whether the `git` executable is a Cygwin program. In that case, this could be reopened, or a new PR opened to propose that. (However, I think, based on https://github.com/gitpython-developers/GitPython/pull/2026#discussion_r2112375887, that the next step may be to move to using `ldd` on the `git` binary after successfully finding it in a `PATH` search.)

PR #2032, opened 2025-05-30
Have CodeQL scan GitHub Actions workflows as well as Python code

EliahKagan:
This regenerates the `codeq.yml` workflow to modernize it and to get current defaults, the most (or perhaps only) important one being to scan both `python` and `actions`, rather than only `python`.

Parts of the default that we deliberately deviate from, or that seem preferable to change based on the comments in the default workflow, are put back.

---

I've tested this in my fork, but I'll still wait for all CI to complete here before merging. Besides being a good usual practice, this also has a chance of catching permissions-related problems.


EliahKagan:
> This pull request sets up GitHub code scanning for this repository.

I guess this comment was automatically posted because the approach I took was to delete the workflow in the first command and regenerate it. (Or maybe this comment is posted for all changes to the workflow now?)

PR #2037, opened 2025-06-07
Refactor Git.{AutoInterrupt,CatFileContentStream} nesting

EliahKagan:
This makes `Git.AutoInterrupt` and `Git.CatFileContentStream` transparent aliases to top-level nonpublic `_AutoInterrupt` and `_CatFileContentStream` classes in the `cmd` module.

This does not change the "public" interface. It also does not change metadata relevant to documentation: the `__name__` and `__qualname__` attributes are set explicitly to the values they had before when these classes were defined nested, so that Sphinx continues to document them (and to do so in full) in `Git` and as `Git.AutoInterrupt` and `Git.CatFileContentStream`.

The purpose of this is to increase readability. The `Git` class is big and complex, with a number of long members and various forms of nesting. Since these two classes can be understood even without reading the code of the `Git` class, moving the definitions out of the `Git` class into top-level nonpublic classes will hopefully increase readability and help with maintenance.

---

I originally intended this as just the first commit of several. But I realized that change is probably the only one that might be controversial: in my view, the flatter layout and physical separation makes for easier reading and emphasizes that, even as they are nested classes of the `Git` class as far as the interface is concerned, these classes are uses by the `Git` and do not themselves use the `Git` class.

In my view, the conceptual burden of setting `__name__` and `__qualname__` to make Sphinx continue working as desired (while still having the metadata at runtime the same as Sphinx claims) is not too big of a conceptual burden. But whether this is better is a subjective question, so I figured it would be good to get a review.

I have verified locally that the intended effect in Sphinx documentation--the effect that the classes continue to be presented, in full, as nested classes of `Git`, just as they may (and should!) continue to be used when they are accessed both within GitPython and by other code that uses GitPython--is what we get.


Byron (approved):
Thanks a lot, this makes perfect sense.
Thanks also for double-checking that the documentation is still working as it did before.

Please note that I didn't validate that the removed and added portions are still the same in the interest of time.

EliahKagan:
> Thanks a lot, this makes perfect sense. Thanks also for double-checking that the documentation is still working as it did before.

No problem!

It was not at all obvious to me what Sphinx would do if I had not modified the metadata. If that had turned out not to be necessary for Sphinx, then I would not be sure whether it would be best to do it or not, since I think the interface is really the same either way. In any case, I think it's better to modify the metadata like this and automatically keep the documentation, than not to do so and have Sphinx treat them as nonpublic and not emit documentation for them at all.

> Please note that I didn't validate that the removed and added portions are still the same in the interest of time.

It looks like some common techniques don't work here, due to a combination of changed indentation, moving the code, changing the lines at the top of the code that is moved (to `_` names), and changing some other lines near the top due different wrapping of docstrings.

I've double-checked it just now by:

1. Checking out the commit with:

   ```powershell
   git switch -d 5a8a4059d646fd313e81365ef240562556d8bd3f
   ```

2. Temporarily manually reindenting the moved lines in my editor, but otherwiser keeping everything the same.

3. Examining a diff made by:

   ```powershell
   git diff --color-moved=dimmed-zebra HEAD^
   ```

It looks right, with only the new, conceptually changed (e.g. adding `_`), and rewrapped lines appearing as red/green rather than as gray.

PR #2060, opened 2025-07-20
Fix name collision

### LINKED ISSUE #2023: Docs: No API Docs for git.config.GitConfigParser
The [API docs](https://gitpython.readthedocs.io/en/latest/reference.html) are great and detailed. However, the class `git.config.GitConfigParser` is only mentioned by name:

Image

The link "Alias of `GitConfigParser" refers to the line above.

It would be great if methods like `get_value(...)` would be documented.

NMertsch:
## The bug
In `git.config.MetaParserBuilder.__new__`:
```python
def __new__(name, bases, clsdict):  # <- `name` is the class name
    ...
    for name, method in methods:  # <- `name` is the current method name
        ...
    return super().__new__(cls, name, bases, clsdict)  # <- `name` is the name of the last method in clsdict
```

## Downstream effects
* `git.config.GitConfigParser.__name__` is `"write"`, not `"GitConfigParser"`
* Sphinx autodoc misinterprets the nature of `git.config.GitConfigParser` as an alias of itself:
* image

Fixes #2023


Byron:
Thanks, this PR would be great to land I am sure!

Do you have an intuition as to why CI is failing here?

NMertsch:
Yes. There was an rst syntax error in CPython's configparser module: https://github.com/python/cpython/blob/v3.8.10/Lib/configparser.py#L768:

```diff
- `vars'
+ `vars`
```

configparser.RawConfigParser is the baseclass of `GitConfigParser` (the class which is now being documented due to the bugfix), so its `get` method is part of the public interface of `GitConfigParser`.

This only occurs in the pipelines for 3.8 and 3.9, so I guess it was fixed later but the fix was never backported.

I'm not sure what's the best way to handle this. I can think of multiple options, but I dislike all of them:
1. remove `-W` from `SPHINXOPTS` in `doc/Makefile` to not treat warnings as errors in the whole project.
2. modify the offending file in the Docs CI jobs before calling `sphinx-build`.
3. suppressing the warning type with `suppress_warnings = ["..."]` ([docs](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings)). There is no way to suppress warnings only for specific files or lines, so this would disable the syntax check in the whole project. Also, I can't find the warning code in the error output, but I can find that out if you choose this option.

NMertsch:
Sorry for causing this mess, I didn't expect any negative side effects from the bugfix.

If you want to close this PR, just go ahead, I don't have strong feeling about this and just wanted to contribute a quick win.

If you want me to investigate further in a specific direction, please let me know and I'll see what is feasible for me.

Thanks for this library, btw! :)

NMertsch:
Sorry, messed up the git history :/ Will push again soon, please don't merge.

NMertsch:
Should be good now, please let me know if there is more to do.
Thank you for the quick and kind responses, I really appreciate it.

Byron:
Thank you, I think the PR is ready to go now!

PR #2073, opened 2025-09-23
__init__: don't run refresh on import

lheckemann:
Consumers of gitpython may not need to use it in all use cases, and may want to be able to run (without using gitpython) in environments where git is not available on PATH. While this can be worked around by setting the GIT_PYTHON_REFRESH environment variable, adding special handling for gitpython means that it can't be imported just like everything else in an import block at the top of the module, and environment variables have potentially undesired propagation behaviour.

Previously, it was also nontrivial to distinguish gitpython failing to import because of missing git or because e.g. gitpython isn't installed at all, because the exception that's raised is an ImportError without further qualification (except in the error message).

Thus, we now no longer perform `refresh` at the module top level.

---

This is _technically_ observable behaviour that could break assumptions made downstream, so I've added a changelog entry with a major version bump. I'm not sure this is "breaking enough" that you would consider it needing a major bump.

Feedback both on the breakingness and on the change in principle very welcome!

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -2,6 +2,19 @@
 Changelog
 =========
 
+4.0.0
+=====
+
+GitPython will no longer throw an ImportError when no git executable can
+be found at import time. Instead, errors are deferred until the first
+attempt at using it. Consumers with special handling for
+the old ImportError behaviour should instead call `git.refresh` and handle
+GitCommandNotFoundErrors themselves.
+
+See the following for all changes.
+https://github.com/gitpython-developers/GitPython/releases/tag/4.0.0
+
+
 3.1.45
 ======
 
--- a/git/__init__.py
+++ b/git/__init__.py
@@ -292,9 +292,4 @@ def refresh(path: Optional[PathLike] = None) -> None:
     GIT_OK = True
 
 
-try:
-    refresh()
-except Exception as _exc:
-    raise ImportError("Failed to initialize: {0}".format(_exc)) from _exc
-
 # } END initialize git executable path
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -857,7 +857,7 @@ class Git(metaclass=_GitMeta):
                     if mode in warn:
                         _logger.critical(err)
                     else:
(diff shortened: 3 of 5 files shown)


Byron (commented):
Thanks for contributing!

My intuition here is that the problem is `ImportError`, and that the change not only removes that but also the `refresh()` call which I'd think is important.

Besides that, I'd definitely be unwilling to to casually bump the major version for this alone, if it is a major bump. It could also be a fix, but I'd be very afraid it's breaking downstream in too many ways.

However, I can see that the issue to solve here is that GitPython might be imported somewhere in the dependency chain on a system that has no Git, without ever running into other issues as no Git specific features are used. Thus it's not needed to fail on import, but rather, it should fail on first use.

This is the direction I'd want to take this - can it fail later? After all, there is plenty of functionality that might work even without Git.

@EliahKagan might have a better feeling about what could be done to improve the underlying problem and avoid a breaking change.

Byron:
Setting this back to draft as I'd not want it merged just yet until it's clear if there are no better solutions to the problem.

lheckemann:
How does the behaviour of this variant look to you? The GitCommandNotFound exception is a bit awkward as is, but that should be easy enough to clean up.

Byron:
It looks like CI is failing, and I don't see my question about the lack of `refresh()` or lazy-refresh addressed.

lheckemann:
I've decided to avoid GitPython and invoke git myself in this particular use case, where a single command is enough. I still think this would be useful, but the problem is solved for my purposes.

Byron:
Thanks for sharing the workaround!

Since Eliah doesn't seem to have time for the review, this PR could hang around here for a long time. Maybe it's best to close it to at least serve people by means of offering a workaround to the problem they also might be having. I certainly don't feel confident this PR is anything we can merge, but I also don't know if maybe I am wrong about this or there is another way to fix a legitimate problem.

Byron:
Closing as the PR is conflicted and lacks interaction. Please feel free to recreate the PR if there is still interest.

Issue #1960, opened 2024-09-06
Git index version 3 support


sebastianst:
Using git python to add files to an index fails at an index head assertion if the local repo is being otherwise worked on with git `2.46.0`:
https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/index/fun.py#L209-L210

Is it planned to add support for version 3 soon?

I couldn't find any ticket tracking this, so apologies if this is a duplicate.

Byron:
There is no plan to add support for extended index flag in V3, even though it would probably be quite easy to add if someone wanted to contribute support for it.

Generally, I probably wouldn't trust the index implementation as it is here and if possible, `repo.git` might be able to do the same with a decent amount of convenience.

mgorny:
For the record, the workaround is to set:

```
git config --global feature.manyFiles false
```

and clone the repo again (or use `-c feature.manyFiles=false` when cloning it).

Byron:
--- agent

This has been implemented by merged PR #2081, “Preliminary support for index format v3,” merged on November 9, 2025. Current GitPython code accepts index versions 1–3 and writes version 3 when extended flags are present. Closing this issue as resolved.

PR #2081, opened 2025-11-02
Preliminary support for index format v3

blahgeek:
https://github.com/gitpython-developers/GitPython/issues/1960

### DIFF
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -1223,7 +1223,7 @@ class TestIndex(TestBase):
         assert index.entries
         assert index.version == 3
         assert len(index.entries) == 4
-        assert index.entries[('init.t', 0)].skip_worktree
+        assert index.entries[("init.t", 0)].skip_worktree
 
         # Write the data - it must match the original.
         with tempfile.NamedTemporaryFile() as tmpfile:


test/test_index.py
```diff
@@ -1218,6 +1218,31 @@ def test_index_add_non_normalized_path(self, rw_repo):
 
         rw_repo.index.add(non_normalized_path)
 
+    @with_rw_directory
+    def test_index_version_v3(self, tmp_dir):
+        tmp_dir = Path(tmp_dir)
+        with cwd(tmp_dir):
+            subprocess.run(["git", "init", "-q"], check=True)
```
Byron: Would it be possible to use the `Git` type instead? That way one would benefit from all the 'extras', especially around finding the right Git binary.

test/test_index.py
```diff
@@ -1218,6 +1218,31 @@ def test_index_add_non_normalized_path(self, rw_repo):
...
+            file = tmp_dir / "file.txt"
+            file.write_text("hello")
+            subprocess.run(["git", "add", "-N", "file.txt"], check=True)
+
+            repo = Repo(tmp_dir)
+
+            assert len(repo.index.entries) == 1
```
Byron: To test this, I think the `index.version` must be exposed and validated here.
I don't see at all why it would use index version 3 here.

test/test_index.py
blahgeek: OK done

PR #2086, opened 2025-11-27
Respect `os.Pathlike`

### LINKED ISSUE #2085: `Pathlike` type hint is not respected
According to [`os.pathlike`](https://docs.python.org/3/library/os.html#os.PathLike), there is a `Pathlike.__fspath__()` method. However, there are many places in the repository where this is ignored and `str()` is used instead.

George-Ogden:
Fixes #2085 
Replaces instances of `str(path)` with `path.__fspath__()` for more general usage.
I also moved the clone tests into a separate file that existed before but only contained a single test.

### DIFF
--- a/test/test_clone.py
+++ b/test/test_clone.py
@@ -53,13 +53,6 @@ class TestClone(TestBase):
         Repo.clone_from(PathLikeMock(original_repo.git_dir), PathLikeMock(os.path.join(rw_dir, "clone_pathlike")))
 
     @with_rw_directory
-    def test_clone_from_pathlike_unicode_repr(self, rw_dir):
-        original_repo = Repo.init(osp.join(rw_dir, "repo-áēñöưḩ̣"))
-        Repo.clone_from(
-            PathLikeMock(original_repo.git_dir), PathLikeMock(os.path.join(rw_dir, "clone_pathlike-áēñöưḩ̣"))
-        )
-
-    @with_rw_directory
     def test_clone_from_pathlib_withConfig(self, rw_dir):
         original_repo = Repo.init(osp.join(rw_dir, "repo"))
 
--- a/test/test_index.py
+++ b/test/test_index.py
@@ -1213,15 +1213,6 @@ class TestIndex(TestBase):
         rw_repo.index.add(PathLikeMock(str(file)))
 
     @with_rw_repo("HEAD")
-    def test_index_add_pathlike_unicode(self, rw_repo):
-        git_dir = Path(rw_repo.git_dir)
-
-        file = git_dir / "file-áēñöưḩ̣.txt"
-        file.touch()
-
-        rw_repo.index.add(PathLikeMock(str(file)))
-
-    @with_rw_repo("HEAD")
     def test_index_add_non_normalized_path(self, rw_repo):
         git_dir = Path(rw_repo.git_dir)
 


Byron:
That's a great incentive, thanks so much! Using `__fspath__()` seems like it's using a private API, so it looks strange to my untrained eye. But if it's correct, I guess it's worth doing anyway?

> Also, I notice you skip mypy because it produces errors. Why not pin the version and increment it periodically?

I don't see why this wouldn't work, great idea! Please feel free to set this up in your next PR :).

git/repo/base.py
```diff
@@ -1361,9 +1359,8 @@ def _clone(
     ) -> "Repo":
         odbt = kwargs.pop("odbt", odb_default_type)
 
-        # When pathlib.Path or other class-based path is passed
-        if not isinstance(path, str):
-            path = str(path)
+        url = os.fspath(url)
```
Byron: URL is not a path though.

Byron (changes requested):
This looks better indeed.
Something I think will be desirable is to actually add tests for non-decodable paths for the functions/types that are affected. Otherwise, how do we know it's working?

Feel free to use AI for that, it's quite good at this usually.

The idea is to prove that the changes actually make something possible that wasn't possible before.

git/repo/base.py
George-Ogden: Here, `url` can be a path if you're cloning a local repo, and if it's not `os.fspath` will leave strings alone.

Byron (commented):
That's great, thanks a lot!

What should be shown in the tests is that it can now handle filepaths that don't decode with the standard encoding.

Candidate tests are:

* clone from a filesystem path
* `index.add` with a strange path
* and probably more - you could intentionally break a piece of code that changed `fspath` and see which tests fail, then it's clear where to add a test with a 'special' path.

Thanks for making this happen, and thanks for your understanding - we can't just make changes hoping it will work or doesn't make things work, but there must be real evidence that this is desirable. And we can only have that with tests that fail without this change.

Byron:
Thanks a lot, this definitely helped me understand that these changes aren't meant to address that one thing I thought they do 😅.

My feeling is that the added tests don't actually fail on master, and thus should be removed. But if they do, please feel free to put them into a separate PR so I see them failing.
Either way, once I hear from you I think the PR can be merged.

George-Ogden:
See #2088 for failing tests

George-Ogden:
I've done that in #2089

Byron:
Thanks a lot for your patience, just one more thing: 171062655e24b6a6ca1a3beab3c7679278350ab5 adds tests which I believe should have shown that they won't work on `master`, but work here. The idea is that the encoding of the files can't be determined if Python would try to decode them, which a lot of code here does.

If you don't feel like this has a chance to be fixed, then I think these tests can be removed by force-pushing without this commit for this PR to be merged. Otherwise, there can be a PR on `master` that shows how these two tests are failing because of some decoding issue, and this PR will be merged as well.

Thank you, and… we will get there :).

George-Ogden:
However, both
https://github.com/gitpython-developers/GitPython/blob/171062655e24b6a6ca1a3beab3c7679278350ab5/test/test_clone.py#L86-L98
and 
https://github.com/gitpython-developers/GitPython/blob/0fcf291d6fc5e2f6212b86d5824704cee69403f3/test/test_index.py#L985-L994
Ran on `main` successfully (and were included before this PR)

The difference in 
https://github.com/gitpython-developers/GitPython/blob/171062655e24b6a6ca1a3beab3c7679278350ab5/test/test_index.py#L1216-L1222
which fails on `main` is that the path is wrapped in a `PathLikeMock`, which has an interface similar to the one described above:
https://github.com/gitpython-developers/GitPython/blob/171062655e24b6a6ca1a3beab3c7679278350ab5/test/lib/helper.py#L55-L60

Byron:
I see, thanks for clarifying and digging out the non-unicode tests on `main` which have been there before and seemed to be working already.

Then I even more so think that the tests in https://github.com/gitpython-developers/GitPython/commit/171062655e24b6a6ca1a3beab3c7679278350ab5 as duplicates of the tests above serve no purpose. The `PathLike` test is already done.
What am I missing?

George-Ogden:
All reverted! I agree that the tests don't add much.

PR #2089, opened 2025-12-02
Failing Pathlike Tests

George-Ogden:
Demonstrates changes in #2086


Byron:
Thanks for trying. I really don't get why 9 tests are failing even though two tests were added in the most recent commit in #2086.
But it's OK.

George-Ogden:
@Byron, this is all the tests I added or modified - I think 9 is the correct number

PR #2092, opened 2025-12-07
Fix diff parsing to support mnemonicPrefix configuration

### LINKED ISSUE #2013: gitconfig: mnemonicPrefix conflict with create_patch
the regex:

https://github.com/gitpython-developers/GitPython/blob/6d09bb610932d6081da7a516f8e7215414d953e5/git/diff.py#L372

is not able to match the diff generated if this flag is enabled in `~/.gitconfig`:
```
[diff]
        mnemonicPrefix = true
```
because it is expecting `[ab]/` as folder prefix for the diff but with mnemonicPrefix enabled it become `[wc]/`

### Diff without `mnemonicPrefix`
```diff
diff --git a/.vscode/launch.json b/.vscode/launch.json
index xxxxxxx..xxxxxxxx
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
```

### Diff with `mnemonicPrefix = true` 
```diff
diff --git c/.vscode/launch.json w/.vscode/launch.json
index xxxxxxx..xxxxxxxx
--- c/.vscode/launch.json
+++ w/.vscode/launch.json
```

### Quick Test
```python
repo = git.Repo('.')
branch_ref = repo.head
ancestor_ref = repo.merge_base(branch_ref, repo.refs["main"])
diff = ancestor_ref[0].diff(None, create_patch=True)
print(diff)
```

MirrorDNA-Reflection-Protocol:
## Summary

Fixes #2013

When `diff.mnemonicPrefix=true` is set in git config, git uses different prefixes for diff paths instead of the standard `a/` and `b/`:
- `c/` for commit
- `w/` for worktree
- `i/` for index
- `o/` for object
- `h/` for HEAD

Previously, the diff regex and `decode_path()` function only accepted `a/` and `b/` prefixes, causing `create_patch=True` diffs to fail parsing when users had this config enabled.

## Reproduction

As described in #2013:
```python
# With ~/.gitconfig containing:
# [diff]
#     mnemonicPrefix = true

repo = git.Repo('.')
ancestor_ref = repo.merge_base(repo.head, repo.refs['main'])
diff = ancestor_ref[0].diff(None, create_patch=True)  # Would fail to parse
```

## Changes

- Update `re_header` regex to accept `[abciwoh]/` prefixes instead of just `[ab]/`
- Update `decode_path()` assertion to accept all valid mnemonic prefixes
- Add test case for mnemonicPrefix-style diffs

## Testing

- New test `test_diff_mnemonic_prefix` verifies parsing works with `c/` and `w/` prefixes
- All existing diff tests pass (except one pre-existing failure on main unrelated to this change)

## References

- Git documentation: https://git-scm.com/docs/git-config#Documentation/git-config.txt-diffmnemonicPrefix
- Issue: #2013

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -371,7 +371,7 @@ class Diff:
 
     # Precompiled regex.
     # Note: The path prefixes support both default (a/b) and mnemonicPrefix mode
-    # which can use prefixes like c/ (commit), w/ (worktree), i/ (index), o/ (object)
+    # which can use prefixes like c/ (commit), w/ (worktree), i/ (index), o/ (object), and h/ (HEAD)
     re_header = re.compile(
         rb"""
                                 ^diff[ ]--git
--- a/git/diff.py
+++ b/git/diff.py
@@ -116,7 +116,7 @@ def decode_path(path: bytes, has_ab_prefix: bool = True) -> Optional[bytes]:
         # Support standard (a/b) and mnemonicPrefix (c/w/i/o/h) prefixes
         # See git-config diff.mnemonicPrefix documentation
         valid_prefixes = (b"a/", b"b/", b"c/", b"w/", b"i/", b"o/", b"h/")
-        assert any(path.startswith(p) for p in valid_prefixes), f"Unexpected path prefix: {path[:10]}"
+        assert any(path.startswith(p) for p in valid_prefixes), f"Unexpected path prefix: {path[:10]!r}"
         path = path[2:]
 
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
Thanks a lot for contributing! I particularly like the exhaustive documentation around the matter, and what I see looks good!

The test is also something I'd love to see improved, and I side with Copilot here.

PR #2098, opened 2026-01-10
DOC: README Add urls and updated a relative url

### LINKED ISSUE #2097: DOC: README Not present or not working urls
## Problem
On [README](https://github.com/gitpython-developers/GitPython/blob/main/README.md) there are some written files like `./requirements.txt`, `./test-requirements.txt`, `./pyproject.toml`... without a url that points directly to themselves.
There is also a file where there is a link but it is a relative url, making it not work on PyPI

## Reproduction of the problem
For the first case: go on [PyPI page](https://pypi.org/project/GitPython/) or [Github repository](https://github.com/gitpython-developers/GitPython), and see for the examples I brought up that they have no url.

For the second case: go on [PyPI page](https://pypi.org/project/GitPython/), on LICENCE section and press on the clickable text "a separate license, detailed here", the relative link goes in a [404 page](https://pypi.org/project/GitPython/fuzzing/README.md#license).

Image
Image

Timour-Ilyas:
Fix #2097

## Solution
Added a url to the text that refers to a specific file in the repository: now each one have direct url, before only part of the quoted files had.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ by setting the `GIT_PYTHON_GIT_EXECUTABLE=<path/to/git>` environment variable.
 - Git (1.7.x or newer)
 - Python >= 3.7
 
-The list of dependencies are listed in `./requirements.txt` and `./test-requirements.txt`.
+The list of dependencies are listed in [`./requirements.txt`](https://github.com/gitpython-developers/GitPython/blob/main/requirements.txt) and [`./test-requirements.txt`](https://github.com/gitpython-developers/GitPython/blob/main/test-requirements.txt).
 The installer takes care of installing them for you.
 
 ### INSTALL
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,7 +46,7 @@ by setting the `GIT_PYTHON_GIT_EXECUTABLE=<path/to/git>` environment variable.
 - Git (1.7.x or newer)
 - Python >= 3.7
 
-The list of dependencies are listed in `./requirements.txt` and `./test-requirements.txt`.
+The list of dependencies are listed in [`./requirements.txt`](https://github.com/gitpython-developers/GitPython/blob/main/requirements.txt) and [`./test-requirements.txt`](https://github.com/gitpython-developers/GitPython/blob/main/test-requirements.txt).
 The installer takes care of installing them for you.
 
 ### INSTALL
```
DausNono: https://source.android.com/docs/setup/download/troubleshoot-sync#configure-proxy-use

PR #2118, opened 2026-03-26
Add `Commit.patch` property to expose commit-to-parent diff text

### LINKED ISSUE #1413: Get patch text under stats from Commit class
Current we can get status for the commit (comparing parent). But can't get patch text.

I would like to have commit.patch to get patch text. Result will be similar to git diff shell command.

```shell
git diff <c1> <c2>
```

```python
all_commits = repo.iter_commits('master')
for commit in all_commits:
    print(commit.stats)
```

If the community is okay with this addition, i can work on this and raise a PR.

Thanks in advance,
Durai Pandian

huyhoang171106:
## Summary

The issue is a missing API feature, not a runtime defect: `Commit` exposes structured diff stats (`commit.stats`) but does not expose the raw patch text for the same comparison. Users currently must manually call lower-level git commands for each commit. The `Commit` class should be extended with a `patch` property (or method) that returns unified diff text against its parent, with correct handling for merge/root commits.

## Files changed

- `git/objects/commit.py` (modified)

## Testing

Closes #1413

### DIFF
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -196,6 +196,13 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         if gpgsig is not None:
             self.gpgsig = gpgsig
 
+    @property
+    def patch(self) -> str:
+        """Textual patch comparing this commit against its first parent."""
+        if not self.parents:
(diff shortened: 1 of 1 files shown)


Byron:
Quick shot == insta-close.

PR #2119, opened 2026-03-26
Add a first-class API to get the latest tag from a remote via `git ls-remote`

### LINKED ISSUE #1071: Getting the latest tag of a remote repo
The following is from a question on [StackOverflow](https://stackoverflow.com/questions/61613647/gitpython-getting-the-latest-tag-of-a-remote-repo/64423285#64423285) I've asked some time ago:

I'd like to be able to check what's the latest tag for a given repo (I'm using CPython here as an example).

The following works:
```python
g = git.cmd.Git()
blob = g.ls_remote('https://github.com/python/cpython', sort='-v:refname', tags=True)
blob.split('\n')[0].split('/')[-1] 
```
and yields 'v3.9.0a6' because the blob looks something like this:
```
'bc1c8af8ef2563802767404c78c8ec6d6a967897\trefs/tags/v3.9.0a6\ndcd4 (...)'
```

Someone suggested I can contribute this solution to `gitpython` and someone else suggested that the `git ls-remote` is pretty stable and well documented so parsing its output might be the right way to go when trying to get the latest tag of a remote repo.

**My question:** would you like me to contribute this type of functionality? And if so, I'd appreciate a hint on where should it be / how it should be named / anything else.

huyhoang171106:
## Summary

Introduce a dedicated convenience method on `Git` for the issue use case (querying a remote URL/repo and returning its newest tag). This avoids requiring users to manually run `ls_remote(...)` and parse tab-delimited output themselves. The method should encapsulate stable parsing behavior and edge-case handling (empty output, annotated tag peeled refs, malformed lines).

## Files changed

- `git/cmd.py` (modified)

## Testing

Closes #1071

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -995,6 +995,28 @@ class Git(metaclass=_GitMeta):
             _warn_use_shell(extra_danger=False)
         return super().__getattribute__(name)
 
+    def latest_remote_tag(self, repository: PathLike) -> Optional[str]:
+        output = self.ls_remote("--tags", "--sort=-version:refname", repository)
+        if not output:
+            return None
+
+        for line in output.splitlines():
+            if not line:
+                continue
+            try:
+                _, ref = line.split("\t", 1)
+            except ValueError:
+                continue
+            if not ref.startswith("refs/tags/"):
+                continue
+            tag = ref[len("refs/tags/") :]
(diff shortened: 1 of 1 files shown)


Byron:
Quick shot == insta-close.

PR #2121, opened 2026-03-26
Add missing `Repo.tracked_files` convenience property

### LINKED ISSUE #857: repo.tracked_files
https://gitpython.readthedocs.io/en/stable/tutorial.html tells that there is `repo.untracked_files` to get list of untracked files.

therefor, there should be `repo.tracked_files` to get list of tracked files.

ps. how do we even do it now? https://gitpython.readthedocs.io/en/stable/reference.html has no mention of "tracked files"...

huyhoang171106:
## Summary

The issue is not a runtime defect, but an API/documentation gap: `Repo` exposes `untracked_files` but has no symmetric `tracked_files` property, which makes common usage inconsistent and discoverability poor. Users currently need to know and call plumbing manually (e.g., `repo.git.ls_files()`), and this is not clearly documented in the API reference.

## Files changed

- `git/repo/base.py` (modified)
- `README.md` (modified)

## Testing

Closes #857

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ or low-level like git-plumbing.
 It provides abstractions of git objects for easy access of repository data often backed by calling the `git`
 command-line program.
 
+For file discovery convenience, repositories expose both `Repo.untracked_files` and `Repo.tracked_files`.
+
 ### DEVELOPMENT STATUS
 
 This project is in **maintenance mode**, which means that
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
Besides what Copilot mentions, returning `str` would be wrong here as the OS encoding of paths doesn't typically match the Python interpreter encoding of strings.

Byron:
Oh, I realize my mistake, no need to try here.

PR #2123, opened 2026-04-11
config: resolve relative includes from relative config paths

### LINKED ISSUE #2103: Relative path include in .gitconfig throws AssertionError
My `.gitconfig` contains:

```
[include]
    path = .gitconfig_common
```

According to the [Git documentation](https://git-scm.com/docs/git-config#_includes), this is allowed to be a _relative path_:

> If the value of the variable is a relative path, the path is considered to be relative to the configuration file in which the include directive was found.

However, GitPython throws:

```
  File "/Users/daniel/.cache/uv/archive-v0/YvUvYFO1C0vkI8VZmPICW/lib/python3.12/site-packages/briefcase/commands/base.py", line 1365, in get_git_config_value
    if git_config.has_option(section, option):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel/.cache/uv/archive-v0/YvUvYFO1C0vkI8VZmPICW/lib/python3.12/site-packages/git/config.py", line 113, in assure_data_present
    self.read()
  File "/Users/daniel/.cache/uv/archive-v0/YvUvYFO1C0vkI8VZmPICW/lib/python3.12/site-packages/git/config.py", line 654, in read
    assert osp.isabs(file_path), "Need absolute paths to be sure our cycle checks will work"
           ^^^^^^^^^^^^^^^^^^^^
AssertionError: Need absolute paths to be sure our cycle checks will work
```

The workaround is to modify my `.gitconfig` and make the path absolute, but I'd like it to work without modifying my `.gitconfig` to suit GitPython.

officialasishkumar:
Fixes #2103

## Summary

- Normalize path-based config inputs before using them for include cycle detection.
- Resolve relative `include.path` entries from that normalized source path when the root config file was supplied as a relative path.
- Add regression coverage with a relative root config and a relative include cycle.

## Testing

- `.venv/bin/pytest --no-cov -q test/test_config.py`
- `.venv/bin/pytest --no-cov -q test/test_config.py test/test_repo.py::TestRepo::test_config_reader`
- `.venv/bin/pytest --no-cov -q test/test_refs.py::TestRefs::test_set_tracking_branch_with_import`
- `.venv/bin/ruff check git/config.py test/test_config.py`
- `.venv/bin/ruff format --check git/config.py test/test_config.py`
- `.venv/bin/mypy git/config.py`
- `git diff --check`

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -247,6 +247,24 @@ class TestBase(TestCase):
             check_test_value(cr, tv)
 
     @with_rw_directory
+    def test_config_include_from_relative_config_path(self, rw_dir):
+        fpa = osp.join(rw_dir, "a")
+        fpb = osp.join(rw_dir, "b")
+
+        with GitConfigParser(fpa, read_only=False) as cw:
+            cw.set_value("a", "value", "a")
+            cw.set_value("include", "path", "b")
+
+        with GitConfigParser(fpb, read_only=False) as cw:
+            cw.set_value("b", "value", "b")
+            cw.set_value("include", "path", "a")
+
+        with GitConfigParser(osp.relpath(fpa), read_only=True) as cr:
+            assert cr.get_value("a", "value") == "a"
+            assert cr.get_value("b", "value") == "b"
+            assert cr.get_values("include", "path") == ["b", "a"]
+
+    @with_rw_directory
     def test_multiple_include_paths_with_same_key(self, rw_dir):
         """Test that multiple 'path' entries under [include] are all respected.
 
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -249,20 +249,22 @@ class TestBase(TestCase):
     @with_rw_directory
     def test_config_include_from_relative_config_path(self, rw_dir):
         fpa = osp.join(rw_dir, "a")
-        fpb = osp.join(rw_dir, "b")
+        nested_dir = osp.join(rw_dir, "subdir")
+        os.makedirs(nested_dir)
+        fpb = osp.join(nested_dir, "b")
 
         with GitConfigParser(fpa, read_only=False) as cw:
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
It looks like Windows has problems, but that might point to some logic error.
There should also be a test that actually validates that cycle-detection, by putting one of the include files in a directory.

officialasishkumar:
@Byron Added the nested-directory include cycle coverage in 483f0c66: the relative root config now includes `subdir/b`, and that nested file includes `../a`. Validation passed with `test/test_config.py`, Ruff, mypy on `git/config.py`, and `git diff --check`.

test/test_config.py
```diff
@@ -246,6 +246,26 @@ def check_test_value(cr, value):
         with GitConfigParser(fpa, read_only=True) as cr:
             check_test_value(cr, tv)
 
+    @with_rw_directory
+    def test_config_include_from_relative_config_path(self, rw_dir):
+        fpa = osp.join(rw_dir, "a")
+        nested_dir = osp.join(rw_dir, "subdir")
+        os.makedirs(nested_dir)
+        fpb = osp.join(nested_dir, "b")
```
Byron, line 254: One of the included files need to be in a subdirectory, and both need to be named "a" so the relative name is the same for an actual cycle check.

Byron (changes requested):
Windows still fails, also the cycle check isn't addressed. It seems like this is a bot account and I don't want to spend more time here.
Will close if the PR isn't green and has its issues addressed next time it comes out of draft.

Byron:
Closing, as it's stale. I vaguely remember that relative includes also already landed by means of another contribution.

PR #2129, opened 2026-04-20
feat(submodule): add deinit method to Submodule (#2014)

### LINKED ISSUE #2014: `git submodule deinit`
## Issue

Currently there is no direct way to run `git submodule deinit` in gitpython. My workaround is using `repo.git.submodule('deinit', ...)` to access git sm directly.

## Expected:

Like remove, add, update, I expected to find deinit present for `git.Submodule` as well.

mvanhorn:
Closes #2014.

Adds a \`Submodule.deinit(force=False)\` method that wraps \`git submodule deinit [--force] -- <path>\`, paralleling the existing \`add\` / \`update\` / \`remove\` methods. Callers no longer need the \`repo.git.submodule('deinit', ...)\` workaround mentioned in the issue.

## Behavior

- Delegates to \`self.repo.git.submodule('deinit', ...)\`, so output, error handling, and subprocess invocation match every other GitPython submodule method.
- Decorated with \`@unbare_repo\` to match \`remove\` and \`add\` — deinit is meaningless in bare repos.
- Leaves \`.gitmodules\` and \`.git/modules/<name>\` intact. A subsequent \`sm.update()\` re-initializes from those retained contents, which matches the git CLI semantics.
- \`force=True\` passes \`--force\` so callers can deinit submodules with local modifications (same semantics as \`git submodule deinit -f\`).

## Why a thin wrapper

The full semantics of \`git submodule deinit\` (recursive children, working-tree checks, .git/modules retention) are already handled correctly by the git porcelain. Re-implementing that logic in Python for parity with the existing \`remove\` method would duplicate significant behavior. A thin wrapper gives callers a typed entry point without duplicating git's state machine.

## Verification

- \`ruff check git/objects/submodule/base.py\` → clean.
- \`mypy git/objects/submodule/base.py\` → no new errors (the file has pre-existing mypy warnings unrelated to this change).
- Imported the class after install: \`Submodule.deinit\` is present with the expected docstring.
- Tests not added in this PR — the existing \`test/test_submodule.py\` suite couldn't run cleanly against this checkout because the \`git/ext/gitdb/gitdb/ext/smmap\` submodule fixture expects \`git submodule update --init --recursive\` to have completed on the clone, which wasn't the case here. Happy to add a targeted test if you can point me at the right fixture setup, or tell me if you'd rather land this as a wrapper-only change first.

🤖 AI-assisted.

### DIFF
--- a/git/objects/submodule/base.py
+++ b/git/objects/submodule/base.py
@@ -49,6 +49,7 @@ from typing import (
     Callable,
     Dict,
     Iterator,
+    List,
     Mapping,
     Sequence,
     TYPE_CHECKING,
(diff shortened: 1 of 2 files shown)


mvanhorn:
Added `test_deinit_calls_git_submodule` in 8e24b0f -- mocks `Git.submodule` and asserts both arg orderings:
- `deinit()` (default) calls with `("deinit", "--", path)`
- `deinit(force=True)` calls with `("deinit", "--force", "--", path)`

Verified locally with `pytest test/test_submodule.py -k deinit -v` (passed) plus `ruff check` and `ruff format --check` clean. Marking the PR ready for human review.

Byron:
I set this back to draft while CI fails.

mvanhorn:
Fixed in 7b0e1ae. It was a mypy error, not a test failure: `self.path` is `str | PathLike[str]` but the deinit args list is `List[str]`, so mypy flagged base.py:1297. Wrapped the path in `str()` (git submodule expects a string path anyway). Verified locally - mypy clean across all 44 files and the deinit test still passes.

Byron:
Thanks a lot!

mvanhorn:
Appreciate the merge, @Byron - deinit rounds out the submodule lifecycle nicely alongside add.

PR #2132, opened 2026-04-21
refactor(git/repo/base): use less strict types in is_ancestor

comeillfoo:
Changes the type hints in the is_ancestor arguments to avoid warnings from type checkers (e.g. Pylance).

### DIFF
--- a/git/repo/base.py
+++ b/git/repo/base.py
@@ -824,7 +824,7 @@ class Repo:
 
         return res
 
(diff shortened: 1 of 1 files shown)


Byron:
Thanks, but is there a way to type this properly?

The warnings this causes are also unspecific enough to make this more of a local fix than anything that should be merged.

Closing, but feel free to reopen if a more thorough fix is possible and desired.

PR #2144, opened 2026-05-07
docs(cmd): clarify Git.execute() string vs list command argument

### LINKED ISSUE #2016: [Bug] GitCommandNotFound when executing repo.git.execute on macOS
**Describe the bug**
When executing repo.git.execute with string-type arguments on macOS, it throws a GitCommandNotFound exception:
```
  File "/Users/macuser/.pyenv/versions/3.12.2/lib/python3.12/site-packages/git/cmd.py", line 1262, in execute
    proc = safer_popen(
           ^^^^^^^^^^^^
  File "/Users/macuser/.pyenv/versions/3.12.2/lib/python3.12/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Users/macuser/.pyenv/versions/3.12.2/lib/python3.12/subprocess.py", line 1953, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git log -n 1'

The above exception was the direct cause of the following exception:

  File "/Users/macuser/test.py", line 36, in <module>
    cmd = repo.git.execute("git log -n 1")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/macuser/.pyenv/versions/3.12.2/lib/python3.12/site-packages/git/cmd.py", line 1275, in execute
    raise GitCommandNotFound(redacted_command, err) from err
git.exc.GitCommandNotFound: Cmd('g') not found due to: FileNotFoundError('[Errno 2] No such file or directory: 'git log -n 1'')
```
I have installed git and it works well on other commands, only repo.git.execute doesn't work'

The error only occurs when passing a string-type argument to repo.git.execute

Works correctly when using string array arguments

Works correctly when using shell=True argument

for example:

```python
# Failing case (string argument)
repo.git.execute("git log -n 1")  # Throws exception

# Working case (array argument)
repo.git.execute(["git", "log", "-n", "1"])  # Executes successfully

# Working case (array argument)
repo.git.execute("git log -n 1", shell=True)  # Executes successfully
```

**Environment**
- Python version: 3.12
- GitPython version: 3.1.43
- Operating System: macOS 14.3.1 (Sonoma)
- Git version: 2.46.0
- By the way, I'm using pyenv

mvanhorn:
## What changed

Rewrites the `:param command:` docstring on `Git.execute()` in `git/cmd.py` to clarify the string-vs-sequence semantics. The previous wording ("the program to execute is the first item in the args sequence or string") was misleading: with `shell=False` (the default), `subprocess.Popen` treats the entire string as a single executable name, so `"git log -n 1"` looks for an executable literally named "git log -n 1" and raises `GitCommandNotFound`.

The new docstring:
- Recommends the sequence form (`["git", "log", "-n", "1"]`) for the default `shell=False` case.
- Documents that a string is accepted but is passed as a single executable name when `shell=False`, with the exact failure mode users see in #2016.
- Points users at `shlex.split(...)` (sequence) or `shell=True` (with the existing warning) for the case where they want a string that gets tokenised.

Closes #2016

## Why

This is the third recurrence of this confusion that surfaces in issues — string commands look like the obvious shape for "run this git invocation", and the failure mode is opaque. The `subprocess.Popen` semantics are doing exactly what they always do; the GitPython docstring was the surface that misled users into expecting otherwise. Updating it costs ~10 lines and short-circuits the next round of issues.

I deliberately kept this docs-only. Auto-splitting a string when it contains spaces would be a behavior change that could break existing callers who do pass a single executable path with whitespace. The acknowledged label on the issue suggests maintainer interest without committing to a specific behavior fix; clarifying the docs is the smallest useful step.

## Verification

`python -c "import git; help(git.cmd.Git.execute)"` renders the new param block cleanly. `python -m py_compile git/cmd.py` is clean. No tests were modified or broken.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1119,9 +1119,16 @@ class Git(metaclass=_GitMeta):
         information (stdout).
 
         :param command:
-            The command argument list to execute.
-            It should be a sequence of program arguments, or a string. The
-            program to execute is the first item in the args sequence or string.
+            The command to execute. A sequence of program arguments is the
+            recommended form when `shell` is ``False`` (the default), e.g.
+            ``["git", "log", "-n", "1"]``.
+
+            A string is accepted, but with `shell` set to ``False`` it is passed
+            as a single executable name to :class:`subprocess.Popen`. For example,
+            ``"git log -n 1"`` looks for an executable literally named
+            ``git log -n 1`` and will fail with :class:`GitCommandNotFound`. To
+            split a command string into argv tokens, pass ``shlex.split(...)`` as
+            a sequence or set `shell` to ``True`` (see the warning below).
 
         :param istream:
             Standard input filehandle passed to :class:`subprocess.Popen`.


Byron (approved):
LGTM, thanks!

PR #2152, opened 2026-05-12
Git test environment consistency

elovelan:
When implementing #2151, I ran into a few issues running tests due to:

1. By default, GitHub now forks only the default branch when forking a repo. One of the docs tests was relying on `origin/master` existing, which is unlikely in a fork scenario, since the common pattern is to set `origin` to the fork and `upstream` to the primary repo.
2. I have a few "convenience" features in my configuration such as `main` as the default repo branch name and `fetch.prune=true`. These break some of the tests so it seemed like it might be a good idea to always use a consistent config file.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,10 @@
+import os
+from test.lib import fixture_path
+
+# ensure all tests run with consistent config since some settings break tests
+# we could also use pytest-env for this but it's less consistent across OSs
+
+os.environ["GIT_CONFIG_NOSYSTEM"] = "true"
+os.environ["GIT_CONFIG_GLOBAL"] = fixture_path("git_config_defaults")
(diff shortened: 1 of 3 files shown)


elovelan:
I'll spin up a Windows VM for testing!

elovelan:
@Byron, this is actually a little bigger than I had anticipated since there's a bit of config manipulation in the Actions workflows. If you like this approach, I'm happy to bring those settings over but I wanted to check with you first!

Byron:
The test suite unfortunately has never been isolated, and this bites from time to time.
Let's make the changes that are needed on CI to make it more resilient. We just have to watch out that these alterations don't affect local runs.

elovelan:
Since I need to rethink this, I'm going to close this for now, so it's not noise in the PR list

PR #2153, opened 2026-05-15
Defer xfail condition evaluation with xfail_if_raises context manager

elovelan:
## Summary

This PR introduces `xfail_if_raises`, a context manager that approximates `@pytest.mark.xfail(..., raises=...)` behavior inside a test body rather than as a decorator.

### Problem

`@pytest.mark.xfail` evaluates its `condition` argument during test **collection** (module import time). In `test_index_mutation`, this condition includes `Git().config("core.symlinks") == "true"`, which:

1. Spawns an external `git config` process even when the test is not being run
2. Can fail or behave unexpectedly during collection if git configuration is unavailable
3. Relies on in-repo code during test discovery, which may not be appropriate in all contexts

### Solution

`xfail_if_raises` moves the xfail logic into the test body. The side-effectful condition is only evaluated if the test actually raises the specified exception(s). This keeps test collection lightweight and avoids relying on in-repo code as part of module import/test discovery.

This change is applied to `test_index_mutation` in `test/test_index.py`.

### AI disclosure

This PR description was prepared with assistance from an agent. All code was written by hand.

### DIFF
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -18,6 +18,7 @@ __all__ = [
     "skipIf",
     "GIT_REPO",
     "GIT_DAEMON_PORT",
+    "xfail_if_raises",
 ]
 
 import contextlib
@@ -35,8 +36,10 @@ import textwrap
 import time
 import unittest
 import venv
+from typing import Union, Type, Tuple
 
 import gitdb
+import pytest
 
 from git.util import rmtree, cwd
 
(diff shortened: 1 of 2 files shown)


elovelan:
I didn't change anything that should have affected this one failed test on Cygwin. I have a Windows VM now so I'll see if I can repro.

EliahKagan:
The failed Cygwin test looks like a known flake with submodule ownership on Cygwin CI that I've been working on and hope to have a PR in for soon. To the best of my knowledge, that is independent of the changes here, though I emphasize that I have not reviewed this PR. I've triggered the test to rerun--my guess is that it will pass.

Byron (approved):

Image

Initially I was confused about `strict` as it's not actually used, but it's Ok and nothing one couldn't figure out and address next time someone gets confused by it.

elovelan:
> Initially I was confused about strict as it's not actually used, but it's Ok and nothing one couldn't figure out and address next time someone gets confused by it.

I debated a YAGNI approach, but I figured as a utility method, I should try to mirror the decorator as closely as I could.

Issue #2025, opened 2025-05-11
`IndexFile.diff(None)` returns empty after `init -> add -> write -> read` sequence on a new repository


ElJaviLuki:
**Environment:**
*   GitPython version: 3.1.44o
*   Git version: git version 2.42.0.windows.2
*   Python version: 3.12.0
*   Operating System: Windows 11 Pro 24H2 26100.3775

**Description:**
When initializing a new repository, adding a file to the index, writing the index to disk, and then explicitly reading the index back, a subsequent call to `repo.index.diff(None)` incorrectly returns an empty `DiffIndex` (an empty list). This occurs even though an external `git status --porcelain` command correctly shows the file as added to the index (stage 'A').

This suggests that the in-memory state of the `IndexFile` object is not correctly reflecting the on-disk state for the `diff(None)` operation under these specific circumstances, even after an explicit `repo.index.read()`.

**Steps to Reproduce:**
```python
import os
import tempfile
import shutil
from git import Repo, IndexFile, Actor

# Setup a temporary directory for the new repository
repo_dir = tempfile.mkdtemp(prefix="test_gitpython_index_issue_")
try:
    # 1. Initialize a new repository
    repo = Repo.init(repo_dir)
    print(f"Repository initialized at: {repo_dir}")
    print(f"Is bare: {repo.bare}") # Should be False

    # 2. Create and add a new file (.gitkeep in this example)
    gitkeep_path = os.path.join(repo.working_tree_dir, ".gitkeep")
    with open(gitkeep_path, 'w') as f:
        f.write("# Initial file\n")
    print(f".gitkeep created at: {gitkeep_path}")

    index = repo.index
    index.add([".gitkeep"]) # Relative path to repo root
    print(f"Added '.gitkeep' to index object in memory.")

    # 3. Write the index to disk
    index.write()
    print(f"Index written to disk at: {index.path}")
    assert os.path.exists(index.path), "Index file should exist on disk"

    # 4. (Optional but good for verification) Check with external git status
    status_output = repo.git.status(porcelain=True)
    print(f"git status --porcelain output: '{status_output}'")
    assert "A  .gitkeep" in status_output or "?? .gitkeep" in status_output # Should be 'A ' after add+write

    # 5. Explicitly re-read the index (or create a new IndexFile instance)
    #    This step is crucial to the bug demonstration.
    index.read() # Force re-read of the IndexFile instance
[...]

PR #2155, opened 2026-05-21
Support index diffs against the empty tree

puneetdixit200:
Related to #2025.

`IndexFile.diff()` is documented as accepting the same targets as `Diffable.diff()`, including `NULL_TREE`, but the index-specific implementation rejected `NULL_TREE` before it could produce a diff. That leaves no direct API path for inspecting staged entries for an initial commit against the empty tree.

This adds explicit `IndexFile.diff(NULL_TREE)` handling using `git diff --cached` against the empty tree object. I kept `IndexFile.diff(None)` as the existing index-vs-working-tree comparison, since that behavior is documented in the quick docs and existing tests.

The regression test creates a new repository, stages a file, writes the index, constructs a fresh `IndexFile`, and verifies that `index.diff(NULL_TREE)` reports the staged addition while `index.diff(None)` remains empty because the working tree matches the index.

Validation:
- `.venv/bin/python -m pytest --no-cov test/test_index.py::TestIndex::test_index_file_diff_null_tree_with_initial_index -q`
- `.venv/bin/python -m ruff check git/index/base.py test/test_index.py`

AI assistance disclosure: OpenAI GPT-5 helped draft and check this change; I reviewed the implementation and tests before submitting.

### DIFF
--- a/git/diff.py
+++ b/git/diff.py
@@ -3,7 +3,7 @@
 # This module is part of GitPython and is released under the
 # 3-Clause BSD License: https://opensource.org/license/bsd-3-clause/
 
-__all__ = ["DiffConstants", "NULL_TREE", "INDEX", "Diffable", "DiffIndex", "Diff"]
+__all__ = ["DiffConstants", "NULL_TREE", "NULL_TREE_SHA", "INDEX", "Diffable", "DiffIndex", "Diff"]
 
 import enum
 import re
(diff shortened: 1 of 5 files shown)


puneetdixit200:
AI agent on behalf of Puneet: pushed `fe4b66d` to use a shared empty-tree SHA constant, accept that SHA for index diffs, and make the existing empty-index assertion explicit.

Verified:
- `.venv/bin/python -m pytest test/test_index.py::TestIndex::test_index_file_diff_null_tree_with_initial_index -q`
- `.venv/bin/python -m ruff check git/diff.py git/index/base.py test/test_index.py`
- `git diff --check`

Byron:
It seems bots are running around and fixing to dos automatically. And in a way, it's fine. This seems to be addressing the TODO legitimately and the bots already hashed it out, so I'll just try rolling with it.

The biggest problem GitPython has is that I don't think it has a human who knows what good Python is supposed to be like these days. So, agents will basically rule this place and all I can do is to apply high level quality constraints - better than nothing.

PR #2157, opened 2026-05-27
refactor: seperate out Progress type

LoeschMaximilian:
I encountered a problem when using mypy in a project, that the documentation of the types of objects accepted by the progress parameter was not consistent with the documentation of `Remote.pull()`


Byron (approved):
Thanks!

PR #2167, opened 2026-07-05
Add Commit.is_shallow property; document stats() limitation at shallow boundary

harshitayadavv:
## Problem

Accessing `.stats` on a commit at the boundary of a shallow clone raises
`git.exc.GitCommandError: fatal: bad object <sha>`. This happens because
the commit object itself still references its parent's SHA, but that
parent was never fetched — it's beyond the shallow depth. There's currently
no way to detect this ahead of time without hitting the crash.

## Reproduction

    git clone --depth 5 <any repo with more history> repo
    cd repo
    python -c "
    import git
    repo = git.Repo('.')
    oldest = list(repo.iter_commits())[-1]
    print(oldest.parents)   # non-empty -- references a SHA that doesn't exist locally
    oldest.stats            # raises GitCommandError: fatal: bad object <sha>
    "

## Fix

Adds a `Commit.is_shallow` property that checks whether the commit's hexsha
appears in the repository's `shallow` file -- a cheap, local check with no
git subprocess call. Also documents this limitation in the `stats`
docstring, pointing to `is_shallow`.

Includes a test (`test_is_shallow`) that reproduces the failure via a real
local shallow clone (using `no_local=True`, since Git ignores `--depth` for
local-optimized clones) and confirms both that the boundary commit reports
`is_shallow == True` and that `.stats` raises as expected on it.

Ran the full test suite locally (Windows, Python 3.11): 660 passed, 9
unrelated pre-existing failures (Windows symlink privilege requirements,
a missing-remote assumption in one tutorial test, and PATH resolution in
shell-impostor tests) -- none touching commit.py or stats.

## AI disclosure

Per CONTRIBUTING.md: I used Claude to help design, implement, and test
this change. I reviewed, ran, and verified every step myself, including
reproducing the underlying bug firsthand before writing the fix.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,5 +57,6 @@ Contributors are:
 -Jonas Scharpf <jonas.scharpf _at_ checkmk.com>
 -Gordon Marx
 -Enji Cooper
+-Harshita Yadav <harshitayadav504 _at_ gmail.com>
 
 Portions derived from other open source works and are clearly marked.
--- a/git/objects/commit.py
+++ b/git/objects/commit.py
@@ -368,6 +368,7 @@ class Commit(base.Object, TraversableIterableObj, Diffable, Serializable):
         kwargs["skip"] = skip
 
         return self.iter_items(self.repo, self, paths, **kwargs)
+
     @property
     def is_shallow(self) -> bool:
         """Check whether this commit is a shallow boundary (graft) commit.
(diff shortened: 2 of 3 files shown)


Byron (changes requested):
Thanks for giving this a shot.

Can we reduce this to just the documentation change for a quick merge? Otherwise, the `.is_shallow()` should rather be on `Repo` to support testing multiple commits at the same time. I wouldn't want to encourage per-commit calls given they are IO-heavy then. Alternatively, offer a way to get a set of hashes that are the shallow boundary so users can do their is-contained checks themselves.

harshitayadavv:
@Byron Done, reduced to just the docstring note on stats(). Happy to follow up separately with a Repo-level version (e.g. a method returning the set of shallow-boundary hashes) if that'd be useful.

PR #2173, opened 2026-07-15
fix: respect `core.hooksPath` git config in commit hooks

### LINKED ISSUE #2083: Commit hooks don't respect core.hooksPath in config.
[Commit hooks are more-or-less hardcoded](https://github.com/gitpython-developers/GitPython/blob/main/git/index/fun.py#L85) to use hooks found in `.git/hooks`. This means that specifying `core.hooksPath` in any git config file (i.e. `/etc/gitconfig`, `~/.gitconfig`, `~/git/config`, `$REPO/.git/config`) has no effect. Is this by design?

mangodxd:
## Description

When `core.hooksPath` is set in any git config file (system, global, or repository level), commit hooks should be resolved relative to that path instead of the default `.git/hooks` directory.

Currently, `run_commit_hook()` always uses `hook_path()` which hardcodes `{git_dir}/hooks/{name}`, ignoring the `core.hooksPath` setting.

## Changes

In `run_commit_hook()`, before falling back to the default hook path via `hook_path()`, the function now reads `core.hooksPath` from the repo configuration using `config_reader().get()`. If the setting exists:
- Hooks are resolved relative to that path
- If the path is relative (following git's behavior), it's resolved against the working tree root
- If not set, the existing behavior is preserved unchanged

## Testing

- Existing test `test_run_commit_hook` continues to pass
- The change is minimal (14 lines added) and non-breaking

Fixes #2083

### DIFF
--- a/git/index/fun.py
+++ b/git/index/fun.py
@@ -86,9 +86,7 @@ def run_commit_hook(name: str, index: "IndexFile", *args: str) -> None:
         Respects the ``core.hooksPath`` git configuration option. When set, hooks are
         resolved relative to that path instead of the default ``.git/hooks`` directory.
     """
-    hooks_path = index.repo.config_reader().get(
-        "core", "hooksPath", fallback=None
-    )
+    hooks_path = index.repo.config_reader().get("core", "hooksPath", fallback=None)
     if hooks_path is not None:
         hp = osp.join(hooks_path, name)
         if not osp.isabs(hooks_path):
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
Thanks!

Please add tests.

mangodxd:
Tests added. Covers absolute hooksPath, relative hooksPath, fallback to default, and missing hook. All passing on this end.

mangodxd:
Thanks for the review @Byron. Tests have been added in the latest commits — three hook-path-specific tests plus a fallback test. Would appreciate a re-review when you get a chance.

Byron:
Windows CI is still failing.

Byron:
Closing in favor of https://github.com/gitpython-developers/GitPython/pull/2159.

PR #2183, opened 2026-07-23
Add support for Python 3.15

hugovk:
The Python 3.15 beta is out. 🚀 

The 3.15 release manager (👋 hi, that's me!) [said](https://discuss.python.org/t/python-3-15-0-beta-4-is-here/108208):

> We ***strongly encourage*** maintainers of third-party Python projects to ***test with 3.15*** during the beta phase and report issues found to [the Python bug tracker](https://github.com/python/cpython/issues) as soon as possible. While the release is planned to be feature-complete entering the beta phase, it is possible that features may be modified or, in rare cases, removed up until the start of the release candidate phase (2026-08-04). Our goal is to have ***no ABI changes*** after beta 4 and as few code changes as possible after the first release candidate. To achieve that, it will be ***extremely important*** to get as much exposure for 3.15 as possible during the beta phase.
> 
> This includes creating pre-release wheels for 3.15, as it helps other projects to do their own testing. However, we recommend that your regular production releases wait until 3.15.0rc1, to avoid the risk of ABI breaks.

Also declare support for 3.13-3.15 with Trove classifiers.

And includes a suggestion: stop testing free-threaded 3.13t. It was experimental, 3.14t and up are properly supported and have better performance, and many other projects are stopping testing 3.13t, such as cibuildwheel and Pillow.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os-type: [ubuntu, macos, windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.13t", "3.14", "3.14t"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.13t", "3.14", "3.14t", "3.15", "3.15t"]
         exclude:
         - os-type: macos
           python-version: "3.7"  # Not available for the ARM-based macOS runners.
@@ -25,14 +25,20 @@ jobs:
           python-version: "3.13t"
         - os-type: macos
           python-version: "3.14t"
+        - os-type: macos
+          python-version: "3.15t"
         - os-type: windows
-          python-version: "3.13"  # FIXME: Fix and enable Python 3.13 and 3.14 on Windows (#1955).
+          python-version: "3.13"  # FIXME: Fix and enable Python 3.13-3.15 on Windows (#1955).
         - os-type: windows
           python-version: "3.13t"
         - os-type: windows
           python-version: "3.14"
         - os-type: windows
           python-version: "3.14t"
+        - os-type: windows
+          python-version: "3.15"
+        - os-type: windows
+          python-version: "3.15t"
         include:
         - os-ver: latest
         - os-type: ubuntu
@@ -44,6 +50,10 @@ jobs:
         - python-version: "3.8"
           build-docs: false
         - experimental: false
+        - python-version: "3.15"
+          experimental: true
+        - python-version: "3.15t"
+          experimental: true
 
       fail-fast: false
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -17,22 +17,24 @@ jobs:
...
-          python-version: "3.13t"
         - os-type: macos
           python-version: "3.14t"
+        - os-type: macos
+          python-version: "3.15t"
         - os-type: windows
-          python-version: "3.13"  # FIXME: Fix and enable Python 3.13 and 3.14 on Windows (#1955).
-        - os-type: windows
-          python-version: "3.13t"
+          python-version: "3.13"  # FIXME: Fix and enable Python 3.13-3.15 on Windows (#1955).
```
Byron, line 29: @EliahKagan I wonder if this comment is still valid? We have been running on Windows for a while, even though maybe not with all tests enabled.

My hope would be that with the improvements to deterministic destruction, more tests can work as filehandles will be closed more reliably.

Byron (approved):
What an honor, thanks a lot for setting this up for us!

And of course, good to see that it just works!

.github/workflows/pythonpackage.yml
EliahKagan: The semantics of [`os.path.isabs`](https://docs.python.org/3/library/os.path.html#os.path.isabs) changed in Python 3.13 on Windows. This had caused the behavior of code like `index.remove('/only/looks/absolute')` to differ, such that it was not immediately obvious what the behavior *should* be. It caused a test to fail, but that could not reasonably be fixed without figuring out what ought to actually happen.

I think it's very likely that GitPython has or had bugs that would make this behavior wrong both before and after the chage, though I am not sure. Really, all uses of `os.path.isabs`, including when spelling it `osp.isabs` after `import os.path as osp`, should be audited, and also compared to the uses of [the `is_absolute` method of `Path`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_absolute), which I believe had always regarded paths beginning with a single separator as non-absolute on Windows.

I had meant to fix this in #1955 but that ended up stalling due to my inability to put much time into GitPython at the time, and I very much agree with its closure--I should've myself closed it much earlier, to make clearer to others that they should feel welcome to contribute tests and fixes.

To be clear, paths on Windows that start with a single `/` or `\` are indeed not absolute, because they are relative to the current drive. (A path that starts with a single `/` on a Unix-like system is of course absolute.) However, it may be that existing code in GitPython also assumes that, even on Windows, all relative paths are relative to a specific directory, rather than unconditionally specifying a root directory but relative to the current drive. So the way GitPython handles relative and absolute paths in general, including but not limited to when it uses `os.path.isabs` or `pathlib.Path.is_absolute` or related functions, should be examined in case there are bugs.

I suspect that, these days, all of this can be examined, and any findings rechecked and remediated, fairly efficiently.

Byron: Thanks so much for the clarification. It took me a while to realize that a leading slash means relative to the current drive, as opposed to the current dir which I was auto-correcting this into :D.

GitPython doesn't have the infrastructure yet for discussion to keep track of things, probably also because I don't feel too compelled to actually do it 😅. Maybe once Gitoxide is in a place where it could be integrated I will feel the need to do a quality push.

PR #2187, opened 2026-07-25
Improve RemoteProgress parse return typing

### LINKED ISSUE #1834: RemoteProgress._parse_progress_line may merit a broader return type
`RemoteProgress._parse_progress_line` is conceptually a "protected" method: code outside GitPython may subclass `RemoteProgress` and override it with another implementation (which may also reasonably call the base class implementation).

The method is conceptually void as implemented in the base class, always implicitly returning `None`, and it is annotated accordingly:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/util.py#L582

However, while it would not ordinarily be useful to do so, it is--or *has* been--also intended that it be overridable to return a non-`None` value. This is mostly for compatibility with existing code that may have done so in the past, but it may be that reasonable cases for doing so in new code can still arise. This is reflected in code of GitPython:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/git/util.py#L685-L697

And in the subclass of `RemoteProgress` introduced in the test suite, which overrides the method:

https://github.com/gitpython-developers/GitPython/blob/afa575454f85b34800460881cc9c3cd7fe78c8e3/test/test_remote.py#L54-L59

But any override that returns a non-`None` value is currently inconsistent with the return type annotation on the base class, which is written as `-> None`. For Liskov substitutability, an overridden method's return type must be the same or a subclass of the base method's return type, and if a subclass has type annotations, and the return type is specified as a non-`Any` type, type checkers such as `mypy` and `pyright` will report a non-`None` return type as a violation.

If possible, the return type in `RemoteProgress` should be broadened, but in a way that does not unduly encourage subclasses to return non-`None`. **See comments on #1788, starting at https://github.com/gitpython-developers/GitPython/pull/1788#issuecomment-1869653326, for a discussion and analysis of this.** This issue, and the above recommendation, is based on the findings and discussion there.

pick7:
## Summary

- Broaden `RemoteProgress._parse_progress_line()` from `None` to `Optional[object]` so subclass overrides may return a value without violating the base type.
- Make the base implementation return `None` explicitly and keep `new_message_handler()` ignoring that return value.
- Add a mypy-covered regression test with an override that returns a non-`None` value.

Closes #1834

## Checks

- `.venv/bin/python -m pytest test/deprecation/test_remote_progress.py`
- `.venv/bin/python -m mypy git/util.py test/deprecation/test_remote_progress.py`
- `.venv/bin/python -m ruff check git/util.py test/deprecation/test_remote_progress.py`
- `.venv/bin/python -m ruff format --check git/util.py test/deprecation/test_remote_progress.py`

## AI agent disclosure

OpenAI Codex prepared this scoped code/test change and PR text. I reviewed the diff and ran the checks above.

### DIFF
--- a/git/util.py
+++ b/git/util.py
@@ -611,13 +611,17 @@ class RemoteProgress:
         self.error_lines: List[str] = []
         self.other_lines: List[str] = []
 
-    def _parse_progress_line(self, line: AnyStr) -> None:
+    def _parse_progress_line(self, line: AnyStr) -> Optional[object]:
         """Parse progress information from the given line as retrieved by
         :manpage:`git-push(1)` or :manpage:`git-fetch(1)`.
 
         - Lines that do not contain progress info are stored in :attr:`other_lines`.
         - Lines that seem to contain an error (i.e. start with ``error:`` or ``fatal:``)
           are stored in :attr:`error_lines`.
+
+        The base implementation returns ``None``. Subclasses may return another
+        value for compatibility with existing overrides, but callers should treat
+        the return value as unspecified.
         """
         # handle
         # Counting objects: 4, done.
@@ -632,7 +636,7 @@ class RemoteProgress:
 
         if self._cur_line.startswith(("error:", "fatal:")):
             self.error_lines.append(self._cur_line)
-            return
+            return None
 
         cur_count, max_count = None, None
         match = self.re_op_relative.match(line_str)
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks, LGTM!

PR #2189, opened 2026-07-25
Honor kill_after_timeout with output streams

### LINKED ISSUE #1762: Git.execute output_stream is not documented to suppress kill_after_timeout, but does

pick7:
## Summary

- Start the existing timeout watchdog when output_stream is used.
- Keep the watchdog active until the subprocess has actually exited.
- Add a regression test where stdout closes before the child process sleeps.

Fixes #1762.

## Validation

- .venv/bin/python -m pytest -o addopts= test/test_git.py
- .venv/bin/python -m ruff check git/cmd.py test/test_git.py
- .venv/bin/python -m ruff format --check git/cmd.py test/test_git.py
- .venv/bin/python -m mypy git/cmd.py

## AI agent disclosure

This pull request was prepared and submitted by OpenAI Codex acting as an AI agent through the contributor account.

### DIFF
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1473,10 +1473,7 @@ class Git(metaclass=_GitMeta):
                 return
 
             def make_timeout_error() -> Union[str, bytes]:
-                err = 'Timeout: the command "%s" did not complete in %d secs.' % (
-                    " ".join(redacted_command),
-                    timeout,
-                )
+                err = f'Timeout: the command "{" ".join(redacted_command)}" did not complete in {timeout:g} secs.'
                 return err if universal_newlines else err.encode(defenc)
 
             def communicate() -> Tuple[AnyStr, AnyStr]:
--- a/git/cmd.py
+++ b/git/cmd.py
@@ -1435,6 +1435,10 @@ class Git(metaclass=_GitMeta):
         if as_process:
             return self.AutoInterrupt(proc, command)
 
+        watchdog: Optional[threading.Timer] = None
+        kill_check: Optional[threading.Event] = None
+        timeout_error: Optional[Callable[[], Union[str, bytes]]] = None
+
         if sys.platform != "win32" and kill_after_timeout is not None:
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, this circumstantially looks good to me.

PR #2200, opened 2026-07-30
Revert "Resolve Windows hook Bash through PATH"

Byron:
Reverts gitpython-developers/GitPython#2199


Byron:
@EliahKagan On the go right now, but let’s start by reverting this.

PR #2209, opened 2026-08-10
index: write blobs via git hash-object, not gitdb's odb.store

### LINKED ISSUE #2021: repo.index.add() method attempts to chmod files which it should not
The behavior of the `repo.index.add()` method behaves differently than the CLI equivalent `git add`:
```
0$ git add *
0$ python3 -c 'import git; git.Repo.init().git.add("*")'
0$ python3 -c 'import git; git.Repo.init().index.add("*")'
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/git/index/base.py", line 885, in add
    entries_added.extend(self._entries_for_paths(paths, path_rewriter, fprogress, entries))
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/git/util.py", line 176, in wrapper
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/git/index/util.py", line 111, in set_git_working_dir
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/git/index/base.py", line 745, in _entries_for_paths
    entries_added.append(self._store_path(filepath, fprogress))
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/git/index/base.py", line 698, in _store_path
    istream = self.repo.odb.store(IStream(Blob.type, st.st_size, stream))
  File "/opt/conda/base/envs/cds/lib/python3.10/site-packages/gitdb/db/loose.py", line 233, in store
    chmod(obj_path, self.new_objects_mode)
PermissionError: [Errno 1] Operation not permitted: '/path/to/repo/.git/objects/02/60b9d2b53981a5ddbe20f3f0aae350b2806d22'
1$ 
```
I don't see any reason that the python-native methods should be doing anything different than what the CLI does.

caroescm:
## Problem

`repo.index.add()` is GitPython's own way of adding files, separate from just running `git add`. To do that, it writes the file's content into `.git/objects` itself using Python, instead of asking `git` to do it. As part of that, it also tries to change the new file's permissions (`chmod`) — and in some setups, that permission change isn't allowed, even though writing the file itself worked fine.

So `index.add()` can crash with `PermissionError` in situations where plain `git add` (and `repo.git.add(...)`, which just runs `git add` directly) work without any problem, on the same file.

There's also a smaller side effect: because GitPython reads the file itself instead of asking `git`, it skips any `.gitattributes` rules (like converting line endings), so the content it stores can end up slightly different from what `git add` would store.

## What I changed

I changed `index.add()` to use the real `git hash-object -w` command to write the file into the object database, instead of doing it in Python. This means:

- No more Python-side `chmod`, so the crash goes away.
- The stored content now matches what `git add` would store, filters included.

Symlinks needed a small extra step: `git hash-object` normally follows a symlink and hashes the file it points to, but Git is supposed to store the symlink's target path itself, not the pointed-to file. So for symlinks, I write the target path to a temporary file and hash that instead.

Fixes #2021

## Testing

- Ran the existing test suite (`test_index.py`, `test_base.py`, `test_repo.py`) — all still pass.
- Compared `index.add()`'s output to real `git add` for a normal file, a symlink, and a file in a subfolder — identical results.
- Recreated the original bug by blocking `chmod` only on `.git/objects` files (same as the error in #2021) — `index.add()` now works instead of crashing.

### DIFF
--- a/git/db.py
+++ b/git/db.py
@@ -5,10 +5,14 @@
 
 __all__ = ["GitCmdObjectDB", "GitDB"]
 
-from gitdb.base import OInfo, OStream
+from subprocess import PIPE
+
+from gitdb.base import IStream, OInfo, OStream
 from gitdb.db import GitDB, LooseObjectDB
 from gitdb.exc import BadObject
+from gitdb.fun import stream_copy
 
+from git.compat import force_text
 from git.util import bin_to_hex, hex_to_bin
 from git.exc import GitCommandError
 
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for tackling this!

I will see if an alternative solution can do the trick, as such a localized override goes against the original architecture.

PR #2216, opened 2026-08-14
fix: handle non-UTF-8 packed refs

### LINKED ISSUE #2064: Encoding issue with tags in packed-refs file
Tags found in the `packed-refs` file that can be created by the Git command `git pack-refs` do not have to be UTF8 encoded. In these cases, GitPython fails to read these tags due to assuming they should be UTF8 compatible.

The source for this issue is found here:
https://github.com/gitpython-developers/GitPython/blob/main/git/refs/symbolic.py#L124

```python
with open(cls._get_packed_refs_path(repo), "rt", encoding="UTF-8") as fp:
``` 

A working example (using a repository with a non-utf8 tag):
```shell
git clone https://github.com/ACRA/acra
cd acra
``` 

Create a Python script with the following content:
```python
import git
repo = git.Repo(".")
print(repo.tags)
```
Execute the script:
```shell
python script.py
```
Result:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 6216: invalid continuation byte`

Sohel2309:
Fixes #2064.

## Bug

GitPython raises `UnicodeDecodeError` when iterating `.git/packed-refs` if a packed ref name contains bytes that are not valid UTF-8.

Git ref names are byte strings and are not guaranteed to contain valid UTF-8. A single non-UTF-8 ref should not make the entire `packed-refs` file unreadable.

## Root Cause

`SymbolicReference._iter_packed_refs()` opened `packed-refs` in text mode with strict UTF-8 decoding:

```python
open(..., "rt", encoding="UTF-8")
```

This caused `UnicodeDecodeError` when a packed ref contained non-UTF-8 bytes.

## Fix

Read `packed-refs` in binary mode and decode each line using `surrogateescape`.

This follows the existing lenient byte-to-string handling used elsewhere in GitPython and preserves the original bytes during round-tripping instead of raising an exception.

## Regression Test

Added a regression test covering a packed ref containing a non-UTF-8 byte and verifying that iterating packed refs does not raise `UnicodeDecodeError`.

## Validation

- Targeted regression test: passed
- `test/test_refs.py`: regression test passed
- `ruff check`: passed
- `ruff format --check`: passed
- `git diff --check`: passed

### DIFF
--- a/git/refs/symbolic.py
+++ b/git/refs/symbolic.py
@@ -149,9 +149,18 @@ class SymbolicReference:
             The packed refs file will be kept open as long as we iterate.
         """
         try:
-            with open(cls._get_packed_refs_path(repo), "rt", encoding="UTF-8") as fp:
-                for line in fp:
-                    line = line.strip()
+            # Read in binary mode and decode leniently: ref names packed by `git
+            # pack-refs` are arbitrary byte strings and are not guaranteed to be valid
+            # UTF-8 (see e.g. core.precomposeUnicode-unaffected filesystems, or refs
+            # created on a system with a different locale). Decoding strictly as UTF-8
+            # would raise UnicodeDecodeError and make the entire packed-refs file
+            # unreadable because of a single such ref. Use the same lenient
+            # byte<->str roundtrip ("surrogateescape") already used elsewhere in
+            # GitPython (see :func:`git.compat.safe_decode`) so that such refs are
+            # preserved rather than crashing iteration.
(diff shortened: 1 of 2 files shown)


Byron:
I don't think that making this pass is the right choice here.
Instead, GitPython shouldn't assume encodings at all, or at most ascii, and get its typing straight.
A lot of work will be needed to fix this, and I plan to remove all the pure python implementations in favor of using the Git binary instead. 
Thanks for your understanding.

PR #2219, opened 2026-08-27
fix: wait for git daemon to listen before connecting

### LINKED ISSUE #1676: Occasional "fatal: unable to connect to localhost" on CI

Vogtinator:
Properly wait instead of hardcoding a short sleep.

Fixes #1676

### DIFF
--- a/test/lib/helper.py
+++ b/test/lib/helper.py
@@ -32,6 +32,7 @@ import logging
 import os
 import os.path as osp
 from stat import S_ISLNK, ST_MODE
+import socket
 import subprocess
 import sys
 import tempfile
@@ -218,8 +219,15 @@ def git_daemon_launched(base_path, ip, port):
                 base_path=base_path,
                 as_process=True,
             )
-        # Yes, I know... fortunately, this is always going to work if sleep time is just large enough.
-        time.sleep(1.0 if sys.platform == "win32" else 0.5)
+
+        # Wait until git daemon listens for connections.
+        for _attempt in range(1, 30):
+            try:
+                socket.create_connection((ip, port), timeout=30).close()
+                break
+            except ConnectionRefusedError:
+                time.sleep(0.5)
+
     except Exception as ex:
         msg = textwrap.dedent(
             """


Byron (approved):
Let's go with this, thank you!

A follow-up could mean to reduce the time to sleep, as I'd expect the socket to become available much earlier, which should make dependent tests start running a little bit faster.

Vogtinator:
The ideal fix is to change `git daemon --detach` in a way that it opens the socket for listening before daemonizing, then the method here can just do a blocking start, ensuring that once it exits with 0 that it accepts connections. Also makes error handling easier: Currently if `git daemon` is not installed, the test suite fails with "connection refused", without a helpful message.

PR #2227, opened 2026-09-03
fix: join backslash line continuations when reading config values

nkbeast:
git config joins an unquoted value that ends in a backslash with the next line: the backslash and the newline are removed and the next line is appended verbatim. Reproducing with git itself:

```
$ printf '[a]\n\tk = line1\\\n line2\n' > g.cfg
$ git config -f g.cfg a.k
line1 line2
```

GitPython only implemented multi-line values for *quoted* strings. For unquoted values the continuation line was silently dropped — `GitConfigParser` read back `line1\` and the ` line2` content vanished, truncating whatever the user had stored (aliases, hook commands, any hand-edited multi-line value).

The fix reads the continuation inside `_read` and joins it, chaining across lines that themselves end in a backslash, with the exact rules git uses:

- a single trailing backslash → continuation, next line appended verbatim (leading whitespace included)
- an even number of trailing backslashes → an escaped one, value ends there (`val\\\\` stays `val\\\\`)
- a lone backslash right at end-of-file → dropped

All five cases are asserted against git's own output in the new test.

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -467,6 +467,19 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
 
         # END string_decode
 
+        def is_line_continuation(value: str) -> bool:
+            quoted = escaped = False
+            for char in value:
+                if escaped:
+                    escaped = False
+                elif char == "\\":
+                    escaped = True
+                elif char == '"':
+                    quoted = not quoted
+                elif char in "#;" and not quoted:
+                    return False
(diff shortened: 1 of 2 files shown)


Byron:
Thanks!

As a note: The current Git configuration file parser is inherently insufficient and has many bugs. Fixing them all isn't the goal for GitPython 3.X, it's only remedy.
Version 4 will address these issues (much more) properly

PR #2229, opened 2026-09-07
Read git's yes/no and on/off booleans in get_value

rawsun007:
> **This pull request was written by Claude Code (Claude Opus 5) working through @rawsun007's account**, identified here per CONTRIBUTING.md's "Prevent agent impersonation". The facts below were measured locally; @rawsun007 has not reviewed the diff line by line yet and will answer review comments himself.

`GitConfigParser` disagrees with itself about the same file. `getboolean` (from `ConfigParser`) accepts all of git's boolean spellings; `get_value` handled only `true`/`false`, under a comment saying it tries "boolean values as git uses them":

| value | `get_value()` | `getboolean()` | `git config --type=bool` |
| --- | --- | --- | --- |
| `yes` | `'yes'` | `True` | true |
| `no` | `'no'` | `False` | false |
| `on` | `'on'` | `True` | true |
| `off` | `'off'` | `False` | false |
| `true` / `false` | `True` / `False` | same | same |

Returning them as strings is worse than merely inexact: `"no"` and `"off"` are non-empty, so

```python
if repo.config_reader().get_value("core", "someflag"):
```

is **True** for a value git reads as false, and nothing raises.

`_string_to_value` now recognises the same spellings `getboolean` does. Non-boolean values are untouched (`"meld"` stays a string) and numeric values keep their current behaviour.

Deliberately not included: `get_value` also diverges on numeric bases and suffixes — `"0x10"` and `"1k"` come back as strings where git reads 16 and 1024, and `"010"` becomes 10 where git reads octal 8. Those change a value rather than its type, so they belong in their own commit; happy to send that separately.

Validation: `test_get_value_reads_git_boolean_spellings` covers the ten spellings and asserts the two accessors agree — it fails on the parent commit. `test/test_config.py` passes (40 passed, 2 skipped). `ruff check` and `ruff format --check` clean at the repo's line length. `test/test_repo.py` errors on my clone because `init-tests-after-clone.sh` has not been run; unchanged by this commit.

### DIFF
--- a/git/config.py
+++ b/git/config.py
@@ -956,11 +956,16 @@ class GitConfigParser(cp.RawConfigParser, metaclass=MetaParserBuilder):
                 continue
         # END for each numeric type
 
-        # Try boolean values as git uses them.
+        # Try boolean values as git uses them. git accepts yes/no and on/off as
+        # well as true/false (git_parse_maybe_bool_text in parse.c), and so does
+        # ConfigParser.getboolean on this class, so only get_value lagged behind.
+        # Leaving them as strings was worse than merely inexact: "no" and "off"
+        # are non-empty, so a caller testing the result got True for a value git
+        # reads as false.
         vl = valuestr.lower()
-        if vl == "false":
+        if vl in ("false", "no", "off"):
             return False
-        if vl == "true":
+        if vl in ("true", "yes", "on"):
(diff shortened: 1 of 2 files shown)


Byron:
Thanks, great catch!

PR #2242, opened 2026-09-16
Move project metadata to pyproject.toml

mwtoews:
This PR is a near 1:1 conversion of metadata from `setup.py` to `pyproject.toml`, as [supported by setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). There are a few differences from the legacy forms:

- Multiple authors and emails are listed
- Multiple URLs are listed
- A keyword "git" is added


Byron:
Thanks a lot, I am definitely interested in mostly getting rid of `setup.py`. Once CI passes, I think this should be merged and if there is trouble with the next publish, it's probably an easy fix, too.

Byron:
Something to consider: 

> - [P2] Raise the setuptools minimum to 62.6 — /Users/byron/dev/github.com/gitpython-developers/GitPython.pyproject-toml/pyproject.toml:2-2
  Setuptools 61.x–62.5.x satisfies this requirement but cannot validate the new `tool.setuptools.dynamic.dependencies` and `optional-dependencies` entries: file-based dependency directives were introduced in 62.6.0. Source builds using those older backends therefore fail despite meeting the declared build requirements. Raise the lower bound to `setuptools >= 62.6`.

mwtoews:
The failing test with Python 3.7 is `test/test_installation.py::TestInstallation::test_installation` with message
> setuptools.config.pyprojecttoml._BetaConfiguration: Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.

This was true at the time (for that version of setuptools) but not today, so PYTHONWARNINGS is set to "default" for Python 3.7.

The setup.py file might be needed until Python 3.9 is the minimum, as it needs to work-around setuptools deprecation messages [described here](https://setuptools.pypa.io/en/latest/userguide/license_migration.html). The workaround is to set "licence" as dynamic, even though it isn't.

mwtoews:
> Something to consider ... Raise the lower bound to `setuptools >= 62.6`.

Done.

PR #2243, opened 2026-09-16
Move [doc] extra to a PEP 735 dependency group; add uv-sync CI check

mwtoews:
This PR has two related changes. The first is to move the `[doc]` optional dependency to a dependency group ([PEP 735](https://peps.python.org/pep-0735/)). This requires [pip v25.1](https://pip.pypa.io/en/stable/news/#v25-1) available for Python 3.9 or later to work; older Python versions will ignore this data. The contents of `doc/requirements.txt` is moved to `pyproject.toml`, and `.readthedocs.yaml` is adjusted to use the dependency group.

The second change is to resolve an incompatibility with the project dependencies identified with [uv sync](https://docs.astral.sh/uv/concepts/projects/sync/):
```shell
$ uv sync --all-extras --all-groups --dry-run
Using CPython 3.14.2
Would create project environment at: .venv
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.8.*'):
  ╰─▶ Because the requested Python version (>=3.7) does not satisfy Python>=3.9 and sphinx==7.4.7 depends on Python>=3.9, we can conclude that sphinx==7.4.7 cannot be used.
      And because only the following versions of sphinx are available:
          sphinx<=7.4.7
          sphinx>=8
      we can conclude that sphinx>=7.4.7,<8 cannot be used.
      And because gitpython[doc] depends on sphinx>=7.4.7,<8 and your project requires gitpython[doc], we can conclude that your project's requirements are unsatisfiable.

hint: While the active Python version is 3.14, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The `requires-python` value (>=3.7) includes Python versions that are not supported by your dependencies (e.g., sphinx==7.4.7 only supports >=3.9). Consider using a more restrictive `requires-python` value (like >=3.9).
```
The fix is to simply to limit the version via `sphinx >=7.4.7,<8 ; python_version >= "3.9"'`. This change should be fine, since versions before Python 3.9 can't install the dependency group and are not tested to build the documentation in the CI.

A CI check is added to see if the dependencies can be resolved for all extras and all groups for all supported versions of Python.


Byron:
Thanks.

PR #2244, opened 2026-09-17
submodule: option to update without fetching

jeffro256:
Enables behavior similar to `git submodule update --no-fetch`. Useful for A) builds with "online" then "offline" stages, and B) saving time while performing many checkouts and recursive submodule updates when you know that you have the commits locally.


jeffro256:
Force=pushed changes to make the linter happy

Byron:
Thanks! Seems simple and useful at the same time.

PR #2249, opened 2026-09-21
fix: skip a UTF-8 BOM when reading config files

geek-alpha:
## Summary

`git` skips a UTF-8 BOM at the start of a config file. `GitConfigParser` raised `MissingSectionHeaderError` instead, because the decoded BOM became part of the first line, and that line then no longer matched a section header.

- Skip the BOM bytes when reading the first line of a config file.
- Add a regression test that fails without the change.

## Why this is a bug and not malformed input

git reads such a file fine, so GitPython disagreeing with git here is the bug. The error is also hard to act on: the file clearly starts with `[core]`, yet the message says it contains no section headers.

## Validation

git 2.47.3, file starting with the three BOM bytes `EF BB BF`:

```
$ git config -f bom.cfg --list
core.bare=true
$ echo $?
0
```

GitPython before the change:

```
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'bom.config', line: 1
'\ufeff[core]\n'
```

After the change `get_value("core", "bare")` is `True`, matching git. A BOM file reached through `include.path` is read the same as git as well (`user.name=bom-inc`).

Test runs:

- `pytest test/test_config.py -o addopts= -q` -> `48 passed, 2 skipped, 20 subtests passed`
- `pytest test/ -o addopts= -q` -> `988 passed, 82 skipped, 1 xfailed`
- `ruff check` / `ruff format --check` on both files -> clean

## AI agent disclosure

Per CONTRIBUTING ("Prevent agent impersonation"): this PR was prepared by an AI
agent (Battle Phoenix) driving the `geek-alpha` account. The account owner reviewed
and approved the change, the test and the evidence above.


Byron (approved):
Great, thanks!

